PR #37, opened 2016-07-12
Adding support for common RESTful HTTP response codes

### LINKED ISSUE #34: Use exceptions to return error responses
I've not delved too deep into this yet, so I'm not sure how straight forward this would be to implement...

Having the ability to use exceptions to return an error response (and correct HTTP response code) would be extremely useful. A pseudo example:

Exception class:
```
class UnauthorizedException(ChaliceError):
    STATUS_CODE = 401
```

App logic:
```
if not authenticate(user):
    raise UnauthorizedException('You are not authenticated')
```

Response example (inc. correct response headers):
```
{"error": e.message}
```

Having quickly checked the source code, it looks like it should just be a cause of creating the exceptions and adding them to the `ALL_ERRORS` list. 

Thoughts?

JamieCressey:
Fixes #34

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -27,17 +27,39 @@ class BadRequestError(ChaliceViewError):
     STATUS_CODE = 400
 
 
+class UnauthorizedError(ChaliceViewError):
+    STATUS_CODE = 401
+
+
+class ForbiddenError(ChaliceViewError):
+    STATUS_CODE = 403
+
+
 class NotFoundError(ChaliceViewError):
     STATUS_CODE = 404
 
 
+class ConflictError(ChaliceViewError):
+    STATUS_CODE = 409
+
+
+class TooManyRequestsError(ChaliceViewError):
+    STATUS_CODE = 429
+
+
 ALL_ERRORS = [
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the pull request, I think this will be a great addition.

I'll need to update one of the unit tests that verifies all the API calls that are made for a deploy command, it's currently failing because of the new API calls we're making to set up the new errors.

Now that the repo's public, I need to get travis set up so we can get this test feedback earlier.

JamieCressey:
Any news on the unit tests, @jamesls?

Happy to help with the Travis setup in anyway I can.

DeviaVir:
@JamieCressey seems ready to go:
https://github.com/awslabs/chalice/commit/c9f71cd084605ba109c10e292f7c591051e09c61

and:
https://travis-ci.org/awslabs/chalice

jamesls:
Sorry for the delay, I've been OOTO until today.  I'm taking a look now.

jamesls:
Added test and rebased against master.  Merged via fef8f1361f997e54c17df4951798c0016de0c1fd.  Thanks again for the pull request!

PR #38, opened 2016-07-12
Add profile support when creating new projects

### LINKED ISSUE #28: Add the ability to specify a profile on the CLI
Some (most?) of us have access to multiple AWS accounts from the command line. It would be super-handy to be able to specify which profile to use with the `--profile` CLI argument like you can with `awscli`.

JamieCressey:
Fixes #28

### DIFF
--- a/chalice/deployer.py
+++ b/chalice/deployer.py
@@ -154,6 +154,7 @@ def node(name, uri_path, is_route=False):
 
 
 class NoPrompt(object):
+
     def confirm(self, text, default=False, abort=False):
         return default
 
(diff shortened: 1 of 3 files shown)


chalice/deployer.py
```diff
@@ -163,10 +164,13 @@ class Deployer(object):
     LAMBDA_CREATE_ATTEMPTS = 5
     DELAY_TIME = 3
 
-    def __init__(self, session=None, prompter=None):
+    def __init__(self, session=None, prompter=None, config=None):
```
jamesls: It might be simpler if we just add a `profile` param to the `__init__` for now.

We can always switch this to `config` if we end up needing more things in the `__init__`.

JamieCressey:
@jamesls - Changed to only pass a `profile` param to `__init__`, rebased and squashed.

jamesls:
Thanks!

Issue #61, opened 2016-07-21
make IAM configurable


sdole:
Running chalice deploy attempts to create a IAM role of the name of the application. Unfortunately, in my enterprise, not all AWS developers are allowed to create an IAM role for security best practices reasons. The entire IAM roles and policy thing should be configurable so that an IAM admin can create the required role/policy beforehand and grant use of it to developers.

sdole:
see pull #62

jamesls:
Thanks for the feature request.  I think this would be good to add.  I'll take a look at your pull request soon.

sdole:
You are welcome. Let me know if you have questions or if you want me to try a different design/approach.

jamesls:
Fixed via https://github.com/awslabs/chalice/pull/85.  Thanks again for the pull request.

PR #62, opened 2016-07-21
allow using a pre-configured IAM role 

sdole:
see issue #61. This is my proposal to fix it.


jamesls:
Sorry for the delay, I've been OOTO until today.

Overall, I think this change looks good.  I do have one question.  In considering additional features such as https://github.com/awslabs/chalice/issues/40 where I'd like to track everything I create so that I can subsequently delete them in a `chalice delete` command, I'm thinking about writing to disk every thing I create, possibly in the `config.json`.  If that's the case, it might help if the config was more explicit about whether or not to create a role and autogenerate a policy.  What do you think about something like this:

```
{"role_arn": "arn:....", "manage_role": false}
 ```

Not really sure what to call the config option, but some sort of boolean config option that explicitly says "don't touch the role_arn".

What do you think?

sdole:
good idea! done. also, changed the config keys to iam_role_arn and manage_iam_role so that it is clear we are talking about iam roles. thanks!

sdole:
Hello, Let me know if you want me to make any changes. I am fairly new to python, so, don't hesitate to comment on my code - i will learn something. thanks.

sdole:
this build seems to have failed due to some documentation changes unrelated to this pull request.

jamesls:
Thanks.  Would you mind squashing your commits and rebasing against master?  That should fix the doc errors you're seeing.

sdole:
squashing and rebase was getting complicated. I just re-did a new fork, manually applied changes and requested a new pull #85. Closing this.

PR #68, opened 2016-07-22
Fix samples in README

abrogger:
The /cities/{city} examples later refer to the parameter as `name` instead of `city`


jamesls:
Thanks!

PR #83, opened 2016-08-01
Print original stacktrace on debug errors

### LINKED ISSUE #50: ChaliceViewError Should Print Real Error
**tl;dr** ChaliceViewError should print the underlying error message instead of hiding it (I think?)

[heavily edited]

By default, ChaliceViewError does not raise the underlying exception, it only prints "An internal server error occurred."
When app.debug = True, ChaliceViewError instead raises the underlying exception.

I think both of these features should be enabled at the same time. ChaliceViewError should print the real message (so that it is logged in CloudWatch) and then return a nicely-wrapped ChaliceViewError to the client. This way the developer can get errors logged without worrying about 'null' returns on bad requests.

I've had to do this to gain this functionality:
`
    except Exception as e:
        print(e)
        raise e
`

And still every single error just gets raised as

`
  File "/var/task/chalice/__init__.py", line 148, in __call__
    raise e
`
Which isn't helpful.

jamesls:
Before:

```
{
    "errorMessage": "Raising an exception!",
    "errorType": "RuntimeError",
    "stackTrace": [
        [
            "/var/task/chalice/__init__.py",
            198,
            "__call__",
            "raise e"
        ]
    ]
}
```

After:
```
{
    "errorMessage": "Raising an exception!",
    "errorType": "RuntimeError",
    "stackTrace": [
        [
            "/var/task/chalice/__init__.py",
            192,
            "__call__",
            "response = view_function(*function_args)"
        ],
        [
            "/var/task/app.py",
            9,
            "index",
            "return a()"
        ],
        [
            "/var/task/app.py",
            12,
            "a",
            "return b()"
        ],
        [
            "/var/task/app.py",
            15,
            "b",
            "raise RuntimeError(\"Raising an exception!\")"
        ]
    ]
}
```

This also shows up in `chalice logs`:

```
2016-08-01 16:14:06.897000 dcd6c3 Raising an exception!: RuntimeError
  File "/var/task/chalice/__init__.py", line 192, in __call__
    response = view_function(*function_args)
  File "/var/task/app.py", line 9, in index
    return a()
  File "/var/task/app.py", line 12, in a
    return b()
  File "/var/task/app.py", line 15, in b
    raise RuntimeError("Raising an exception!")
RuntimeError: Raising an exception!
```

Fixes #50.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,3 +18,5 @@ Next Release
   (`#79 <https://github.com/awslabs/chalice/issues/79>`__)
 * Ignore lambda expressions in policy analyzer
   (`#74 <https://github.com/awslabs/chalice/issues/74>`__)
+* Print original error traceback in debug mode
+  (`#50 <https://github.com/awslabs/chalice/issues/50>`__)
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -190,11 +190,11 @@ class Chalice(object):
                                        event['stage-variables'])
         try:
             response = view_function(*function_args)
-        except Exception as e:
+        except Exception:
             if self.debug:
                 # If the user has turned on debug mode,
                 # we'll let the original exception propogate so
                 # they get more information about what went wrong.
-                raise e
+                raise
             raise ChaliceViewError("An internal server error occurred.")
         return response


woodlandh:
Thanks for implementing this!

PR #87, opened 2016-08-03
Add policy mapping for ListObjectsV2

daha:
The S3 operation ListObjectsV2 is missing in the mapping to policies. This pull request adds mapping for ListObjectsV2.


jamesls:
Thanks!

PR #90, opened 2016-08-07
fixing cli helloworld demo 

patrickpierson:
demo code that gets added when running ```chalice new-project helloworld``` is invalid.  when running ```chalice deploy``` an error is returned ```Error when deploying: Route cannot end with trailing slash: /users/``` This fixes it.


jamesls:
Thanks!

PR #95, opened 2016-08-10
allow overriding the profile used at deploy time.

binarydud:
This would allow multiple accounts to use the same codebase for deployments


b-b3rn4rd:
good stuff, definitely need this option for instance, when environments are separated by accounts

jamesls:
Thanks for the pull request.

Issue #96, opened 2016-08-10
support for non JSON content-types input


binarydud:
Similar to #30, it would be nice to be able to use incoming content-types that are not `application/json`. The use case i'm thinking about in particular is for building slack bots using chalice. A custom slack command sends a `application/x-www-form-urlencoded` payload to the receiving application.

I've got a proof of concept that I will put up in a pull request this evening.

jamesls:
This would be useful, I look forward to seeing what you come up with.

jamesls:
Merged in https://github.com/awslabs/chalice/commit/207fbf39b1d2ba07b5fa1f0508b9afae230eded4.  I added a few tests and documentation.

Thanks again for the pull request, I think this will be a really useful feature.

PR #101, opened 2016-08-11
add incoming content types to routes

binarydud:
this is my proof of concept for adding the ability to accept content-types other than application/json, see #96 .

The app.route decorator would look something like:
```
@app.route(
    '/',
    methods=['POST', 'GET'],
    content_types=['application/json', 'application/x-www-form-urlencoded']
)
def index():
    ...
```

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -67,7 +67,10 @@ class Request(object):
         self.uri_params = uri_params
         self.method = method
         #: The parsed JSON from the body.
-        self.json_body = body
+        if self.headers.get('Content-Type', '') == "application/json":
+            self.json_body = body
+        else:
+            self.json_body = None
         # This is the raw base64 body.
         # We'll only bother decoding this if the user
         # actually requests this via the `.raw_body` property.
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -97,7 +97,8 @@ class RouteEntry(object):
             methods,
             authorization_type=None,
             authorizer_id=None,
-            api_key_required=False):
+            api_key_required=False,
+            content_types=None):
         self.view_function = view_function
         self.view_name = view_name
         self.uri_pattern = path
@@ -108,6 +109,7 @@ class RouteEntry(object):
         #: A list of names to extract from path:
         #: e.g, '/foo/{bar}/{baz}/qux -> ['bar', 'baz']
         self.view_args = self._parse_view_args()
+        self.content_types = content_types
 
     def _parse_view_args(self):
         if '{' not in self.uri_pattern:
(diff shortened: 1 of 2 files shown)


nmyster:
👍 is this being merged anytime soon?

nmyster:
I've tested this and works perfectly

JamieCressey:
I'm personally of the opinion Chalice should remain minimalistic but
provide the extensibility to develop 3rd party modules; similar to Flask.

"Chalice-HTML" for returning HTML content
"Chalice-Events" for running Chalice functions upon AWS events (SNS, S3,
DynamoDB, CloudWatch Schedules etc.)

jamesls:
Thanks for the pull request.

I'd like to spend some more time experimenting with this pull request.  I have a few use cases I want to try out first.

While I do want chalice to remain minimalistic, I think I'm inclined to say that "rest APIs" is where the line should be drawn (vs. `chalice-events`, etc).  As long as it's still pretty straightforward to create APIs that work with XML/HTML, I'd be inclined to support this.

I'll have more concrete feedback in the next day or so.

jamesls:
Overall, I think this pull request is pretty close.  I just had a few questions/comments.

One of the things I'm noticing is that the body type will vary based on the content type.  For JSON, you get the `json.loads(body)` value.

However, for other content types, you get a string value.  This is misleading if you're using `json_body`, which is arguably misnamed to begin with.  More concretely:

```
@app.route(
    '/',
    methods=['POST', 'GET'],
    content_types=['application/json', 'application/x-www-form-urlencoded',
                   'application/xml'],
)
def index():
    return app.current_request.json_body
```

* application/json - `app.current_request.json_body` is the parsed JSON content
* application/xml - `app.current_request.json_body` is a string
* application/x-www-form-urlencoded - `app.current_request.json_body` is a string

Questions:

* Should `json_body` return None if the content type is not application/json?
* Should there be some sort of helper properties or parsing mechanisms for working with other content types?  e.g parsing form encoded bodies, parsing xml, etc.  I think these would be useful but I think these would bloat the API.

binarydud:
> Should json_body return None if the content type is not application/json?

I think that makes a lot of sense. Generally if I'm accepting application/xml, i'm not going to be using a property with json in the name

> Should there be some sort of helper properties or parsing mechanisms

I think that this would bloat and lead to some long term maintenance issues. If someone wants to use another content type, i think it makes sense to use the raw body and parse manually.

binarydud:
@jamesls i'll try taking a stab in the pr of changing the json_body property to be None if the content-type isn't application/json

binarydud:
@jamesls updated the pr to set json_body to None if the Content-Type header is not application/json

jamesls:
Thanks taking a look.

My initial reasoning for wanting helpers for parsing incoming data was to give something similar to what flask has via `request.form['myfield']`, but there's actually quite a bit of work involved to make that work properly so I'm ok with not adding that as part of this change.  It can always be added later.

PR #120, opened 2016-09-14
Prevent trailing whitespace in policy.json.

jarpy:
Python's `json.dumps` function places trailing whitespace on some lines when
`indent` is specified. Among other things, this can cause spurious diffs
when a correctly formatted, non-autogen policy is processed for deployment.

Adding an appropriate `seperators` argument suppresses the trailing
whitespace.

REF: https://docs.python.org/2.7/library/json.html


jamesls:
Thanks for the pull request.

Issue #103, opened 2016-08-16
Deploying Error (Error: Error when deploying:)


aidanoldershaw:
For a while, every other time I tried to deploy, I received the following error:

`Error: Error when deploying:`

No additional information was given. Occasionally, a different error message would appear, e.g. 

`Error: Error when deploying: list index out of range`

but deployment would generally work the second time around. Now, however, I get nothing but the first error, and cannot deploy at all. The contents of my config.json file are as follows:
```
{
  "manage_iam_role":false,
  "iam_role_arn": "arn:aws:iam::...:role/lambda_dynamo",
  "lambda_arn": "arn:aws:lambda:us-west-2:...:function:chalice_api",
  "app_name": "chalice_api",
  "stage": "dev"
}
```

kishinmanglani:
Running into the same issue

jamesls:
I'm trying to repro the issue but I'm not able to.  Do you have a specific set of steps that show the issue?  

Here's what I tried:

```
chalice new-project foo
chalice deploy
# (update the .chalice/config.json to match what you have, except I filled in my own iam_role_arn and lambda_arn)
chalice deploy
```

I'll see about adding more debug info on the error cases, but if someone can provide a set of steps that demonstrate the issue this would really help in resolving the issue.

kishinmanglani:
This sort of fixed itself for me. A little while after commenting it worked. If I encounter the issue again, I'll try to figure out steps to repro

Prefinem:
Also ran into this issue

    Initial creation of lambda function.
    Updating IAM policy.
    Creating deployment package.
    Error: Error when deploying:

Merlineus:
I have the same issue
`
Initial creation of lambda function.
Updating IAM policy.
Creating deployment package.
Error: Error when deploying:
`

hallazzang:
I tried to fix this issue, please check #124

Prefinem:
@hallazzang It works now on my system.  Thanks!

jamesls:
#124 has been merged, closing out issue.

Issue #31, opened 2016-07-12
running deploy on Windows raises "assert os.path.isfile(pip_exe)"


rupello:
I'm following the tutorial instructions and getting the error below:

> chalice deploy                                                                                                            
Initial creation of lambda function.                                                                                        
Updating IAM policy.                                                                                                        
Creating deployment package.                                                                                                
Traceback (most recent call last):                                                                                          
  File "C:\Python27\Lib\runpy.py", line 162, in _run_module_as_main                                                         
    "__main__", fname, loader, pkg_name)                                                                                    
  File "C:\Python27\Lib\runpy.py", line 72, in _run_code                                                                    
    exec code in run_globals                                                                                                
  File "C:\Users\rlloyd\venvs\chalice-demo\Scripts\chalice.exe\__main__.py", line 9, in <module>                            

...snipped...

_first_time_lambda_create   
    config['project_dir'])                                                                                                  
  File "c:\users\rlloyd\venvs\chalice-demo\lib\site-packages\chalice\deployer.py", line 633, in create_deployment_package   
    assert os.path.isfile(pip_exe)                                                                                          
AssertionError

rupello:
Here is a workaround for this

deployer.py line #632:
#pip_exe = os.path.join(venv_dir, 'Scripts', 'pip')
pip_exe = os.path.join(venv_dir, 'Scripts', 'pip.exe')

deployer.py line #644:
        #python_dir = os.listdir(os.path.join(venv_dir, 'lib'))[0]
        #deps_dir = os.path.join(venv_dir, 'lib', python_dir,
        deps_dir = os.path.join(venv_dir, 'lib', 
                                'site-packages')

jamesls:
Thanks for reporting.  I'll get this updated.

sdole:
I had the same issue. listed workaround worked for me. thanks.

jamesls:
Fixed via: https://github.com/awslabs/chalice/pull/124

PR #124, opened 2016-09-21
Fix issue #103

hallazzang:
I think the problem is related to OS.
In my Windows machine, the directory `<projectpath>/.chalice/venv` structure looks like this:

```
venv/
    include/
        ...
    Lib/
        ...
        site-packages/
            ...
        ...
    Scripts/
        ...
        pip.exe
        ...
    tcl/
        ...
    pip-selfcheck.json
```

I modified some codes to correct the paths.


jamesls:
Thanks for the pull request.  I think this will also fix #31 right?

hallazzang:
Yes, I think so :)

Issue #129, opened 2016-09-29
Content-Type lookup is case sensitive


chrisgilmerproj:
Using `fetch()` method from javascript and it does a lowercase operation before it send out my header.  That means `Content-Type` becomes `content-type`.

This problem was noted in the code already:

https://github.com/awslabs/chalice/blob/master/chalice/app.py#L72-L75

Turns out this problem happened recently with the addition of 1a2ae877e0708be8a046c90e194908f414adb1fa.

I suggest just checking for both:

```py
if self.headers.get('Content-Type') == 'application/json' or \
        self.headers.get('content-type') == 'application/json' :
    self.json_body = body
```

Headers should be case-insensitive, so perhaps the class should lowercase all the headers as it reads them?

jamesls:
Closing, https://github.com/awslabs/chalice/pull/130 has been merged.

chrisgilmerproj:
Thanks so much for the work on this!

PR #130, opened 2016-09-30
Content type lookup with lowercased key

hadrien:
To address issue #129

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -57,13 +57,32 @@ ALL_ERRORS = [
     TooManyRequestsError]
 
 
+class CaseInsensitiveMapping(object):
+    """Case insensitive and read-only mapping."""
+
+    def __init__(self, mapping):
+        self._dict = {k.lower(): v for k, v in mapping.iteritems()}
+
+    def __getitem__(self, key):
+        return self._dict.__getitem__(key.lower())
+
+    def __contains__(self, key):
+        return self._dict.__contains__(key.lower())
+
+    def get(self, key, default=None):
+        return self._dict.get(key.lower(), default)
+
+    def __repr__(self):
+        return repr(self._dict)
+
+
 class Request(object):
     """The current request from API gateway."""
 
     def __init__(self, query_params, headers, uri_params, method, body,
                  base64_body, context, stage_vars):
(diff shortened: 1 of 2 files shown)


chrisgilmerproj:
This looks great.  Thanks for the quick turnaround on it.

chrisgilmerproj:
@hadrien - any chance this will get added and released this week?  I don't mean to pressure I'd just like to fix my app so I can deploy it.

hadrien:
@chrisgilmerproj I am not a part of aws so I cannot give you any detail, I do not have merging permission: Actually, I need that fix too :-)

@jamesls Whenever you have the time to give feedback, it will be super appreciated, thanks!

chrisgilmerproj:
Oh! Interesting.  Well I appreciate the work @hadrien - thank you very much for your time on this.  It's going to be a big help for me and I hope others as well.

jamesls:
Thanks for the pull request.  In general I think this is a good change to add, the only problem I see is that this line:

```
self.headers = {k.lower(): v for k, v in headers.items()}
```

means that anyone previously relying on upper/title cased header names will now be broken.  Given the preview status of this project, this is still ok, but I wonder if it makes more sense to just have a case insensitive dictionary used for headers instead.

Thoughts?

hadrien:
@jamesls Totally agree, I was being lazy in the first place. 

Please give me feedback on my last commit. 

When/If PR is approved, should I squash commits to make a cleaner history?

jamesls:
No need to squash, I'll go ahead and merge.  Thanks for updating.

jamesls:
Quick update, I had to make a couple of small changes to this.  I want to preserve the property that `app.current_request.to_dict()` will always return JSON serializable data so I made a small update to the `to_dict()` method.

I wasn't able to `dict(request['headers'])` so I updated the code to subclass from `collections.Mapping` to make this possible.  Otherwise everything else is unchanged.

I also added a few tests to catch this in the future (the fact that `to_dict()` is json serializable).

jamesls:
Rebased/merged in https://github.com/awslabs/chalice/commit/d69a2481a918f01a3926ea224e42f6e9f02faf6b

PR #141, opened 2016-10-17
Deploy with --no-autogen-policy errors with EntityAlreadyExists message

tiyberius:
I found that I had over 100 roles in my account, and the Boto list_roles() call that defaults to only returning 100 results, which caused Chalice to not think that the role existed. Therefore Chalice would try to create it and that's when the error happened. These changes should address that issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ sample/
 .cache
 venv
 docs/build/
+.idea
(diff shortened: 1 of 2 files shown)


tiyberius:
I fixed a minor typo and updated the branch in my fork, how do we update this pull request with that update?

jamesls:
Thanks for the bug fix.

After looking at this, I'm not actually sure if we need to just `list_roles()` at all (I don't remember the rationale for using `list_roles()` initially.  I think I can just use `get_role()` directly.  This should avoid pagination altogether and speed things up.

tiyberius:
Oh duh, why didn't I think of that! `get_role()` is so much better

PR #146, opened 2016-10-25
Initial support for multi file apps

jamesls:
There's been several approaches proposed:

1. Support either 'app.py' *or* an ``app/`` directory.
2. Recursively find all python modules/packages (similar to what
   ``find_packages(.)`` does.
3. Have some type of manifest file that includes/excludes files.
4. Include a hard coded directory if it exists.

This commit implements option 4, which I think has a good balance
of control vs. ease of use:

* If you want something included in the deployment package, put it
  in this directory.  This works whether it's a python file or not.
* If you don't something included in your deployment package, don't
  include it in ``chalicelib/``.

cc @kyleknap @JordonPhillips

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,10 @@ check:
 	# Basic error checking in test code
 	pyflakes tests/unit/ tests/functional/
 	##### DOC8 ######
-	# Correct rst formatting for docstrings
+	# Correct rst formatting for documentation
 	#
 	##
-	doc8 docs/source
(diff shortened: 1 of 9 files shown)


jamesls:
Docs/tests updated.  I haven't any feedback on this PR nor on the original issue, so let's try this approach out for now.  I imagine we may need to iterate on this, so I'd like to get this initial version available for people to try out.

nkconnor:
I don't see any info on this in the README

Issue #159, opened 2016-11-09
Return 405 when a method is unsupported


grahamlyons:
As per RFC 7231 [rfc7231#section-6.5.5](https://tools.ietf.org/html/rfc7231#section-6.5.5) we should send a `405` status code rather than `500` when an unsupported method is used to access a resource.

jamesls:
https://github.com/awslabs/chalice/pull/160 has been merged.

PR #160, opened 2016-11-09
Return 405 Method Not Allowed

grahamlyons:
Addresses issue #159 to return a `405` status code when the method is not supported by the route.


jamesls:
Thanks for the pull request.  Looks like there's a failing test so I'll go ahead and fix that up before merging.

PR #161, opened 2016-11-09
Support adding VPC functionality now that Lambda can run in VPC

sellers:
Enable config.json to support new list options for 

```

vpc_subnets: ["subnet-xxx", "subnet-yyy"]

security_groups: ["sg-xxx", "sg-yyy"]
```

update the views to support being able to pass in the subnets. 

Please review and use features as fit.   Missing = unit tests.


iw:
Hi, Is the `VPC` `config` support still intended to be released?

PR #168, opened 2016-11-14
Add policy mapping for ElasticSearch

shin1985:
http://boto3.readthedocs.io/en/latest/reference/services/es.html?highlight=es#id1

`boto 1.4.1` and higher has `ElasticSearch` support as well.
An operation was requested that requires `es` support to be policy allowed in, but `policies.json` was configured without es support.

### DIFF
--- a/chalice/policies.json
+++ b/chalice/policies.json
@@ -326,6 +326,23 @@
     "UnassignPrivateIpAddresses": "ec2:UnassignPrivateIpAddresses",
     "UnmonitorInstances": "ec2:UnmonitorInstances"
   },
+  "es": {
+    "ESHttpDelete": "es:ESHttpDelete",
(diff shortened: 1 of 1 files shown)


chalice/policies.json
```diff
@@ -326,6 +326,23 @@
     "UnassignPrivateIpAddresses": "ec2:UnassignPrivateIpAddresses",
     "UnmonitorInstances": "ec2:UnmonitorInstances"
   },
+  "es": {
+    "ESHttpDelete": "es:ESHttpDelete",
```
jamesls: How are these used?  The ES* keys don't map to boto3 operations so they'll never get triggered during policy generation.  Can we remove these?

jamesls:
I went ahead and just removed the data plane policy actions for now.  Otherwise merging.  Thanks for the pull request.

PR #170, opened 2016-11-15
Fix bug for finding root resources if multiple already exist

michael-dev2rights:
I'm experimenting with using chalice as a library.  I found that if there are multiple resources already created then get_root_resource_for_api may not actually return the root resource. 

This fixes that.  I haven't shown that it fixes any particular bug in chalice as an application.


jamesls:
Thanks for the pull request.  Looks like the travis build is failing for this, so we'll need to get those tests updated.  I'd also like to get a test added for this bug fix to ensure there's no future regressions.

andriisoldatenko:
I think we can close this PR because of #205 has been merged.

PR #171, opened 2016-11-15
Make creation not fail even if things already exist.  

michael-dev2rights:
I want to use bits of chalice as a library which would allow deploying parts of an API even if other parts are already deployed.  

In order to work in this direction I made a patch which makes the actual API creation functions run inside a try/except.  Currently it's a bit of a hack which I'm showing primarily to get comments.  If you would be interested to merge then I would be willing to clean this up as suggested.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ docs/build/
 .idea
 *.py?
 __pycache__/
+*.pyc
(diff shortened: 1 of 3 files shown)


jamesls:
Closing out old PR.  The internals have changed a bit so we now use swagger to generate the API, and we no longer use the low level `create_rest_resource()` API calls.

PR #174, opened 2016-11-16
Introduce a 204 No Content handler

WrathOfChris:
Needed to return a 204 NO CONTENT from a Chalice API.  This creates a `SuccessNoContent` exception that returns a `ChaliceViewError` but with STATUS_CODE set to 204.

This is not likely a sustainable design for extending chalice, however for this specific use without any content returned, it works.

### DIFF
--- a/chalice/__init__.py
+++ b/chalice/__init__.py
@@ -1,7 +1,7 @@
 from chalice.app import Chalice
 from chalice.app import (
     ChaliceViewError, BadRequestError, UnauthorizedError, ForbiddenError,
-    NotFoundError, ConflictError, TooManyRequestsError
+    NotFoundError, ConflictError, TooManyRequestsError, SuccessNoContent
 )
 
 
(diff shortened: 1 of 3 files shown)


jamesls:
Thanks for the pull request.  Just wanted to comment on this to let you know I'm aware of this feature gap and it's something I want to address.

I think you make a good point that the current approach isn't sustainable.  I think I can address this more generally by switching to a [lambda proxy integration](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html) and removing all the client side VTL code.  If this works out, then you'd be able to return and arbitrary status code and headers.

WrathOfChris:
Thanks for the reply, and a great talk at re:invent - if I hadn't had a plane to catch, I would have stayed until the end to say hi in person.

No worries on not merging this PR - for me it solved a specific integration problem in a proof of concept app, but I knew it wasn't likely to work beyond this quick fix.

Can't wait to see what you have in store with lambda-proxy!

PR #181, opened 2016-11-28
Add before_response and after_response decorators

kislyuk:
@jamesls, I'm wondering if this is something you would be open to copying from Flask, I find myself needing these a lot.

If you're open to this in principle, I can add tests and docs.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -281,8 +281,8 @@ class Chalice(object):
                     break
             else:
                 response = view_function(*function_args)
-            for postproc_func in self.after_request_funcs:
-                response = postproc_func(response)
+                for postproc_func in self.after_request_funcs:
+                    response = postproc_func(response)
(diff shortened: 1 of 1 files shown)


jamesls:
I'm certainly open to this feature.  I think this could be really useful.

I always like hearing a few concrete use cases.  Would you mind sharing some of the specific things this would help you with?

kislyuk:
I use it for code that is shared between all or many routes: authentication/authorization (without API Gateway custom authenticators, which are complex to set up and seem to add latency), common parameter/header parsing and erroring, logging, etc. It can also be useful for some types of rate limiting/backpressure, etc.

jmehnle:
@jamesls, between this and #719, what are your thoughts?

jamesls:
Closing out old PR, superceded via https://github.com/aws/chalice/pull/1514

PR #184, opened 2016-11-29
Parse and pass through query string params in local mode

kislyuk:


uncompiled:
I hope this gets merged in as I just came across this issue.

```
File ".../chalice/local.py", line 54, in match_route
    raise ValueError("No matching route found for: %s" % url)
```

jamesls:
Thanks for the fix.  I'm just going to add a quick unit test for this before merging to ensure we don't regress on this in the future.

Issue #180, opened 2016-11-25
Doesn't detect JSON when Content-Type contains charset


arundhaj:
The app doesn't recognize the JSON content when the header is `Content-Type: application/json;charset=UTF-8`

jamesls:
Fixed via #185.

PR #185, opened 2016-11-30
fix for content-type that contain charset

arundhaj:
Fix for the issue [#180](https://github.com/awslabs/chalice/issues/180) that contains charset in `content-type` header

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -96,7 +96,7 @@ class Request(object):
         #: The parsed JSON from the body.  This value should
         #: only be set if the Content-Type header is application/json,
         #: which is the default content type value in chalice.
-        if 'application/json' in self.headers.get('content-type'):
+        if 'application/json' in self.headers.get('content-type', ''):
             self.json_body = body
         else:
             self.json_body = None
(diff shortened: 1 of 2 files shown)


arundhaj:
I found a bug when the API has no header. Update the condition to fix it.

       if 'application/json' in self.headers.get('content-type', {}):

tadashi-aikawa:
I encountered this problem, too.
I hope this pull request will merge into master.

laurrentt:
Same here, I tried too hook my chalice app with OpsGenie but they send `'application/json;charset=UTF-8'` as `content-type`.

jamesls:
Thanks for the pull request.  I'm going to add a test for this first and then I'll go ahead and merge.

Looking at the grammar for [content-type](https://tools.ietf.org/html/rfc7231#section-3.1.1.1):

```
  media-type = type "/" subtype *( OWS ";" OWS parameter )
     type       = token
     subtype    = token

   The type/subtype MAY be followed by parameters in the form of
   name=value pairs.

     parameter      = token "=" ( token / quoted-string )
```

I think we can tighten this up to be `.startswith('application/json')` instead of `in`.

PR #186, opened 2016-11-30
Local development server to log on stdout

arundhaj:
When running chalice local server, this change would enable the app to log in stdout.

The following code that logs in CloudWatch when running on AWS Lambda, will also work during local.

```
import logging

logger = logging.getLogger()
logger.setLevel(logging.INFO)

@app.route('/')
def index():
    logger.info('Log in stdout when local and in CloudWatch on AWS Lambda')
    return {'hello': 'world'}
```

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -5,6 +5,8 @@ This is intended only for local development purposes.
 """
 import json
 import functools
+import logging
+import sys
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the pull request.  I made one small change and moved this logic out of module scope and into the `local` function in `chalice.cli`.  Otherwise looks good.

PR #187, opened 2016-12-01
Add support for serializing decimals

jamesls:
Lambda supports serializing decimals to json,
so we need to support that as well in chalice local.

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -4,6 +4,7 @@ This is intended only for local development purposes.
 
 """
 import json
+import decimal
 import functools
 from collections import namedtuple
 from BaseHTTPServer import HTTPServer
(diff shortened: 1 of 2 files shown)


arundhaj:
Can we support datetime serialization as well?


```
from datetime import datetime
```

Append the following
```
if isinstance(obj, decimal.Decimal):
    return float(obj)
elif isinstance(obj, datetime):
    f = '%Y-%m-%d %H:%M:%S'
    return o.strftime(f)
```

jamesls:
This PR is just for `chalice local`, so I'd prefer to keep parity with what lambda actually does (which doesn't serialize datetimes).  Otherwise you'd have behavior where it works in `chalice local` but fails when you deploy.

PR #188, opened 2016-12-03
Fix not all resources being retrieved due to pagination

wesleysee:
get_resources by default only retrieves 25 resources per API call, thus
not all resources are returned if the number exceeds this amount.

This adds pagination handling, as well as bumping the limit to 100, such
that less API calls are needed if necessary.


jamesls:
Thanks for the pull request.  Botocore has a useful [paginator](http://botocore.readthedocs.io/en/latest/topics/paginators.html) feature that can take care of all the details for paginating a result.  I updated the code to use the paginator interface.

Otherwise looks good.

Issue #192, opened 2016-12-06
Add support for optional deployment options


wscoble:
Specifically, I would like to see deployment options in the config.json file to handle timeout and memory size.

I will work on a PR for this feature request as well.

wscoble:
Actually, this seems to be supported, just not documented yet.

kyleknap:
Support for configuration of lambda function timeout and memory size was completed in this merged PR: https://github.com/awslabs/chalice/pull/351 under this proposal: https://github.com/awslabs/chalice/issues/347. Closing out issue.

PR #193, opened 2016-12-06
add MemorySize support in config and aws client, with tests

wscoble:
Referencing #192 to add memory size option in config with 128 default as well as for function create and update. I added a documentation entry that's probably not perfect, but should suffice for now.

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -44,8 +44,9 @@ class TypedAWSClient(object):
             raise
         return True
 
-    def create_function(self, function_name, role_arn, zip_contents):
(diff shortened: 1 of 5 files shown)


wscoble:
I agree with you @ButtaKnife about making it more flexible with kwargs. There is a lot of opportunity in the aws client in chalice for this as well.

wscoble:
@ButtaKnife I actually have something working that is a bit more full-featured and includes all the boto3 parameters for create_function and update_function_configuration. I will open a new PR with that code once I have the tests fleshed out.

PR #195, opened 2016-12-07
Add support for more lambda options in config.json and allow config changes between deploys

wscoble:
Updated config to include a `lambda` section that contains boto3 options for create_function and update_function_configuration. I did not write any documentation yet, but if this solution is something that is wanted, I will make the effort to do the documentation.

Ref: #193 PR comments

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -45,7 +45,7 @@ class TypedAWSClient(object):
         return True
 
     def create_function(self, function_name, role_arn, zip_contents, config):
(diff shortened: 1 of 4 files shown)


wscoble:
Sorry about the many commits, guess its time for some sleep.

artkay:
+1

jamesls:
Looks like this was addressed in https://github.com/awslabs/chalice/issues/347.  Let us know if there's anything else missing you'd like to configure.

Issue #196, opened 2016-12-11
Passing query params breaks route in local


atomaka:
When using `chalice local`, a route will not match if parameters are passed via the query string.

Given:
```
from chalice import Chalice

app = Chalice(app_name='web')

@app.route('/')
def index():
    name = app.current_request.query_params.get('name')
    return {'hello': name}
```

Running `chalice local` and then visiting `http://localhost:8000/?name=test` will result in `ERR_EMPTY_RESPONSE` with an error tracing back to `raise ValueError("No matching route found for: %s" % url)`.

Running `chalice  deploy` and then visiting the provided URL functions as expected and will return `{"hello": "test"}`.

jamesls:
Fixed via https://github.com/awslabs/chalice/pull/184

PR #197, opened 2016-12-12
Accept query parameters when running locally

atomaka:
In #196, I  advised that query string parameters do not work when routing locally.

I believe `urlparse.parse_qsl` emulates the behavior of the AWS API Gateway by only accepting the last query value passed if multiple matching keys exist.

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -5,6 +5,7 @@ This is intended only for local development purposes.
 """
 import json
 import functools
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks for the pull request.  This should be fixed now via (https://github.com/awslabs/chalice/pull/184).  Closing as dupe.

PR #204, opened 2016-12-28
Check if runtime is Python2

andriisoldatenko:
Add ability to check python runtime

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from setuptools import setup, find_packages
-
+import sys
 
 with open('README.rst') as readme_file:
     README = readme_file.read()
@@ -13,6 +13,8 @@ install_requires = [
     'typing==3.5.2.2',
 ]
 
+if sys.version_info < (3, 0):
+    raise RuntimeError("chalice requires only Python 2.7")
 
 setup(
     name='chalice',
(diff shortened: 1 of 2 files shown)


jamesls:
I think this can still work if I just skip the `pip install -e` on python3.  I'll update this before merging.

andriisoldatenko:
will update.

jamesls:
@andriisoldatenko Thanks!  To get the mypy stuff working, we can do something like this for now:

```
  - if [[ $TRAVIS_PYTHON_VERSION != 3.* ]]; then pip install -e .
  - if [[ $TRAVIS_PYTHON_VERSION != 3.* ]]; then travis_retry pip install -r requirements-dev.txt
  - if [[ $TRAVIS_PYTHON_VERSION == 3.* ]]; then pip install mypy-lang==0.4.5; fi
```

Probably worth moving out to a separate shell script if this gets any more complicated, unless there's a better way to do this.

andriisoldatenko:
@jamesls  @dstufft thanks guys for suggestions.

setup.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from setuptools import setup, find_packages
-
+import sys
```
dstufft: This import is harmless but unused.

jamesls:
Thanks!

PR #214, opened 2017-01-01
Added python 3.6 to tox file

andriisoldatenko:


andriisoldatenko:
@jamesls Please review.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, py33, py34, py35
+envlist = py26, py27, py33, py34, py35, py36
```
andriisoldatenko: I know, but IMHO code should be compatible with python 3 to better switching.
But if you want I can remove this file.

gliptak:
I found this reference allowing to run Python 3.x

http://www.cloudtrek.com.au/blog/running-python-3-on-aws-lambda/

andriisoldatenko:
@gliptak It's known workaround :)

jamesls:
Hi, thanks for the pull request.  I'm going to go ahead and remove this tox file for now.  While I agree that we should probably just update chalice to support python2/3, there's a bit of dev work involved to make that happen.  But once that does happen we can add back the tox.ini.

Issue #231, opened 2017-01-27
Can Chalice handle Cloudwatch scheduled events?


carlkibler:
Primary question: Is there a trick within Chalice to set a fallback handler with no expectations of a URL or content type, to handle scheduled activations?

--
Background:
I have several endpoints in Chalice all working well to send and receive text messages. I have set up a Cloudwatch Scheduled Event that can run Lambda functions regularly. Chalice does not handle these well, or I haven't found a way to do it at least.

The `chalice logs` shows this for every invocation:
```
2017-01-27 01:11:27.412000 90b269 Unknown request. (Did you forget to set the Content-Type header?): ChaliceError
Traceback (most recent call last):
  File "/var/task/chalice/app.py", line 236, in __call__
    "Unknown request. (Did you forget to set the Content-Type "
```

The scheduled trigger has no options to set inputs to the Lambda function, which conceptually doesn't match how Chalice's routing expects to work, thus it falls back to the above error.

So my question is:
Is there a trick within Chalice to set a fallback handler with no expectations of a URL or content type, to handle scheduled activations?

If not, would that be a valid feature request for Chalice? Today, it looks like I will have to separately write a regular Lambda function and manage it manually, separate from my main Chalice project.

carlkibler:
I decided to play with this more. I added logging to the __call__ method and have the regular Lambda event and context data. 
event:
```
{u'account': u'804035054667', u'region': u'us-west-2', 
u'detail': {}, u'detail-type': u'Scheduled Event', u'source': u'aws.events', 
u'version': u'0', u'time': u'2017-01-27T20:02:26Z', 
u'id': u'0cfcac58-adfd-4e03-a4aa-f22b5c3c49f9', u'resources': [u'arn:aws:events:us-west-2:804035054667:rule/10min-scheduler']}
```

So the source and detail-type field could be the hook to redirect the Lambda activation to a separate function. I think another decorator would be appropriate for handling scheduled events.

Single handler by default:
```
@app.scheduled
def do_my_thing():
    <code>
```

And optional ability to handle specific named scheduled rules:
```
@app.scheduled('10min-scheduler')
def do_my_thing():
    <code>
```

carlkibler:
Small update: I have this working in my fork. Will file pull request shortly.

jamesls:
This is an interesting one.

There's been several requests from people to not require API gateway, and/or pair lambda with something else.  Cloudwatch events, Dynamodb streams, kinesis, etc.

[...]

PR #234, opened 2017-02-07
Add @schedule decorator for CloudWatch-triggered events

carlkibler:
First pass at solving Issue #231. It adds a new decorator that lets the user specify functions as the handler for CloudWatch timed alerts. Chalice can send handle periodic events at intervals of 1 minute or greater with crontab fidelity.

1. The major code change is in __call__() method to detect one of the properties specific to these events and setting default values for some of the properties used to create the Request object. 
2. Rest of code is copy of route bookkeeping code.

If this pull request can be accepted (with whatever code review changes requested) I will write documentation and more tests.

CloudWatch allows quite a few more rules based on pattern matching of events, but I was focused on (what I assume) is the most valuable use for Chalice: timed events.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -241,7 +241,7 @@ class Chalice(object):
         content_types = ['application/json']
         cors = False
         if kwargs:
-            raise TypeError('TypeError: route() got unexpected keyword '
+            raise TypeError('TypeError: schedule() got unexpected keyword '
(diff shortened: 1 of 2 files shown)


stealthycoin:
Cloudwatch events support was merged in according to [this proposal](https://github.com/awslabs/chalice/issues/390) so I am going to close this PR.

Issue #249, opened 2017-03-03
Try to autoinstall manylinux1 wheels when possible


jamesls:
In order to create a [lambda deployment package](http://docs.aws.amazon.com/lambda/latest/dg/lambda-python-how-to-create-deployment-package.html), chalice currently will:

* Create a virtualenv
* pip install your requirements files into the virtualenv
* copy over the chalice app, `chalicelib`, and `vendor` directory into site-packages
* zip up the site-packages directory

This process works as long as all the packages are pure python packages.  If you specify something like `numpy`, we'll install the packages based on whatever platform you're running on, which is problematic.  For example, on a Mac, we'll install the `numpy` binaries for mac, zip them up and send them over to lambda, and `import numpy` will fail when the lambda function is invoked.

We've received a number of issues about this:

* https://github.com/awslabs/chalice/issues/239
* https://github.com/awslabs/chalice/issues/105
* https://github.com/awslabs/chalice/issues/106

What I'd propose is a best-effort attempt at using manylinux1 wheels:

* For each package, if there's a manylinux1 wheel available, then use it
* Otherwise if there's a suitable non-platform specific wheel, then use it
* Otherwise if it's a pure python package, just fall back to the normal sdist install
* Otherwise if it requires C extensions and we don't have a manylinux1 available, fail (or maybe just warn).

I'm not sure how feasible it is to implement the above logic, but that would be my ideal behavior.

PR #258, opened 2017-03-18
Add 'package' command and SAM support

jamesls:
This adds support for a 'package' command.  This command takes the
current application and produces artifacts that can subsequently be
deployed by cloudformation.  The intent is to allow for decoupling the
packaging from the deployment process, which provides a number of
benefits.  First, something besides chalice can do the deployment.  For
many customers that have existing deployment tooling based around
cloudformation, they would be able to leverage these tools to handle
deployment.  In addition to existing customer tooling, this also makes
it possible to integrate with other AWS services.  By creating a
separate app package, chalice can now integrate with AWS CodePipeline.
This allows a user to set up a full CD pipeline that deploys your
chalice app whenver you 'git push' to a repo.

The second benefit is that by decoupling the packaging from the
deployment, this allows a user to modify the app package before
deploying.  While chalice strives to expose APIs that make app
configuration simple, the user can now have a system in place that
packages the app, modifies the app package, and then deploys the app.

Missing pieces:

* Updating the 'deploy' command to be able to deploy an app package.

NOTE: This is branched off of `add-authorizers` so it includes that PR as well (https://github.com/awslabs/chalice/pull/256)


chalice/deploy/packager.py
```diff
@@ -0,0 +1,213 @@
...
+        # The original implementation used Popen(['virtualenv', ...])
+        # However, it's hard to make assumptions about how a users
+        # PATH is set up.  This could result in using old versions
+        # of virtualenv that give confusing error messages.
+        # To fix this issue, we're calling directly into the
+        # virtualenv package.  The main() method doesn't accept
+        # args, so we need to patch out sys.argv with the venv
+        # dir.  The original sys.argv is replaced on exit.
+        original = sys.argv
+        sys.argv = ['', venv_dir, '--quiet']
```
dstufft: Why not use ``virtualenv. create_environment()`` instead of monkeypatching ``sys.argv``?

chalice/deploy/packager.py
```diff
@@ -0,0 +1,213 @@
...
+        deps_dir = compat.site_packages_dir_in_venv(venv_dir)
+        assert os.path.isdir(deps_dir)
+        # Now we need to create a zip file and add in the site-packages
+        # dir first, followed by the app_dir contents next.
+        dirname = os.path.dirname(os.path.abspath(package_filename))
+        if not os.path.isdir(dirname):
+            os.makedirs(dirname)
+        with zipfile.ZipFile(package_filename, 'w',
+                             compression=zipfile.ZIP_DEFLATED) as z:
+            self._add_py_deps(z, deps_dir)
```
dstufft: This is going to fail for any C package unless the host OS is on a compatible with Lambda Linux right? Would it be better to instead do something like:

```
$ pip wheel -r requirements.txt
$ # Delete any wheel files that are not pure Python, recording the names of the files deleted.
$ pip download --only-binary=:all: --platform manylinux1_x86_64 --no-deps <the list of names from above>
$ # Fail if there are any downloads that didn't get satisified
```

Once you have a list of wheels, it's not particularly hard to "install" them manually into a zip file for unpacking later.

The advantage of this is that it will generally allow binary packages (where possible), but will fail loud and obviously when those binary packages do not exist. You could possibly do ``CC=/bin/false`` when running ``pip wheel`` which will force packages with optional C extensions to fail at building (typically), but not all of them will actually produce a pure python wheel (e.g. PyYAML produced a wheel that internally is pure python, but whose filename still marks it as binary).

chalice/deploy/packager.py
jamesls: Thanks for the suggestion, I wasn't aware of `create_environment()`, that's certainly preferable to patching out sys.argv.  I'll give this a shot.

chalice/deploy/packager.py
jamesls: I think this would definitely be better.  That work is tracking in #249.

I should have clarified in the diff, but this `packager.py` code is all pre-existing, but is showing up as new code because I restructured some of the files.  I definitely want to circle back on this, a lot of this is old code that could use some rework.

dstufft: Ah, gotcha. That makes sense then.

PR #260, opened 2017-03-23
Fixed position of `local` in `chalice local` example

dlemphers:
The position of the port (--port) argument when starting chalice in local mode was incorrect.


jamesls:
Thanks!

PR #261, opened 2017-03-27
Add --single-file to package command

jamesls:
This allows the package command to produce
a single file for all the packaged assets instead
of an output directory.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -8,6 +8,8 @@ import json
 import logging
 import os
 import sys
+import tempfile
+import shutil
 
 import botocore.exceptions
 import click
@@ -22,6 +24,8 @@ from chalice.config import Config
 from chalice.deploy import deployer
 from chalice.logs import LogRetriever
 from chalice.package import create_app_packager
+from chalice.utils import create_zip_file
+
 
 TEMPLATE_APP = """\
 from chalice import Chalice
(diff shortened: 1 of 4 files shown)


jamesls:
Working on the codecov now.

chalice/cli/__init__.py
```diff
@@ -283,13 +287,29 @@ def generate_sdk(ctx, sdk_type, outdir):
...
 @cli.command('package')
-@click.argument('outdir')
+@click.option('--single-file', is_flag=True,
+              default=False,
+              help=("Create a single packaged file. "
+                    "By default, the 'out' argument "
+                    "specifies a directory in which the "
+                    "package assets will be placed.  If "
+                    "this argument is specified, a single "
+                    "zip file will be created instead."))
```
kyleknap: May want to include information on when to use it. Was not completely obvious on why it should be used.

chalice/utils.py
```diff
@@ -1,8 +1,29 @@
...
+    specified by the `outfile` argument.
+
+    """
+    with zipfile.ZipFile(outfile, 'w',
+                         compression=zipfile.ZIP_DEFLATED) as z:
+        prefix_len = len(source_dir) + 1
+        for root, _, filenames in os.walk(source_dir):
+            for filename in filenames:
+                full_name = os.path.join(root, filename)
+                archive_name = full_name[prefix_len:]
```
kyleknap: Why not use ``os.path.relpath`` with the ``source_dir`` as the starting point?

chalice/utils.py
```diff
@@ -1,8 +1,29 @@
...
+    # type: (str, str) -> None
+    """Create a zip file from a source input directory.
+
+    This function is intended to be an equivalent to
+    `zip -r`.  You give it a source directory, `source_dir`,
+    and it will recursively zip up the files into a zipfile
+    specified by the `outfile` argument.
+
+    """
+    with zipfile.ZipFile(outfile, 'w',
```
kyleknap: Nice. Don't have to worry about python2.6's lack of context managers.

chalice/cli/__init__.py
jamesls: I think it will be more useful once I add the `chalice deploy --package` flag.  I'll update these docs once I add support for that command, right now there's not much of an advantage to using this.

jamesls:
@kyleknap I believe I've incorporated all the feedback.

chalice/cli/__init__.py
kyleknap: That makes sense.

PR #264, opened 2017-03-28
Introduce completely separate chalice stages

jamesls:
This introduces a change to how stages work in chalice.
Previously a "stage" in chalice corresponded to an API gateway
stage.  While this could theoretically work for lambda functions
as well (via aliasing and function versions) it would not necessarily
scale with other AWS resources.  Now, chalice "stages" are
entirely new sets of resources.  For example, if I run:

```
$ chalice deploy dev  # Note 'dev' can be omitted, it's the default
$ chalice deploy prod
```

I will have two completely separate sets of API gateway rest APIs,
Lambda functions, and IAM roles.
To help track this, a ".chalice/deployed.json" is written on deploys
that contains the deployed values per-stage.

There's still some additional work remaining, primarily against
upgrading from old versions to this new version.  We should be warning
or erroring out in that scenario, but that will be addressed as a
separate commit.

Additionally, the "chalice url/logs" command can be updated to use
the ".chalice/deployed.json" file as well, but that will be added as
a separate commit.

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -20,8 +20,7 @@ import zipfile
 import shutil
 import json
 
-import botocore.session
-import botocore.exceptions
+import botocore.session  # noqa
 from typing import Any, Optional, Dict, Callable, List  # noqa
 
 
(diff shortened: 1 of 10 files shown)


chalice/awsclient.py
```diff
@@ -138,6 +138,17 @@ def get_rest_api_id(self, name):
                 return api['id']
         return None
 
+    def rest_api_exists(self, rest_api_id):
+        # type: (str) -> bool
+        """Check if an an API Gateway REST API exists."""
+        try:
+            self._client('apigateway').get_rest_api(restApiId=rest_api_id)
+            return True
+        except botocore.exceptions.ClientError as e:
```
jamesls: Yeah I'll get these updated, there's a few places I'm catching `ClientError` that can be updated.  I'll just need to bump the min botocore version to 1.5.0.

recursify:
@jamesls - thanks for doing this!  Chalice seemed promising, but was a non-starter unless it had full dev/prod/test environments.  I'll check out the `package` branch and give it a whirl.

PR #267, opened 2017-03-29
Fix use of markdown in RST docs

gene1wood:
There are instances of use of Markdown syntax in the RST docs. This fixes those.


jamesls:
Thanks for the fix.

PR #270, opened 2017-03-30
Add --stage argument to chalice commands

jamesls:
This pr builds on https://github.com/awslabs/chalice/pull/264
to provide a consistent CLI experience when working with chalice stages.

There's several changes in this PR.

First, the code now clearly indicates whether we mean "chalice stage", which is
a completely separate collection of AWS resources, or an "API Gateway stage"
which is a specific configuration within a chalice stage (it's essentially the
URL prefix to use for your REST API for now).  `chalice_stage_name` is
generally used when we mean a chalice stage and `api_gateway_stage` is used
when we mean the latter, and with the exception of the actual CLI option
(`--stage`) every existing reference to a "stage" has been renamed to one
of those two values.

Next, I've deprecated the form "chalice deploy <stage>".  You now get an error
message if you use this form which instructs you use to use
`--api-gateway-stage`:

```
$ chalice deploy prod
You've specified a deploy command of the form 'chalice deploy <stage>'
This form is deprecated and will be removed in a future version of chalice.
You can use the --api-gateway-stage to achieve the same functionality, or the newer '--stage' argument if you want an entirely set of separate resources.
```

To get the new stage functionality introduced in #264 you need to use
`chalice deploy --stage prod`.

And finally, I've added a `--stage` option to the remaining set of chalice
commands that should be stage specific:

* url
* logs
* generate-sdk
* package

All of these commands have also been updated to pull the deployed values
from the `.chalice/deployed.json` file instead of trying to query it using
the SDK as it was doing previously.


chalice/cli/factory.py
```diff
@@ -0,0 +1,128 @@
...
+            sys.path.append(self.project_dir)
+        try:
+            app = importlib.import_module('app')
+            chalice_app = getattr(app, 'app')
+        except Exception as e:
+            # TODO: better error.
+            exception = Exception(
+                "Unable to import your app.py file: %s" % e
+            )
+            # exception.exit_code = 2
```
stealthycoin: Whats this?

chalice/config.py
```diff
@@ -76,22 +110,18 @@ def chalice_app(self):
...
-    def _chain_lookup(self, name):
-        # type: (str) -> Any
-        all_dicts = [
-            self._user_provided_params,
-            self._config_from_disk,
-            self._default_params
-        ]
+    def _chain_lookup(self, name, varies_per_chalice_stage=False):
+        # type: (str, bool) -> Any
+        all_dicts = [self._user_provided_params]
```
stealthycoin: all_dicts is probably not the best name considering that another dict is added to it if `varies_per_chalice_stage` is true. How bout something like search_dicts?

chalice/cli/__init__.py
```diff
@@ -167,32 +149,67 @@ def local(ctx, port=8000):
...
-        e.exit_code = 2
-        raise e
+def deploy(ctx, autogen_policy, profile, api_gateway_stage, stage,
+           deprecated_api_gateway_stage):
+    # type: (click.Context, bool, str, str, str, str) -> None
+    if api_gateway_stage is not None and \
+            deprecated_api_gateway_stage is not None:
+        raise _create_deprecated_stage_error(api_gateway_stage,
+                                             deprecated_api_gateway_stage)
+    if api_gateway_stage is None and deprecated_api_gateway_stage is not None:
```
kyleknap: Shouldn't this just be ``deprecated_api_gateway_stage is not None``?

chalice/cli/factory.py
jamesls: Old code I need to update.  This code was originally from the cli [__init__.py](https://github.com/awslabs/chalice/blob/master/chalice/cli/__init__.py#L101) which used click exceptions to propagate exit codes.

chalice/config.py
kyleknap: Yeah or ``cfg_dicts_chain``?

chalice/config.py
```diff
@@ -8,14 +8,70 @@
...
+        }
+
+    If the currently configured chalice stage is "dev", then
+    the config.environment_variables would be::
+
+        {"TABLE": "foo", "S3BUCKET": "devbucket"}
+
+    """
     def __init__(self,
+                 chalice_stage='dev',
```
kyleknap: There is a lot of hardcoded default ``dev`` values. Wonder if it makes sense to have it be a constant to import or required instead of optional in case it ever needs to be updated?

chalice/cli/factory.py
```diff
@@ -0,0 +1,128 @@
...
+        user_provided_params = {}  # type: Dict[str, Any]
+        default_params = {'project_dir': self.project_dir}
+        try:
+            config_from_disk = self.load_project_config()
+        except (OSError, IOError):
+            raise RuntimeError("Unable to load the project config file. "
+                               "Are you sure this is a chalice project?")
+        app_obj = self.load_chalice_app()
+        user_provided_params['chalice_app'] = app_obj
+        if autogen_policy is not None:
```
stealthycoin: Why would autogen_policy be `None`?

chalice/config.py
jamesls: Fixed

tests/functional/cli/test_factory.py
```diff
@@ -0,0 +1,71 @@
...
+    log_name = 'botocore.endpoint'
+    assert_has_no_request_body_filter(log_name)
+
+    factory.create_botocore_session(debug=True)
+
+    assert_request_body_filter_in_log(log_name)
+    assert logging.getLogger('').level == logging.DEBUG
+
+
+def test_can_create_botocoreo_session_cli_factory(clifactory):
```
kyleknap: typo botocoreo --> botocore

chalice/cli/__init__.py
```diff
@@ -13,20 +12,25 @@
...
+from chalice.config import Config  # noqa
 from chalice.logs import LogRetriever
-from chalice.package import create_app_packager
 from chalice.utils import create_zip_file, record_deployed_values
 
 
+# This is the version that's written to the config file
+# on a `chalice new-project`.  It's also how chalice is able
+# to know when to warn you when changing behavior is introduced.
+CONFIG_VERSION = '2.0'
```
kyleknap: We should maybe throw errors if we encounter a config version higher than that version of the chalice can support

chalice/cli/__init__.py
jamesls: Updated, that was before adding the mutually exclusive part validation so the first clause isn't needed anymore.

chalice/cli/factory.py
jamesls: I'm double checking on this, but I think this code is no longer relevant.

The idea a while back was to toggle between three states: not set, true, and false.  This specific check was to check if the user explicitly set a true/false value and if so then add it to the dict of user provided params.  The "not set" was needed because of config precedence.  If this option was not set, then move on to the next dict in the `_chain_lookup` of the `Config` obj.

jamesls: It looks like you could never set autogen_policy from the config file, only via the CLI.

I think the policy management and configuration could use a refresh.  This would be a fairly large change (I think) so I'll address this in a separate PR.

chalice/cli/__init__.py
```diff
@@ -201,11 +218,17 @@ def deploy(ctx, autogen_policy, profile, stage):
 @click.option('--include-lambda-messages/--no-include-lambda-messages',
               default=False,
               help='Controls whether or not lambda log messages are included.')
+@click.option('--stage', default='dev')
```
jamesls: No, not right now.  I'm hesitant to allow people to change the default to something like "prod".  I'd prefer if you're deploying from a script you have to run `--stage prod`.

chalice/cli/__init__.py
jamesls: Seems reasonable, I'll update.

chalice/cli/__init__.py
```diff
@@ -167,32 +149,67 @@ def local(ctx, port=8000):
...
+        "consider using the '--stage' argument."
+    ) % (option, positional_arg)
+    exception = click.ClickException(message)
+    exception.exit_code = 2
+    return exception
+
+
+def _warn_pending_removal(deprecated_stage):
+    # type: (str) -> None
+    click.echo("You've specified a deploy command of the form "
```
jamesls: Fixed.  I updated all the error messages to print to stderr.

chalice/config.py
```diff
@@ -8,14 +8,70 @@
...
+              }
+            }
+          }
+        }
+
+    If the currently configured chalice stage is "dev", then
+    the config.environment_variables would be::
+
+        {"TABLE": "foo", "S3BUCKET": "devbucket"}
+
```
jamesls: Yeah.  I updated the docs.

chalice/config.py
jamesls: I created a `constants.py` and moved a few of the constants over to that module.

jamesls:
@kyleknap I believe all the feedback's been incorporated.

chalice/cli/__init__.py
kyleknap: That makes sense to be explicit, especially for debugging purposes.

jamesls:
Cool, also @stealthycoin I improved the error messages on import errors:

```
 $ chalice deploy
Unable to import your app.py file:

File "/private/tmp/error/errors/chalicelib/james.py", line 1
  this is a syntax error in chalicelib/james.py

SyntaxError: invalid syntax
```

jamesls:
Rebased and merged in https://github.com/awslabs/chalice/commit/76bd8e161c024dedf8a2fae869124de599f19654

PR #271, opened 2017-04-01
Remove references to lambda_arn

jamesls:
The `config.lambda_arn` was previously used to lookup
the value of a previously deployed lambda function.
This is now handled by .chalice/deployed.json.

To accomplish this, I had to update the signature of the
API Gateway deployer to pass in the deployed lambda_arn, which
is needed for the swagger document.  I was considering using
the "existing resources" object, but it's not clear if allowing
a partially hydrated deployed resources object is a good idea or not.
For now I'm just explicitly passing in the data I need from
previous deployement steps (just the lambda_arn).

### DIFF
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -23,6 +23,7 @@ from chalice.constants import CLOUDWATCH_LOGS
 
 
 NULLARY = Callable[[], str]
+OPT_RESOURCES = Optional[DeployedResources]
 
 
 def create_default_deployer(session, prompter=None):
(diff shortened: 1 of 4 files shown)


chalice/deploy/deployer.py
```diff
@@ -296,22 +297,22 @@ def __init__(self, aws_client):
...
                     existing_resources.rest_api_id):
             print "API Gateway rest API already found."
             rest_api_id = existing_resources.rest_api_id
-            return self._create_resources_for_api(config, rest_api_id)
+            return self._create_resources_for_api(config, rest_api_id,
+                                                  lambda_arn)
         else:
             print "Initiating first time deployment..."
-            return self._first_time_deploy(config)
+            return self._first_time_deploy(config, lambda_arn)
```
stealthycoin: It seems like you flip-flop between calling it `handler_arn` and `lambda_arn`. Given that this PR is titled remove references to `lambda_arn`, it seems like you should stick with `handler_arn` to make it clearer.

jamesls: It should always be `lambda_arn` internally in the code.  External references (for example in deployed.json) use `api_handler_arn` because it's referencing the specific lambda function which is the API handler.  Internally in the code, a good chunk of the libraries are generic in the sense that they just require a lambda_arn, they have no context on whether or not it's the api handler arn.

Just to clarify, the PR is to remove `config.lambda_arn` because it's been replaced with all the deployed.json code.

PR #272, opened 2017-04-03
Add support for per-stage IAM policy files

jamesls:
This reworks some of the existing IAM policy logic to properly be "stage aware":

* If you disable autogen policy, then we'll look for a `.chalice/policy-<stage>.json` file.
* You can change this value by using the newly added `iam_policy_file`
* To ease upgrading, if you're deploying to the `dev`, stage we'll also check `.chalice/policy.json` (the pre-existing filename) if there's no `.chalice/policy-dev.json` file.

Note that this builds on https://github.com/awslabs/chalice/pull/271, so you'll see that diff as well until #271 is merged.

Also, I'll be sending a follow up PR that adds docs/upgrade notes for all the package related changes.


chalice/package.py
```diff
@@ -26,9 +27,11 @@ def create_app_packager(config):
             CFNSwaggerGenerator('{region}', '{lambda_arn}'),
             PreconfiguredPolicyGenerator(
                 config,
-                ApplicationPolicyHandler(osutils))),
+                ApplicationPolicyHandler(
+                    osutils, AppPolicyGenerator(osutils)))),
         LambdaDeploymentPackager(),
-        ApplicationPolicyHandler(osutils),
+        # TODO: remove duplication here.
```
kyleknap: Did you miss a todo here? Seems like you can share the instance of the policy generator.

tests/unit/deploy/test_deployer.py
```diff
@@ -149,18 +149,71 @@ def test_policy_autogenerated_when_enabled(app_policy,
...
+                           chalice_stage='dev')
+
+    previous_policy = '{"Statement": ["foo"]}'
+    filename = os.path.join('.', '.chalice', policy_filename)
+    in_memory_osutils.filemap[filename] = previous_policy
+    generated = app_policy.generate_policy_from_app_source(config)
+    assert generated == json.loads(previous_policy)
+
+
+def test_can_provide_stage_specific_policy_for_other_stage(app_policy,
```
kyleknap: Is there a test for when ``autogen_policy`` is ``True``, but the stage is not dev? Looks like you have one for when the stage is dev, but not when it is the non default stage.

chalice/package.py
jamesls: Ended up being removed due to refactoring in the env var PR (https://github.com/awslabs/chalice/pull/273/files#diff-7637287bcfcc038a2e99fb26ac726348R32)

tests/unit/deploy/test_deployer.py
jamesls: Just pushed a commit that adds a test for this case.

chalice/package.py
kyleknap: Sounds good

PR #243, opened 2017-02-21
Environment Variable and VPC Support

emellis:
Adds support for lambda environment variables to chalice.

PR #273, opened 2017-04-03
Add support for environment_variables config option

jamesls:
These changes include:

This builds on the initial work from @emellis (#243) and adds
support for the environment variables config option.

* Updating the config object with an environment_variables property
* Adding the concept of "chain merge" to config.  This allows
a dict to be updated (merged) instead of returning on the first value
found.
* Update API based deployer to inject env vars when needed.  This is
needed for both create_function and update_function calls.
* Changed update_function_code to just update_function which will
optionally update function config if required
* Update packager to add environment variables to SAM template.

NOTE: This doesn't merge cleanly on the package branch, and I based
it off the latest PR (#272).  The only commit you need to review here
is the latest commit: https://github.com/jamesls/chalice/commit/507dd116f36c038dabc1b9287c60d99fe3339955

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -48,8 +48,9 @@ class TypedAWSClient(object):
         except client.exceptions.ResourceNotFoundException:
             return False
 
-    def create_function(self, function_name, role_arn, zip_contents):
-        # type: (str, str, str) -> str
+    def create_function(self, function_name, role_arn, zip_contents,
+                        environment_variables=None):
+        # type: (str, str, str, Optional[Dict[str, str]]) -> str
         kwargs = {
             'FunctionName': function_name,
             'Runtime': 'python2.7',
@@ -58,6 +59,8 @@ class TypedAWSClient(object):
             'Role': role_arn,
             'Timeout': 60,
         }
+        if environment_variables is not None:
+            kwargs['Environment'] = {"Variables": environment_variables}
         client = self._client('lambda')
         attempts = 0
         while True:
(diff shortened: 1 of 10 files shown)


chalice/awsclient.py
```diff
@@ -75,10 +78,17 @@ def create_function(self, function_name, role_arn, zip_contents):
...
-    def update_function_code(self, function_name, zip_contents):
-        # type: (str, str) -> Dict[str, Any]
-        return self._client('lambda').update_function_code(
+    def update_function(self, function_name, zip_contents,
+                        environment_variables=None):
+        # type: (str, str, Optional[Dict[str, str]]) -> Dict[str, Any]
+        lambda_client = self._client('lambda')
+        return_value = lambda_client.update_function_code(
             FunctionName=function_name, ZipFile=zip_contents)
+        if environment_variables is not None:
```
emellis: The way this code sits currently won't delete environment variables if the user removes all references to it from their config file. I think it makes sense to have the deployment reflect the config as much as possible. Maybe use something like the following if this is functionality that you want.
`
if not environment_variables:
    environment_variables = {}
lambda_client.update_function_configuration(
    FunctionName=function_name,
    Environment={"Variables": environment_variables})
`

kyleknap: @emellis that makes sense to me as well. The one downside of doing this is that the ``UpdateFunctionConfiguration`` operation will be called every time even when no environment variables were configured. @jamesls is there any other downsides with this approach?

jamesls: Good point, I was hoping to avoid calling update_function_configuration unless needed, I think it's best to just call it every time.  Otherwise I'd have to store previously deployed function config values which I'd prefer not to do (I plan on adding support for all the various lambda config options, just wanted to start with environment_variables to make sure there were no issues)

It is an extra function call, but I don't think it will slow things down too much.

I'll get this updated.

jamesls:
Updated accordingly, should be ready for another look.  I also rebased against the latest package branch so the diff should be easier to read.

tests/functional/test_awsclient.py
```diff
@@ -25,8 +25,11 @@ def test_deploy_rest_api(stubbed_session):
 
 
 def test_update_function_code_only(stubbed_session):
```
kyleknap: Probably will want to update the test name now.

jamesls: Fixed.

PR #276, opened 2017-04-03
Merge package branch

jamesls:
This merges the `package` branch that's been tracking all the misc features for packaging/stage support.  This code has all been reviewed before so I'm just giving one last look before merging.


cotozachoroba:
Nice

PR #277, opened 2017-04-03
Add generate-pipelines command

jamesls:
This will create a cfn template for a starter CD pipeline
based on your chalice app.  I see this expanding over
time to allow more granular choices about the initial pipeline,
but for now a user will have to modify the generated template
before deploying.

NOTE: to test I had to change the `sudo pip install chalice` line to point to a copy that contained the package command.  This won't be an issue once the package command is released on pypi.

### DIFF
--- a/chalice/constants.py
+++ b/chalice/constants.py
@@ -70,3 +70,142 @@ CLOUDWATCH_LOGS = {
     ],
     "Resource": "arn:aws:logs:*:*:*"
 }
+
+
+CODEBUILD_POLICY = {
+    "Version": "2012-10-17",
+    # This is the policy straight from the console.
+    "Statement": [
+        {
+            "Action": [
+                "logs:CreateLogGroup",
+                "logs:CreateLogStream",
+                "logs:PutLogEvents"
+            ],
+            "Resource": "*",
+            "Effect": "Allow"
+        },
+        {
+            "Action": [
+                "s3:GetObject",
+                "s3:GetObjectVersion",
+                "s3:PutObject"
+            ],
+            "Resource": "arn:aws:s3:::*",
+            "Effect": "Allow"
+        }
+    ]
+}
+
+CODEPIPELINE_POLICY = {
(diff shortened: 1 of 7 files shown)


chalice/constants.py
```diff
@@ -70,3 +70,142 @@ def index():
...
+                "s3:GetObjectVersion",
+                "s3:PutObject"
+            ],
+            "Resource": "arn:aws:s3:::*",
+            "Effect": "Allow"
+        }
+    ]
+}
+
+CODEPIPELINE_POLICY = {
```
kyleknap: So is this the minimal amount of permissions needed to get codepipeline working? Seems a little weird that there is a good amount of services included in it that don't really need to be like beanstalk and opsworks

jamesls: That's from the console setup, yeah.

jamesls:
The process isn't really defined at this point.  It will really depend on the types of changes people want made.  If anything I would expect more parameterization to happen, i.e github instead of codecommit, more stages, etc.  Right now just have to modify the template manually, but eventually we'd expose CLI options for these.  Or at least that's what I had in mind.

chalice/constants.py
kyleknap: Yeah. I am fine keeping it as it. It just pointed it out as it looked weird given you really only used half of those services and normally for IAM policies you keep it as minimal as possible.

kyleknap:
@jamesls I guess I am wondering are you following up with documentation on how you can use the ``generate-pipeline`` command? It would be nice to at least have some sort of guidance on how to use these since it is not really defined as of now. It definitely took me some time to figure out how the workflow it would work.

jamesls:
@kyleknap How about this?  I can update the PR to add docs for the `generate-pipeline` command so you can run `chalice generate-pipeline --help` to get more info on what this command does.

kyleknap:
@jamesls sounds good to me.

jamesls:
Cool added docs.

Issue #279, opened 2017-04-05
deploy failed with chalice 0.7


jcbertin:
I have the following here when invoking :

```
$ chalice --debug deploy --no-autogen-policy
...
2017-04-05 19:59:32,901 - botocore.hooks - DEBUG - Event choose-signer.lambda.CreateFunction: calling handler <function set_operation_specific_signer at 0x10afd9e60>
2017-04-05 19:59:32,962 - botocore.auth - DEBUG - Calculating signature using v4 auth.
2017-04-05 19:59:32,962 - botocore.auth - DEBUG - CanonicalRequest:
POST
/2015-03-31/functions

host:lambda.us-east-1.amazonaws.com
x-amz-date:20170405T175932Z

host;x-amz-date
xxxx
2017-04-05 19:59:32,963 - botocore.auth - DEBUG - StringToSign:
AWS4-HMAC-SHA256
20170405T175932Z
20170405/us-east-1/lambda/aws4_request
xxxx
2017-04-05 19:59:32,963 - botocore.auth - DEBUG - Signature:
xxxx
2017-04-05 19:59:32,963 - botocore.endpoint - DEBUG - Sending http request: <PreparedRequest [POST]>
2017-04-05 19:59:32,964 - botocore.vendored.requests.packages.urllib3.connectionpool - INFO - Starting new HTTPS connection (1): lambda.us-east-1.amazonaws.com
2017-04-05 19:59:39,248 - botocore.vendored.requests.packages.urllib3.connectionpool - DEBUG - "POST /2015-03-31/functions HTTP/1.1" 201 523
2017-04-05 19:59:39,248 - botocore.parsers - DEBUG - Response headers: {'date': 'Wed, 05 Apr 2017 17:59:38 GMT', 'x-amzn-requestid': 'a00605ef-1a29-11e7-ba6e-0bd33930e47f', 'content-length': '523', 'content-type': 'application/json', 'connection': 'keep-alive'}
2017-04-05 19:59:39,248 - botocore.parsers - DEBUG - Response body:
[...]

PR #282, opened 2017-04-05
Fixed api_key_required=True config.

jcbertin:
Fixes issue #279


jamesls:
Merged in d47a7dc, thanks for the fix.

PR #283, opened 2017-04-06
Fixed profile config.

jcbertin:


jcbertin:
The problem is that CLIFactory always use self.profile for botocore session which is set in cli/__init__.py :129 to argument given when invoking chalice command or None. Thus its always ignore profile setting given in config.json…

jcbertin:
ping

jcbertin:
Rebased PR.
Added specific tests.

PR #286, opened 2017-04-08
Add binary_media_types and support the binary request and response.

haginaga:
This will support binary data.
Ref. https://aws.amazon.com/jp/about-aws/whats-new/2016/11/binary-data-now-supported-by-api-gateway/

Ex1. upload binary data, and then store binary data in S3 buckets.
```
# add binary_media_types to the API gateway
app.add_binary_media_types(['application/octet-stream'])

@app.route('/upload', methods=['POST'], content_types=['application/octet-stream'])
def upload:
    # Body is binary data.
    body = app.current_request.raw_body

    # Use boto3 s3 client
    s3.put_object(
        Bucket=S3_BUCKET,
        Key=KEY,
        Body=body,
        ContentType='application/octet-stream'
    )

    return {}
```

Ex2. send a image file, and then response a resized image file.
app: 
```
# add binary_media_types to the API gateway
app.add_binary_media_types(['image/png', 'image/jpeg'])

@app.route('/resize', methods=['POST'], content_types=['image/png'])
def resize:
    # Body is binary data.
    body = app.current_request.raw_body

    # Resize a image.
    resized_image = _resize(body)

    # create a response with make the 'is_binary' attribute True
    return Response(body=resized_image,
                    status_code=200,
                    headers={'Content-Type': 'image/png'},
                    is_binary=True)
```
If make the 'is_binary' attribute True, the response is binary data.
If make the 'is_binary' attribute False, the response is base64 encoded data.
If the 'is_binary' attribute is omitted, it is False.

client:
```
curl -s -X POST -H 'Content-Type: image/png' -H 'Accept: image/png' --data-binary @foo.png ¥
https://xxxxxx.execute-api.region.amazonaws.com/dev/resize > /tmp/resized_foo.png
```
The client needs to add the 'Accept' header to receive binary data.
If the 'Accept' header is omitted, this response body is base64 encoded.


ogrodnek:
Thanks, @haginaga ! Looking forward to using this!

stealthycoin:
Closing due to #352 being merged.

PR #297, opened 2017-04-19
 Return empty dict when headers are None

dharrisio:
When testing a method from the API Gateway Console a 502 status is
returned because of an attempt to iterate over a `NoneType` object.

To reproduce (v0.8.0):
- `chalice new-project chalice-hello-world`
- `cd chalice-hello-world && chalice deploy`
- Navigate to GET method in console and test method

This should result in a 502 response, "message": "Internal server error".


jamesls:
Thanks for the pull request.

PR #303, opened 2017-04-24
Lambda permissions

emellis:
This solves a couple of issues I discovered when trying to integrate a chalice app into my CI/CD pipeline. The first was that underscores are not allowed in CloudFormation logical names. To solve the first issue I am simply removing any underscores found in the view names when we generate the SAM template. Second, I noticed that on some API gateway endpoints I was getting invalid lambda permissions. The addition of an explicit lambda permission for the create API gateway seems to have completely solved this on my end. I put it up here but if we want to handle it in some other way I am open to other suggestions.


kapilt:
This has already been resolved afaics, and this pr can be closed, it would be nice to reference the pr that actually did resolve this issue.

PR #311, opened 2017-04-28
Add support for a more granular CORSConfig object.

stealthycoin:
Allows specification of specific access control headers. Docs still
need to be written.

Implements: https://github.com/awslabs/chalice/issues/70#issuecomment-248787037

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -100,6 +100,54 @@ class CaseInsensitiveMapping(Mapping):
         return 'CaseInsensitiveMapping(%s)' % repr(self._dict)
 
 
+class CORSConfig(object):
+    """A cors configuration to attach to a route."""
+
+    _REQUIRED_HEADERS = ['Content-Type', 'X-Amz-Date', 'Authorization',
+                         'X-Api-Key', 'X-Amz-Security-Token']
+
+    def __init__(self, allow_origin='*', allow_headers=None,
+                 expose_headers=None, max_age=None, allow_credentials=None):
+        self.allow_origin = allow_origin
+
+        if allow_headers is None:
+            allow_headers = list(self._REQUIRED_HEADERS)
+        else:
+            allow_headers.extend(self._REQUIRED_HEADERS)
+        self._allow_headers = allow_headers
+
+        if expose_headers is None:
+            expose_headers = []
+        self._expose_headers = expose_headers
+
+        self._max_age = max_age
+        self._allow_credentials = allow_credentials
+
+    @property
+    def allow_headers(self):
+        return ','.join(self._allow_headers)
+
+    def get_access_control_headers(self):
+        headers = {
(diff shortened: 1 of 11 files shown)


chalice/app.py
```diff
@@ -100,6 +100,54 @@ def __repr__(self):
...
+                         'X-Api-Key', 'X-Amz-Security-Token']
+
+    def __init__(self, allow_origin='*', allow_headers=None,
+                 expose_headers=None, max_age=None, allow_credentials=None):
+        self.allow_origin = allow_origin
+
+        if allow_headers is None:
+            allow_headers = list(self._REQUIRED_HEADERS)
+        else:
+            allow_headers.extend(self._REQUIRED_HEADERS)
```
jamesls: What if a passed in `allow_header` contains one of the `_REQUIRED_HEADERS`?  Seems like we should check for existing values or use a `set()`.

chalice/app.py
```diff
@@ -167,6 +215,10 @@ def __init__(self, view_function, view_name, path, methods,
         #: e.g, '/foo/{bar}/{baz}/qux -> ['bar', 'baz']
         self.view_args = self._parse_view_args()
         self.content_types = content_types
+        if cors is True:
```
jamesls: Can you add a comment about why the identity check is being used instead of just boolean comparison.  This is probably gonna catch other dev's attention because it's not a common pattern so it's worth explaining why we need to do this.

chalice/app.py
```diff
@@ -354,7 +406,7 @@ def _validate_response(self, response):
                                        (header, value))
 
     def _cors_enabled_for_route(self, route_entry):
-        return route_entry.cors
+        return route_entry.cors is not None
 
-    def _add_cors_headers(self, response):
-        response.headers['Access-Control-Allow-Origin'] = '*'
+    def _add_cors_headers(self, response, cors):
+        response.headers.update(cors.get_access_control_headers())
```
jamesls: I realize the old code didn't do this, but I wonder if we should be overriding headers if they provide these values themselves, i.e a custom `Response` value.

chalice/app.py
stealthycoin: Yea I was considering pulling it out into a _configure_cors and put a comment in there. I'll just do the comment.

chalice/app.py
stealthycoin: Hmmm, that seems like the correct behavior to me. I wouldn't want my explicitly set headers getting clobbered by my framework.

chalice/app.py
stealthycoin: Good point, I'll fix that.

jamesls:
Looks good, just a few things remaining:

* Double check on the integration tests and if we need to update them for the more granular preflight request
* Documentation for this feature.

PR #319, opened 2017-04-29
Update example to work with Python 2 and 3

alasdairnicol:


alasdairnicol:
This example doesn't work with Chalice 0.8.0, so it probably shouldn't be merged until there has been a release which includes `chalice.compat`.

jamesls:
Thanks for the PR.  For now, I'll go ahead and update the example to just to a version check for python2/3.  I'll need to figure out how to best deal with `chalice.compat` but for now I'd prefer to get the example working in python2/3.

PR #322, opened 2017-05-01
Add back custom auth

jamesls:
This needs to go through the `define_authorizer` in order to work
with the swagger generation.

### DIFF
--- a/chalice/deploy/swagger.py
+++ b/chalice/deploy/swagger.py
@@ -23,6 +23,8 @@ class SwaggerGenerator(object):
         }
     }  # type: Dict[str, Any]
 
+    _KNOWN_AUTH_TYPES = ['cognito_user_pools', 'custom']
+
     def __init__(self, region, lambda_arn):
         # type: (str, str) -> None
         self._region = region
@@ -52,7 +54,6 @@ class SwaggerGenerator(object):
     def _add_to_security_definition(self, security, api_config, authorizers):
         # type: (Any, Dict[str, Any], Dict[str, Any]) -> None
         for auth in security:
-            # TODO: Add validation checks for unknown auth references.
             name = list(auth.keys())[0]
             if name == 'api_key':
                 # This is just the api_key_required=True config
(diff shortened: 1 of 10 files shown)


chalice/deploy/swagger.py
```diff
@@ -62,18 +63,44 @@ def _add_to_security_definition(self, security, api_config, authorizers):
...
                     'name': authorizer_config['header'],
                     'x-amazon-apigateway-authtype': auth_type,
                     'x-amazon-apigateway-authorizer': {
+                    }
+                }
+                if auth_type == 'custom':
+                    auth_config = {
+                        'type': auth_type,
+                        'authorizerUri': authorizer_config['authorizer_uri'],
+                        'authorizerResultTtlInSeconds': 300,
```
stealthycoin: Any reason not to expose this in the config?

jamesls:
Ok, after chatting about this, I'd like to propose some changes here.

First, we should have a consistent pattern for configuring advanced resources.  I like the pattern used with the `CORSConfig` PR: https://github.com/awslabs/chalice/pull/311, which is you pass in a specific object that represents the configuration you'd like for that route.  The `define_authorizer` relies on string name references.  It's nice that it decouples the order of the auth needed, but I think a better option here is to just pass in an `Authorizer` object, with a different subclass for each type you need.  With the current PR, I have the union of args needed for both the cognito user pools auth and custom auth, and would need to add additional validation to get decent error messages (what if you specify both?  or you specify neither?).  So the new version would look like this:

```python
from chalice.app import CustomAuthorizer
from chalice.app import CognitoUserPoolsAuthorizer
from chalice import Chalice

app = Chalice(app_name='testauth')
custom = CustomAuthorizer('MyAuth', header='Authorization', provider_uri='...')
cognito_auth = CognitoUserPoolsAuthorizer(
    'MyUserPool', header='Authorization', provider_arns=['myarn...'],
)

@app.route('/auth1', authorizer=custom)
def custom_auth(): ...

@app.route('/cognito-auth', authorizer=cognito_auth)
def cognito_auth(): ...
```

To help transition, we should keep `define_authorizer(...)`, which only works with cognito user pools, for a period of time.  We could add a deprecation warning and eventually remove it.

Let me know if anyone has concerns about this.

jamesls:
@stealthycoin Feedback incorporated, ready for another look.

PR #323, opened 2017-05-03
First crack at `chalice delete`

stealthycoin:
PR for the codecov.

Closes https://github.com/awslabs/chalice/issues/40

cc @jamesls


chalice/utils.py
```diff
@@ -7,6 +7,27 @@
...
 
 
+def remove_stage_from_deployed_values(key, filename):
+    # type: (str, str) -> None
+    """Delete a top level key from the deployed JSON file."""
+    final_values = {}  # type: Dict[str, Any]
+    try:
+        with open(filename, 'r') as f:
+            final_values = json.load(f)
+    except IOError:
```
jamesls: I'd prefer we just stick with a single way of dealing with files, either as os.path.isfile() used in `record_deployed_values` or the try/except here.  I'm also pretty sure you can get OSErrors as well.

chalice/deploy/deployer.py
```diff
@@ -141,15 +141,29 @@ def confirm(self, text, default=False, abort=False):
         # type: (str, bool, bool) -> bool
         return default
 
+    def Abort(self):
```
stealthycoin: Should I wrap `print`/`echo` up in there as well? Or is that outside the scope of a `prompter`?

chalice/deploy/deployer.py
```diff
@@ -141,15 +141,29 @@ def confirm(self, text, default=False, abort=False):
...
+        # type: (APIGatewayDeployer, LambdaDeployer, NoPrompt) -> None
         self._apigateway_deploy = apigateway_deploy
         self._lambda_deploy = lambda_deploy
+        self._prompter = prompter
+
+    def delete(self, config, chalice_stage_name=DEFAULT_STAGE_NAME):
+        # type: (Config, str) -> None
+        existing_resources = config.deployed_resources(chalice_stage_name)
+        if existing_resources is None:
+            print('No existing resources found.')
```
stealthycoin: Good point, I'll update that.

chalice/utils.py
stealthycoin: Yea, this was me struggling to get codecov to accept that I was writing to a file. It didn't work, I'll put it back to `os.path.isfile()`.

chalice/deploy/deployer.py
jamesls: Well, after thinking about this, I'm not even sure if abort() make sense here.  This interface is just about prompting.  Aborting doesn't really make sense here.  Does this need to be an explicit part of this interface?  Can we not just raise an exception?

stealthycoin: So the existing prompter is basically used for control flow around which resources are deployed/deleted. I think rather than raising an exception or aborting outright I could also just add an `else` clause and move the actual resource deletion into that. And then take the abort method off of the Prompter interface.

PR #327, opened 2017-05-04
Code cleanup: fixup linting issues

jamesls:
You now get a clean run through pylint.
I've also gone through the pylint ignores
and updated them based on what's appropriate
for this project.

I've also changed `make check` to have pylint
check everything.  It was previously only checking
for errors.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -36,7 +36,7 @@ extension-pkg-whitelist=
 # operator. Joining a lot of strings can lead to a maximum recursion error in
 # Pylint and this flag can prevent that. It has one side effect, the resulting
 # AST will be different than the one from reality.
-optimize-ast=no
+optimize-ast=yes
 
 
 [MESSAGES CONTROL]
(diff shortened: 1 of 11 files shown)


jamesls:
@stealthycoin Should be good to look at again.  I wasn't using the same commit off the master branch for pylint, and the newer version has a few errors I wasn't accounting for.

.pylintrc
```diff
@@ -36,7 +36,7 @@ extension-pkg-whitelist=
 # operator. Joining a lot of strings can lead to a maximum recursion error in
 # Pylint and this flag can prevent that. It has one side effect, the resulting
 # AST will be different than the one from reality.
-optimize-ast=no
```
stealthycoin: Is this comment relevant anymore since you changed the `optimize-ast` parameter?

jamesls: That comment comes from the default generated pylint file.

PR #334, opened 2017-05-11
Added IamAuthorizer class to support AWS_IAM authorizor on API Gateway

zepplen:


jamesls:
Awesome!  Thanks for the pull request, I'm trying this out now.

PR #340, opened 2017-05-15
Tag deployed lambda functions

jamesls:
Makes it possible for customers to easily see
which lambda functions are used for chalice.
Also opens up the possibility of us not having
to rely on the local deployed.json.  That would
be future work though.  This commit
only adds support for tagging the functions.

The format uses colon separated values because
Lambda does not allow CSV values for the value
of a tag.

I also bumped the min version of botocore to the
version where tagging support was added to lambda.

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -66,7 +66,7 @@ class TypedAWSClient(object):
     def create_function(self, function_name, role_arn, zip_contents,
                         environment_variables=None, runtime='python2.7',
                         tags=None):
-        # type: (str, str, str, _STR_MAP, str, Optional[Dict[str, str]]) -> str
+        # type: (str, str, str, _STR_MAP, str, _STR_MAP) -> str
         kwargs = {
             'FunctionName': function_name,
             'Runtime': runtime,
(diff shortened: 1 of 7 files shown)


chalice/awsclient.py
```diff
@@ -64,8 +64,9 @@ def get_function_configuration(self, name):
         return response
 
     def create_function(self, function_name, role_arn, zip_contents,
-                        environment_variables=None, runtime='python2.7'):
-        # type: (str, str, str, _ENV_VAR, str) -> str
+                        environment_variables=None, runtime='python2.7',
+                        tags=None):
+        # type: (str, str, str, _STR_MAP, str, Optional[Dict[str, str]]) -> str
```
kyleknap: Why was ``_STR_MAP`` not used for tags type annotation? ``_STR_MAP`` is the same as ``Optional[Dict[str,str]]`` right?

kyleknap:
Oh also do we need to update the SAM template generator for this as well?

jamesls:
@kyleknap Looks like I didn't push the SAM changes.  I have it coded locally, I'll get this updated shortly.

jamesls:
@stealthycoin Yeah that will be part of the PR that adds support for user tagging.

jamesls:
@kyleknap Feedback incorporated should be good to look at again.

PR #345, opened 2017-05-22
Ensure make check is ran in travis

kyleknap:
Looking at the travis configuration, it does not actually run `make check`. It only runs `make typecheck` which is completely different. I ran into this when I ran `make prcheck` when developing on python 3. If it fails (as I suspect it will), I will send updates in this PR to make Travis run cleanly again.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -329,7 +329,7 @@ max-branches=12
 max-statements=30
 
 # Maximum number of parents for a class (see R0901).
-max-parents=5
+max-parents=6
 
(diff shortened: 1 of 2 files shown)


kyleknap:
Cool. Test failed. Will update the code to get everything passing.

kyleknap:
@jamesls @stealthycoin Fixed the linters. It ended up being just configuration updates. I tried to be pretty descriptive in the commits and comments. No actual code changes in `chalice` were needed to get these to pass.

PR #350, opened 2017-05-23
Fix importing vendor lib during config creation.

### LINKED ISSUE #349: Custom Package in vendor folder not recognized
I have a custom package called datareferencecommon that exists in our company's private artifactory.  As instructed in chalice/docs/source/topics/packaging.rst I added a vendor folder at the same level as my app.py file.  I created a datareferencecommon folder and loaded all of the *.py files into this folder.  Folder looks like this:
```
root
    |--- chalicelib
    |--- app.py
    |--- requirements.txt
    |--- vendor
          |--- datareferencecommon
                 |--- __init__.py
                 |--- config.py
                 |--- discovery.py
                 |--- utils.py
```

In app.py, there is a statement: `from datareferencecommon import discovery, utils`
I ran chalice deploy from the root folder and receive an error "no module named datareferencecommon".  What am I doing wrong?  Thanks.

Also, am I understanding this correctly?  chalice deploy won't install from a private artifactory even if the ~/.pypirc file exists and points appropriately to the private artifactory?

stealthycoin:
Adds the vendor directory to the path when loading the app locally. This
is required for both deployment and local testing since both commands
load the app.py file, which will fail if anything is imported from the
vendor directory without it being added to the pythonpath. fixes #349 

I also manually tested `chalice deploy` and `chalice local`

cc @jamesls @kyleknap

### DIFF
--- a/chalice/cli/factory.py
+++ b/chalice/cli/factory.py
@@ -131,6 +131,15 @@ class CLIFactory(object):
         # type: () -> Chalice
         if self.project_dir not in sys.path:
             sys.path.insert(0, self.project_dir)
+        # The vendor directory has its contents copied up to the top level of
+        # the deployment package. This means that imports will work in the
+        # lambda function as if the vendor directory is on the python path.
+        # For loading the config locally we must add the vendor directory to
+        # the path so it will be treated the same as if it were running on
+        # lambda.
+        vendor_dir = os.path.join(self.project_dir, 'vendor')
+        if os.path.isdir(vendor_dir):
+            sys.path.insert(0, vendor_dir)
         try:
             app = importlib.import_module('app')
             chalice_app = getattr(app, 'app')
--- a/chalice/cli/factory.py
+++ b/chalice/cli/factory.py
@@ -138,7 +138,7 @@ class CLIFactory(object):
         # the path so it will be treated the same as if it were running on
         # lambda.
         vendor_dir = os.path.join(self.project_dir, 'vendor')
-        if os.path.isdir(vendor_dir):
+        if os.path.isdir(vendor_dir) and vendor_dir not in sys.path:
             sys.path.insert(0, vendor_dir)
         try:
             app = importlib.import_module('app')
--- a/tests/functional/cli/test_factory.py
+++ b/tests/functional/cli/test_factory.py
@@ -104,15 +104,11 @@ def test_can_import_vendor_package(clifactory):
     vendedlib_dir = os.path.join(clifactory.project_dir, 'vendor', 'vendedlib')
     os.makedirs(vendedlib_dir)
     open(os.path.join(vendedlib_dir, '__init__.py'), 'a').close()
-    with open(os.path.join(vendedlib_dir, 'utils.py'), 'w') as f:
-        f.write(
-            'def data_getter():\n'
-            '    return "some data value"\n'
-        )
+    open(os.path.join(vendedlib_dir, 'submodule.py'), 'a').close()
     app_py = os.path.join(clifactory.project_dir, 'app.py')
     with open(app_py, 'r+') as f:
         data = f.read()
         f.seek(0, 0)
-        f.write('from vendedlib.utils import data_getter\n%s' % data)
+        f.write('from vendedlib import submodule\n%s' % data)
     config = clifactory.create_config_obj()
     assert isinstance(config, Config)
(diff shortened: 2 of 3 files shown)


chalice/cli/factory.py
```diff
@@ -131,6 +131,15 @@ def load_chalice_app(self):
         # type: () -> Chalice
         if self.project_dir not in sys.path:
             sys.path.insert(0, self.project_dir)
+        # The vendor directory has its contents copied up to the top level of
+        # the deployment package. This means that imports will work in the
+        # lambda function as if the vendor directory is on the python path.
+        # For loading the config locally we must add the vendor directory to
+        # the path so it will be treated the same as if it were running on
+        # lambda.
+        vendor_dir = os.path.join(self.project_dir, 'vendor')
```
jamesls: You should use the same logic as above and ensure we don't add this multiple times.

tests/functional/cli/test_factory.py
```diff
@@ -97,3 +97,22 @@ def test_filename_and_lineno_included_in_syntax_error(clifactory):
...
+        f.write(
+            'def data_getter():\n'
+            '    return "some data value"\n'
+        )
+    app_py = os.path.join(clifactory.project_dir, 'app.py')
+    with open(app_py, 'r+') as f:
+        data = f.read()
+        f.seek(0, 0)
+        f.write('from vendedlib.utils import data_getter\n%s' % data)
+    config = clifactory.create_config_obj()
```
jamesls: Is this right?  I would expect us to try to do something with `data_getter` and it's return value.

chalice/cli/factory.py
stealthycoin: Updated.

tests/functional/cli/test_factory.py
```diff
@@ -97,3 +97,18 @@ def test_filename_and_lineno_included_in_syntax_error(clifactory):
...
+    vendedlib_dir = os.path.join(clifactory.project_dir, 'vendor', 'vendedlib')
+    os.makedirs(vendedlib_dir)
+    open(os.path.join(vendedlib_dir, '__init__.py'), 'a').close()
+    open(os.path.join(vendedlib_dir, 'submodule.py'), 'a').close()
+    app_py = os.path.join(clifactory.project_dir, 'app.py')
+    with open(app_py, 'r+') as f:
+        data = f.read()
+        f.seek(0, 0)
+        f.write('from vendedlib import submodule\n%s' % data)
+    config = clifactory.create_config_obj()
```
jamesls: So after looking at this again, I'd still prefer the test to be explicit.  This is implicitly relying on the fact that if we can't import the module then this code will propagate some exception in `create_config_obj`.  I'd rather be more explicit, ensure that not only do we not raise an exception but we are in fact able to access the thing we're importing.  As an example:

```
# submodule.py:

SOME_VALUE = 'from-submodule'

# app.py

from vendedlib import submodule
app = Chalice(app_name='foo')
app.from_submodule = submodule.SOME_VALUE

# test
assert app.from_submodule == 'from-submodule'
```

Or something along those lines.  It's just more explicit about successfully importing the vendored submodule

tests/functional/cli/test_factory.py
```diff
@@ -97,3 +97,21 @@ def test_filename_and_lineno_included_in_syntax_error(clifactory):
...
+
+def test_can_import_vendor_package(clifactory):
+    # Tests that vendor packages can be imported during config loading.
+    vendedlib_dir = os.path.join(clifactory.project_dir, 'vendor', 'vendedlib')
+    os.makedirs(vendedlib_dir)
+    open(os.path.join(vendedlib_dir, '__init__.py'), 'a').close()
+    with open(os.path.join(vendedlib_dir, 'submodule.py'), 'a') as f:
+        f.write('CONST = "foo bar"\n')
+    app_py = os.path.join(clifactory.project_dir, 'app.py')
+    with open(app_py, 'r+') as f:
```
jamesls: This doesn't need to block merging, but wouldn't just be simpler to open the file at the end and append `from vendedlib import submodule\napp.imported_value = submodule.CONST`?  Not really sure if the `.seek(0, 0)` and `.seek(0, 1)` make the test any clearer.

jamesls:
Oh and don't forget to add an entry to the changelog, thanks.

PR #351, opened 2017-05-23
Add additional lambda configuration support

kyleknap:
Added the following configuration options:

* `lambda_timeout`: Sets the maximum time duration of a lambda function
* `lambda_memory_size`: Sets the amount of memory to use in the lambda function
* `tags`: Sets the tags for the lambda function

Implements https://github.com/awslabs/chalice/issues/347

I also noticed that SAM/Cloudformation deployments were not using the default timeout of 60 for chalice applications. So I updated that as well.


chalice/utils.py
```diff
@@ -4,7 +4,24 @@
 
 from typing import IO, Dict, Any  # noqa
 
+from chalice import __version__ as chalice_version
 from chalice.constants import WELCOME_PROMPT
+from chalice.config import Config  # noqa
+
+
+def get_application_tags(config):
```
jamesls: Capturing offline discussion: I'm fine with moving this over to the `Config` object.

chalice/awsclient.py
```diff
@@ -104,10 +109,16 @@ def delete_function(self, function_name):
         except lambda_client.exceptions.ResourceNotFoundException:
             raise ResourceDoesNotExistError(function_name)
 
-    def update_function(self, function_name, zip_contents,
-                        environment_variables=None,
-                        runtime=None, tags=None):
-        # type: (str, str, _STR_MAP, _OPT_STR, _STR_MAP) -> Dict[str, Any]
+    def update_function(self,
+                        function_name,  # type: str
```
jamesls: Please align the type definitions vertically.  There's an example of this in [pep 484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code).

chalice/awsclient.py
```diff
@@ -121,15 +132,55 @@ def update_function(self, function_name, zip_contents,
...
             kwargs['Runtime'] = runtime
+        if timeout is not None:
+            kwargs['Timeout'] = timeout
+        if memory_size is not None:
+            kwargs['MemorySize'] = memory_size
         lambda_client.update_function_configuration(**kwargs)
         if tags is not None:
-            lambda_client.tag_resource(Resource=return_value['FunctionArn'],
-                                       Tags=tags)
+            function_arn = return_value['FunctionArn']
```
jamesls: The tag manipulation (figuring what exactly to tag) feels too level for this function.  Please move this out into a separate method, something like:

```python
if tags is not None:
    self._update_function_tags(return_value['FunctionArn'], tags)
```

chalice/awsclient.py
```diff
@@ -121,15 +132,55 @@ def update_function(self, function_name, zip_contents,
...
+            remote_tags = lambda_client.list_tags(
+                Resource=function_arn)['Tags']
+
+            # Remove any tags that exist remotely but are not included
```
jamesls: Seems like this is just basic set/dict manipulation.  Can't this whole thing be simplified to something like:

```python
tags_to_remove = set(tags_requested) - set(remote_tags)
tags_to_add = {k: v for k, v in tags_requested.items()
              if k not in remote_tags or v != remote_tags[k]}
self._set_function_tags(tags_to_remove, tags_to_add)
```

I find this easier to read and think it's a simpler as well.

chalice/deploy/deployer.py
```diff
@@ -324,18 +325,19 @@ def _first_time_lambda_create(self, config, function_name, stage_name):
...
-            config.lambda_python_version,
-            self._function_tags(config, stage_name),
-        )
-
-    def _function_tags(self, config, stage_name):
-        # type: (Config, str) -> Dict[str, str]
-        tag = 'version=%s:stage=%s:app=%s' % (chalice_version,
-                                              stage_name, config.app_name)
-        return {'aws-chalice': tag}
+        client_kwargs = {
```
jamesls: I'd prefer not to use this style, having the chain of `if <param> is not None...` makes the code harder to read and I don't know how well the type checker check this style.  Why can't we just pass through `config.lambda_timeout` and `config.lambda_memory_size`?  They'll still be `None` if they're not set right?

kyleknap: That won't work because I have the `create_function` taking in `int` as its type with defaults of 60 and 128 respectively. I chose that in the `create_function` (instead of having a default of `None`) because it was consistent with the `runtime` argument having a default string and not `None`. And `None` cannot be passed into botocore client call because it will not pass validation.

I was going over this prior to your comments and I did not like it as well. One thing that I though of to clean this up was to set defaults in the config object so that if I call `config.lambda_timeout` and `config.lambda_memory_size` I will always at least get at least the default back, which would be an integer. It really helps clean up the logic by hoisting this. What are your thoughts on this?

chalice/awsclient.py
kyleknap: Nice. Forgot Python 2.6 support in chalice is not a thing.

chalice/awsclient.py
kyleknap: Sure. That seems fine with me.

chalice/awsclient.py
kyleknap: Sure. Is there possibly a linter for this? I have not used type hints too much so I was not aware of that portion of the PEP.

chalice/deploy/deployer.py
jamesls: I'd prefer not to do that (setting defaults) at the config layer.  We've ran into this several times in the past (botocore config) where we can't differentiate between whether a value was explicitly set *or* the value is a default value. We need to have at least one layer somewhere that can differentiate between those two cases.

I would prefer moving the `if not None` checks into the `create_function` call.  We need that kwargs style anyways for botocore client calls, and it moves the if checks down a layer.

If you want to modify the `runtime` value to be consistent I think that's fine as well (there's no backwards compatibility issues with this API).

kyleknap: That's fine to. I can update it to that.

FWIW, I do not necessarily think it would be difficult to determine if memory size and timeout defaulted if we needed the logic. The way I would have written it would be to add a default argument to the chain lookup and we could have added logic to detect if it defaulted by making the constant an object that you could use an identity check on. But it is fine if we do not want to set the precedent.

kyleknap:
@jamesls I updated based on feedback. It should be good to be looked at again.

jamesls:
@kyleknap  Thanks.  Is the codecov report still accurate?  Looks like it's missing tests when the timeout/memory_size args are omitted?

kyleknap:
@jamesls Yeah looks like there were some missing tests. Did not realize codecov re-edits its original comment. The diff has 100% coverage now with the tests I added.

kyleknap:
I added a changelog entry for each feature introduced in this PR. Merging PR.

Issue #348, opened 2017-05-23
[proposal] Binary Payload Support


stealthycoin:
## Overview
This feature will add support for round-tripping binary data to a chalice deployed app. This includes correctly configuring API Gateway to handle the binary content types and correctly interacting with the API Gateway interface from Lambda. Ideally this would mostly be a transparent change for users, only requiring minor changes in some cases.

Some `Content-Types` will be treated as binary data automatically and will require no change in code to be properly handled by API Gateway.
Default content types that are considered binary:
```
application/epub+zip
application/octet-stream
application/ogg
application/pdf
application/vnd.amazon.ebook
application/x-7z-compressed
application/x-bzip
application/x-bzip2
application/x-rar-compressed
application/x-tar
application/zip
audio/aac
audio/midi
audio/ogg
audio/webm
audio/x-wav
font/ttf
font/woff
font/woff2
image/bmp
image/gif
image/jpeg
image/png
image/svg+xml
image/tiff
image/webp
image/x-icon
video/3gpp
video/3gpp2
video/mpeg
video/ogg
video/webm
```

## Motivation
Open issues related to binary payloads:
#223 
#250  

## Interface
The interface will be very minimal since this change can mostly be taken care of by the framework automatically. In the case where someone wants to send a binary payload and the content type they are using (or forced to use if the request is coming from a third party they do not have control over) is not in this default list, they can customize the list by calling an app level function (`app.api.define_binary_types`) to add more types. 
Example of adding a custom binary type `application/custom-type`:
```python
from chalice import Chalice
app = Chalice('custom-binary')
app.api.define_binary_types(['application/x-special-binary-type'])
```

## Implementation
API Gateway takes requests and matches their `Content-Type` header against a user provided list per restapi resource. Chalice will provide this list through the swagger file property `x-amazon-apigateway-binary-media-types`. When API Gateway transforms this request to lambda event, it will base64 encode the body, and add a top level key called `isBase64Encoded` with a value of `True`. Chalice will check for the presence of the `isBase64Encoded=True` in the event, and base64decode the body before forwarding it on to the corresponding route handler.

For responses, Chalice will check the headers dictionary for a `Content-Type` key with a value matching a known binary type, using the same list that was provided to API Gateway. If the type matches it will base64 encode the body, and attach the `isBase64Encoded=True` key/value to the dictionary it sends to API Gateway. 

## Example
[...]

PR #352, opened 2017-05-24
Added support for binary payloads.

stealthycoin:
Payloads with a known binary Content-Type will be automatically
propagated to lambda with app.current_request.raw_body as a bytes-like
object. Responses which have a bytes body and a known binary Content-Type
header will be base64 encoded before being sent to API gateway, and API
gateway will correctly send the data depending on what Accept headers
were present on the request.

Implements #348 

## Example usage
```python
app = Chalice(app_name='test-binary')

@app.route('/binary', methods=['POST'],
           content_types=['application/octet-stream'])
def binary_round_trip():
    return Response(
        app.current_request.raw_body,
        headers={
            'Content-Type': 'application/octet-stream'
        },
        status_code=200)
```

```
$ chalice deploy
.
.
.
$ test-binary echo -n -e '\xDE\xAD\xBE\xEF' | curl -s -H 'Accept: application/octet-stream' -H 'Content-Type: application/octet-stream' -X POST $(chalice url)binary --data-binary @- | xxd
0000000: dead beef                                ....
$ curl -s -H 'Accept: application/octet-stream' -H 'Content-Type: application/octet-stream' -X POST $(chalice url)binary --data-binary @image.png > round_tripped.png
$ diff image.png round_tripped.png
$
```

## Still to do
* Discuss and implement an interface for modifying which types are considered binary, as well as choose a small set of defaults. I picked a small number of defaults for my tests and nothing more.
* Documentation

cc @jamesls @kyleknap

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@ Next Release (TBD)
   (`#351 <https://github.com/awslabs/chalice/pull/351>`__)
 * Fix vendor directory contents not being importable locally
   (`#350 <https://github.com/awslabs/chalice/pull/350>`__)
+* Add support for binary payloads
+  (`#352 <https://github.com/awslabs/chalice/pull/352>`__)
+
 
 0.8.2
 =====
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ Next Release (TBD)
 * Fix vendor directory contents not being importable locally
   (`#350 <https://github.com/awslabs/chalice/pull/350>`__)
 * Add support for binary payloads
-  (`#352 <https://github.com/awslabs/chalice/pull/352>`__)
+  (`#348 <https://github.com/awslabs/chalice/issues/348>`__)
 
 
 0.8.2
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -18,6 +18,25 @@ from collections import Mapping
 _PARAMS = re.compile(r'{\w+}')
 
 
+def prepare_response_for_apigateway(response, binary_types):
+    response_dict = response.to_dict()
+    response_headers = CaseInsensitiveMapping(response_dict['headers'])
+    content_type = response_headers.get('content-type', '')
+    body = response_dict['body']
+
+    # Encode the response as binary if required.
+    # To keep json types working as expected they are encoded to a utf-8
+    # byte sequence if the content type is application/json before being
+    # converted to a base64 string.
+    if _matches_content_type(content_type, binary_types):
+        if content_type.startswith('application/json'):
+            body = body.encode('utf-8')
+        body = _base64encode(body)
+        response_dict['isBase64Encoded'] = True
+    response_dict['body'] = body
+    return response_dict
+
+
 def handle_decimals(obj):
     # Lambda will automatically serialize decimals so we need
     # to support that as well.
@@ -30,6 +49,21 @@ def error_response(message, error_code, http_status_code):
     body = {'Code': error_code, 'Message': message}
     response = Response(body=body, status_code=http_status_code)
     return response.to_dict()
+
+
+def _base64encode(data):
+    if not isinstance(data, bytes):
+        raise ValueError('Expected bytes type for body with binary '
+                         'Content-Type. Got %s type body instead.'
+                         % type(data))
+    data = base64.b64encode(data)
+    return data.decode('ascii')
+
+
+def _matches_content_type(content_type, valid_content_types):
+    if ';' in content_type:
+        content_type = content_type.split(';', 1)[0].strip()
+    return content_type in valid_content_types
 
 
 class ChaliceError(Exception):
@@ -236,13 +270,13 @@ class Request(object):
         self.headers = CaseInsensitiveMapping(headers)
         self.uri_params = uri_params
         self.method = method
-        if is_base64_encoded:
-            body = self._base64decode(body)
-        self.raw_body = body
+        self._is_base64_encoded = is_base64_encoded
+        self._body = body
         #: The parsed JSON from the body.  This value should
         #: only be set if the Content-Type header is application/json,
         #: which is the default content type value in chalice.
         self._json_body = None
+        self._raw_body = None
         self.context = context
         self.stage_vars = stage_vars
 
@@ -251,6 +285,17 @@ class Request(object):
             encoded = encoded.encode('ascii')
         output = base64.b64decode(encoded)
         return output
+
+    @property
+    def raw_body(self):
+        if self._raw_body is None:
+            if self._is_base64_encoded:
+                self._raw_body = self._base64decode(self._body)
+            elif not isinstance(self._body, bytes):
+                self._raw_body = self._body.encode('utf-8')
+            else:
+                self._raw_body = self._body
+        return self._raw_body
 
     @property
     def json_body(self):
@@ -273,39 +318,17 @@ class Response(object):
         if headers is None:
             headers = {}
         self.headers = headers
-        self.case_insensitive_headers = CaseInsensitiveMapping(headers)
         self.status_code = status_code
-        self._is_base64_encoded = False
-
-    def _base64encode(self, data):
-        if not isinstance(data, bytes):
-            raise ValueError('Expected bytes type for body with binary '
-                             'Content-Type. Got %s type body instead.'
-                             % type(data))
-        data = base64.b64encode(data)
-        return data.decode('ascii')
-
-    def encode_as_binary(self):
-        # To keep json types working as expected they are encoded to a utf-8
-        # byte sequence if the content type is application/json before being
-        # converted to a base64 string.
-        if self.case_insensitive_headers.get('content-type', '').startswith(
-                'application/json'):
-            self.body = json.dumps(self.body).encode('utf-8')
-        self.body = self._base64encode(self.body)
-        self._is_base64_encoded = True
 
     def to_dict(self):
         body = self.body
+        if not isinstance(body, str) and not isinstance(body, bytes):
+            body = json.dumps(body, default=handle_decimals)
         response = {
             'headers': self.headers,
             'statusCode': self.status_code,
+            'body': body
         }
-        if self._is_base64_encoded:
-            response['isBase64Encoded'] = True
-        elif not isinstance(body, str):
-            body = json.dumps(body, default=handle_decimals)
-        response['body'] = body
         return response
 
 
@@ -353,9 +376,20 @@ class APIGateway(object):
 
     _DEFAULT_BINARY_TYPES = [
         'application/octet-stream',
+        'application/x-tar',
+        'application/zip',
+        'audio/basic',
+        'audio/ogg',
+        'audio/mp4',
+        'audio/mpeg',
+        'audio/wav',
+        'audio/webm',
         'image/png',
         'image/jpg',
-        'image/gif'
+        'image/gif',
+        'video/ogg',
+        'video/mpeg',
+        'video/webm',
     ]
 
     def __init__(self):
@@ -429,20 +463,6 @@ class Chalice(object):
             return view_func
         return _register_view
 
-    def _validate_route_content_types(self, name, content_types):
-        # A routes content_types be homogeneous in their binary support.
-        binary, non_binary = [], []
-        for content_type in content_types:
-            if content_type in self.api.binary_types:
-                binary.append(content_type)
-            else:
-                non_binary.append(content_type)
-        if binary and non_binary:
-            raise ValueError(
-                'In view function "%s", the content_types %s support binary '
-                'and %s do not. All content_types must be consistent in their '
-                'binary support.' % (name, binary, non_binary))
-
     def _add_route(self, path, view_func, **kwargs):
         name = kwargs.pop('name', view_func.__name__)
         methods = kwargs.pop('methods', ['GET'])
@@ -458,8 +478,6 @@ class Chalice(object):
         if kwargs:
             raise TypeError('TypeError: route() got unexpected keyword '
                             'arguments: %s' % ', '.join(list(kwargs)))
-        self._validate_route_content_types(name, content_types)
-
         if path in self.routes:
             raise ValueError(
                 "Duplicate route detected: '%s'\n"
@@ -505,8 +523,8 @@ class Chalice(object):
         if route_entry.content_types:
             content_type = self.current_request.headers.get(
                 'content-type', 'application/json')
-            if not self._matches_content_type(content_type,
-                                              route_entry.content_types):
+            if not _matches_content_type(content_type,
+                                         route_entry.content_types):
                 return error_response(
                     error_code='UnsupportedMediaType',
                     message='Unsupported media type: %s' % content_type,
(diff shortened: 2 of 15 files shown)


chalice/app.py
```diff
@@ -17,6 +18,13 @@
 _PARAMS = re.compile(r'{\w+}')
 
 
+def base64decode(encoded):
```
jamesls: I only see this being used in one spot.  I'd prefer to just move it into the class that's using it.

chalice/app.py
```diff
@@ -266,15 +274,29 @@ def __init__(self, body, headers=None, status_code=200):
...
         body = self.body
-        if not isinstance(body, str):
-            body = json.dumps(body, default=handle_decimals)
-        return {
+        response = {
             'headers': self.headers,
             'statusCode': self.status_code,
-            'body': body,
         }
+        if self.headers.get('Content-Type') in binary_types:
```
jamesls: I think we should use the case insensitive mapping like we do for incoming requests.  Also we should think about smarter content type matching.  There were a few bugs related to that for the content type matching on receiving a request in `Chalice.__call__` (stuff about charsets I believe).  Ideally we reuse the same logic throughout the code base for content type matching

chalice/app.py
```diff
@@ -266,15 +274,29 @@ def __init__(self, body, headers=None, status_code=200):
...
+            body = self._base64encode(body)
+            response['isBase64Encoded'] = True
+        elif not isinstance(body, str):
```
jamesls: It's admittedly an edge case, but shouldn't we allow a user to mark `application/json` as a binary type and yet still return a dict/list, etc:

```
app.api.whatever_the_api_is(['application/json'])

@app.route('/')
def foo():
   return {"this will be json": "bar"}
```

Weird use case, but it seems odd that we should start breaking if they happen to mark that as a binary type.

chalice/app.py
```diff
@@ -426,11 +466,14 @@ def __call__(self, event, context):
         view_function = route_entry.view_function
         function_args = [event['pathParameters'][name]
                          for name in route_entry.view_args]
+        body = event['body']
+        if event.get('isBase64Encoded'):
+            body = base64decode(body)
```
jamesls: Given the request object handles json loading, seems reasonable to move this decoding responsibility into the `Request` class as well.

jamesls:
I'm looking at this now, but high level question.  I think I'd prefer to move the docs out to the proper `docs/` directory either under the 'views' topic or its own new topic.  The README is started to get pretty long and the tutorial stuff was intended to show case some of the key parts of the framework.

What do you all think?

cc @stealthycoin @kyleknap

tests/integration/testapp/app.py
```diff
@@ -73,6 +73,17 @@ def form_encoded():
     }
 
 
+@app.route('/binary', methods=['POST'],
```
jamesls: Can you add an integration test that exercises the `.api.binary_types` functionality?

stealthycoin: Yep

CHANGELOG.rst
```diff
@@ -15,6 +15,9 @@ Next Release (TBD)
   (`#351 <https://github.com/awslabs/chalice/pull/351>`__)
 * Fix vendor directory contents not being importable locally
   (`#350 <https://github.com/awslabs/chalice/pull/350>`__)
+* Add support for binary payloads
+  (`#352 <https://github.com/awslabs/chalice/pull/352>`__)
```
jamesls: Crap, I should have caught this earlier, but you can all link to the proposals instead of the pull requests.  I think that's more useful for people browsing the changelog.

chalice/app.py
```diff
@@ -317,12 +349,30 @@ def __eq__(self, other):
         return self.__dict__ == other.__dict__
 
 
+class APIGateway(object):
```
jamesls: Please document this in the `docs/source/api.rst`.  We should at least tell people what the default list of binary types is.

kyleknap:
@jamesls  +1 on moving this to a doc section.

While detailed README's are helpful, they are just a getting started document. When it starts getting overly detailed, people stop reading through it. Plus this in general should be in the documentation on read the docs and not only in the readme.

stealthycoin:
Sure I think I can move it to view.rst and maybe make a brief callout to the fact that binary is supported in the README under `request content types/customizing the http response` sections.

chalice/app.py
```diff
@@ -379,6 +429,20 @@ def _register_view(view_func):
             return view_func
         return _register_view
 
+    def _validate_route_content_types(self, name, content_types):
```
jamesls: Can we move this to the validation already in deployer.py?  I'd prefer to actually move validation out of app and into deployment (in general).  Along the same lines as the previous comment, we want to minimize overhead as much as possible, and we don't need to be doing validation every time the app is imported.

chalice/app.py
```diff
@@ -379,6 +429,20 @@ def _register_view(view_func):
...
+        binary, non_binary = [], []
+        for content_type in content_types:
+            if content_type in self.api.binary_types:
+                binary.append(content_type)
+            else:
+                non_binary.append(content_type)
+        if binary and non_binary:
+            raise ValueError(
+                'In view function "%s", the content_types %s support binary '
+                'and %s do not. All content_types must be consistent in their '
```
jamesls: Why is this the case?  Is this a limitation with API gateway?

chalice/app.py
```diff
@@ -230,11 +231,13 @@ class Request(object):
...
 
     def __init__(self, query_params, headers, uri_params, method, body,
-                 context, stage_vars):
+                 context, stage_vars, is_base64_encoded):
         self.query_params = query_params
         self.headers = CaseInsensitiveMapping(headers)
         self.uri_params = uri_params
         self.method = method
+        if is_base64_encoded:
+            body = self._base64decode(body)
```
jamesls: We should be consistent with how json body is handled.  Any kind of processing on data should be done lazily.  We want the runtime to have as minimal overhead as possible.

chalice/app.py
```diff
@@ -264,17 +273,40 @@ def __init__(self, body, headers=None, status_code=200):
         if headers is None:
             headers = {}
         self.headers = headers
+        self.case_insensitive_headers = CaseInsensitiveMapping(headers)
```
jamesls: Why are we adding a case insensitive headers?  Seems confusing to have both, and it's not obvious when you'd use one or the other.  I suggest we have one attr, `headers`, and that's it.

chalice/app.py
```diff
@@ -264,17 +273,40 @@ def __init__(self, body, headers=None, status_code=200):
...
+
+    def _base64encode(self, data):
+        if not isinstance(data, bytes):
+            raise ValueError('Expected bytes type for body with binary '
+                             'Content-Type. Got %s type body instead.'
+                             % type(data))
+        data = base64.b64encode(data)
+        return data.decode('ascii')
+
+    def encode_as_binary(self):
```
jamesls: So I read the comment, looked at `test_invalid_binary_response_body_throws_value_error`, and I still don't understand why this is needed.  I suspect customers will also be confused what this public method does and why it's here.

chalice/app.py
```diff
@@ -317,12 +349,30 @@ def __eq__(self, other):
         return self.__dict__ == other.__dict__
 
 
+class APIGateway(object):
+
+    _DEFAULT_BINARY_TYPES = [
```
jamesls: We should evaluate some of the common binary types.  I realize there's a limit, but I think we should add more common types here besides images, especially now that customers can override the list if they want complete control.

jamesls:
Cool, and also I'd suggest fleshing out the docs in more detail.  Now that it's in a separate topic guide, you can go into much more detail, show how to expand the list of binary values, maybe give more realistic examples (use an image or something), etc.

stealthycoin:
Yea sounds good to me, I was being overly concise due to it being in a "tutorial". I really need to describe the 6mb lambda limit as well.

tests/unit/test_app.py
```diff
@@ -77,6 +78,49 @@ def name(name):
     return demo
 
 
+def test_mixed_content_types():
+    demo = app.Chalice('demo-app')
+
+    with pytest.raises(ValueError):
+        @demo.route('/index', content_types=['application/octet-stream',
+                                         'text/plain'])
```
jamesls: Hmm, I'm guessing we're not running flake8 on our tests...we should probably get that updated (as a separate PR).

chalice/app.py
```diff
@@ -450,7 +516,11 @@ def __call__(self, event, context):
                                                     function_args)
         if self._cors_enabled_for_route(route_entry):
             self._add_cors_headers(response, route_entry.cors)
-        return response.to_dict()
+        if self._matches_content_type(response.case_insensitive_headers.get(
```
jamesls: I don't see the _matches_content_type being tested with response headers.  If it isn't tested, can you add something that verifies headers with the `; extra-stuff` appended will still match appropriately?

tests/unit/test_app.py
stealthycoin: Yea I manually run it now and then. I've caught quite a few mistakes in the existing tests, but I don't consistently remember to run it.

chalice/app.py
stealthycoin: Ah, that was an existing function that was already tested I thought.

chalice/app.py
stealthycoin: Yea I added a decent number more after this PR which will be in the next version.

chalice/app.py
stealthycoin: Not technically, but it changes the interface and I couldn't think of any good use to have that rather than just writing two separate routes to handle your data accordingly. And it could be a source of bugs if you are not careful, since if you do not read the documentation carefully and you mix the types you may get rare crashes which may not be immediately clear.

I believe if you send a bytes type to lambda and it isn't base64 encoded properly then you get a "circular reference" error during json encoding which isn't very clear. As a user of chalice I think I would prefer to simply be called out on this ahead of time and be forced to separate my logic into two routes.

Maybe there is a common use case I'm not thinking of?

chalice/app.py
stealthycoin: Similar to the other comment, not sure if I am missing/making up use cases. But I thought it could be useful in the case where a user wants to return binary data that is received as a base64 string rather than the raw bytes. For example something that is intended to be printed straight to an ascii term. 

Maybe thats a contrived use case?

chalice/app.py
stealthycoin: Ah sure. I hadn't seen those, I just put it with the other route validation `ValueErrors` I may strip this check out entirely if it isn't deemed useful.

chalice/app.py
stealthycoin: Sure.

CHANGELOG.rst
stealthycoin: Yep.

tests/unit/test_app.py
jamesls: Yeah I'd say for now, just make sure the tests your adding pass flake8.  I'd prefer not to see a huge diff of flake8 fixups in this PR and just handle it as a separate PR.

chalice/app.py
jamesls: Yes, the underlying function is tested, but your use case is not given this is new code.  We're testing that "the content type value returned by a user can have '; extra-stuff' in it and still match" regardless of however the underlying functionality, `_matches_content_type` has been tested.

chalice/app.py
jamesls: Could you elaborate on how it change the interface? Seems like you should be able to say:

```
@app.route(..., content_types=['text/plain', 'image/png'])
def foo():
    if a:
        return Response('text content', headers={'Content-Type': 'text/plain')
    return Response(image_binary-data, headers={'Content-Type': 'img/png')
```

stealthycoin: `raw_body` will be `bytes` if it was triggered by a binary `Content-Type` under py3. 

Lets say my api is for resizing images. So I have something like this:
```python
@app.route(..., content_types=['image/tff', 'image/png', ...])
def foo():
    resized = _resize_image_and_make_png(app.current_request.raw_body)
    return Response(status_code=200,
                 body=resized,
                 headers={'Content-Type': 'image/png'})
```

I may think everything is fine for a long time because tff isn't super common and maybe I don't do an integration test using that format, and when I send a tiff up it will error out. Because I am passing a string to the resized function. I may even have trouble debugging this issue because inspecting this function, everything appears to be fine. It's not until my assumption that `image/tff` is treated as a binary type is broken that I will even look for documentation regarding what is and is not a binary type. I may never have even looked at the docs regarding this in the past since binary will 'just work' for the most part. 

This error message prevents this entire class of fault, and calls out which types are mismatched. With this information I will be prompted to go look at the documentation regarding which types aren't binary, and will be lead to the docs regarding `app.api.binary_types`. With that information I can align the content-types I accept with the binary_types.

jamesls: So is the issue that `.raw_body` might be `bytes()` or sometimes `str()`?  This came up in a semi-related issue: https://github.com/awslabs/chalice/issues/355, and I'm slightly inclined to make `.raw_body` always be `bytes()`.  It's what I'd expect based on the name...

stealthycoin: So that doesn't really fix the issue though. Because for `image/tff` if I send the `Accept: image/tff` header API gateway will send me back base64 encoded string instead of the bytes for that image. Whereas for all the other types it will send me. 

I'm not against forcing the `raw_body` to be `bytes`, I think that makes sense, but I think there is also a difference in behavior that is very subtle and difficult do debug. I would rather be explicit about it.

jamesls:
Pushed a few changes, let me know if there's any feedback.

chalice/app.py
```diff
@@ -266,15 +292,42 @@ def __init__(self, body, headers=None, status_code=200):
...
+        response_headers = CaseInsensitiveMapping(response_dict['headers'])
+        content_type = response_headers.get('content-type', '')
+        body = response_dict['body']
+
+        # Encode the response as binary if required.
+        # To keep json types working as expected they are encoded to a utf-8
+        # byte sequence if the content type is application/json before being
+        # converted to a base64 string.
+        if _matches_content_type(content_type, binary_types):
+            if _matches_content_type(content_type, ['application/json']):
```
kyleknap: Even with the comment, I am still having trouble following why there is this special case for application/json and not any other content types. Specifically, I am not sure what "working as expected" is suppose to mean.

chalice/app.py
```diff
@@ -450,12 +532,35 @@ def __call__(self, event, context):
...
+    def _validate_binary_response(self, request_headers, response_headers):
+        # Validates that a response is valid given the request. If the response
+        # content-type specifies a binary type, there must be an accept header
+        # that is a binary type as well.
+        request_accept_header = request_headers.get('accept')
+        response_content_type = response_headers.get(
+            'content-type', 'application/json')
+        response_is_binary = response_content_type in self.api.binary_types
+        request_accepts_binary = request_accept_header in self.api.binary_types
+        if response_is_binary != request_accepts_binary:
```
kyleknap: This could be simplified to:
```python
return response_is_binary and requests_accepts_binary
```

chalice/app.py
```diff
@@ -266,15 +292,42 @@ def __init__(self, body, headers=None, status_code=200):
         self.headers = headers
         self.status_code = status_code
 
-    def to_dict(self):
+    def to_dict(self, binary_types=None):
```
kyleknap: Is this method typically used by the end user? I wonder if it makes sense to expose a ``binary_types`` parameter as a user typically does not have to worry about specifying what is a binary type; they just need to specify the content type.

chalice/app.py
stealthycoin: JSON is automatically handled for the user, so there is a special case when they make `application/json` a binary type. Since the contract for `application/json` is you can send any json serializable type as the `body` and it will work. But for a binary type you must send a `bytes()` type. So if you specify `application/json` as a binary type, we do the conversion first, which results in a string. Not base64 encodable in py3. Since JSON is utf-8 we can encode it using that to a bytes type and then base64 encode that.

That way you can still send a dict/list/jsonish type as a body and have a json binary type and it will still do the right thing.

chalice/deploy/deployer.py
```diff
@@ -59,6 +59,7 @@ def validate_configuration(config):
     """
     routes = config.chalice_app.routes
     validate_routes(routes)
+    validate_route_content_types(routes, config.chalice_app.api.binary_types)
```
stealthycoin: I am not sure about this either.  I went with public for mine since 2/3 were public.

jamesls: There's not really a strong reason.  I think a reasonable refactoring in the future would be to move all of this over to a validation class with each of these being subclasses.

Initially there was only a handful of validations.  There was a prior discussion above about moving more validation that can be done at deploy time out of the runtime critical path.

chalice/app.py
jamesls: No, this is not typically used by an end user (but could be if they wanted), it's used by the chalice handler.

jamesls:
@stealthycoin Pushed a fix for the binary response validation (https://github.com/awslabs/chalice/pull/352/commits/6790b6cabe611a800087341ce837bb82812c6c7f)

chalice/deploy/deployer.py
kyleknap: Yeah that would make sense to move these out to subclasses in the future given we are probably only going to be adding more validations. But it does not need to be addressed in this PR.

chalice/app.py
kyleknap: Cool. Could you add this information to the comment? Not sure if you need the whole thing but mainly clarifying the "working as expected" part.

chalice/app.py
kyleknap: Makes sense. The only other place that you could put it is in the initializer, but that is something the user uses directly so it would be weird to add a binary_types there.

chalice/app.py
jamesls: I pushed an update that adds a comment.  @kyleknap was there anything I missed?

PR #353, opened 2017-05-24
Change update_function behavior to be PUT

kyleknap:
This fixes any bug related to if a user specifies a non-default `timeout` or `memory_size`, deploys the application, then removes the `timeout` or `memory_size` configuration and redeploys the application, the `timeout` or `memory_size` would not go back to being a the default values.


chalice/awsclient.py
```diff
@@ -79,7 +80,7 @@ def create_function(self,
         # type: (...) -> str
         kwargs = {
             'FunctionName': function_name,
-            'Runtime': 'python2.7',
+            'Runtime': DEFAULT_LAMBDA_RUNTIME,
```
jamesls: Should we even have this as a default value?  I think it should always match whatever version of python you're using and perhaps even be a required param.  The deployer will always pass this value anyways.

kyleknap: I think it should always be included in both the CreateFunction (which it is required) and the UpdateFunction (which it is should be required based on the PUT semantics we want). Whether it is through a default or making it required, I do not have too strong of an opinion. What are the downsides with making it default? Is that you may be unknowingly change the runtime if you omit it?

kyleknap:
@jamesls I would not be opposed to having integration tests for it. How exhaustive would you think these would need to be?

chalice/awsclient.py
stealthycoin: +1 to required, always current python version, no default value.

jamesls:
@kyleknap  I'd suggest just starting with the `update_function` since that's what triggered this PR.

chalice/awsclient.py
jamesls: >  Is that you may be unknowingly change the runtime if you omit it?

Yeah that's the main downside.  I can't think of a use case where, if I'm using python3, I'd silently want the runtime to change to 2.7.

kyleknap: Fair enough I will make that required for create_function and update_function.

JordonPhillips:
Looks good to me. Are we still gonna go for integration tests?

kyleknap:
Closing this PR out. I like my other approach in this PR: https://github.com/awslabs/chalice/pull/357 much more and is much more thorough with tests.

PR #357, opened 2017-05-25
Add chalice default logic to deployer

kyleknap:
Originally implemented in: https://github.com/awslabs/chalice/pull/353, but did not like how the awsclient was intertwined with chalice specific logic. This made it difficult to test the actual values sent to AWS Lambda. Most of the code added are test specific so we now should have a lot better test coverage in this portion of the code.

### DIFF
--- a/tests/unit/deploy/test_deployer.py
+++ b/tests/unit/deploy/test_deployer.py
@@ -613,7 +613,14 @@ def test_cant_have_options_with_cors(sample_app):
 
 class TestLambdaInitialDeploymentWithConfigurations(object):
     @fixture(autouse=True)
-    def setup(self, app_policy):
+    def setup_deployer_dependencies(self, app_policy):
+        # This autouse fixture is used instead of ``setup_method`` because it:
+        # * Is ran for every test
+        # * Allows additional fixtures to be passed in to reduce the number
+        #   of fixtures that need to be supplied for the test methods.
+        # * ``setup_method`` is called before fixtures get applied so
+        #   they cannot be applied to the ``setup_method`` or you will
+        #   will get a TypeError for too few arguments.
         self.package_name = 'packages.zip'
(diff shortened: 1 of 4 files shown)


tests/unit/deploy/test_deployer.py
```diff
@@ -607,3 +609,284 @@ def badview():
 
     with pytest.raises(ValueError):
         validate_routes(sample_app.routes)
+
+
+class TestLambdaInitialDeploymentWithConfigurations(object):
+    @fixture(autouse=True)
```
jamesls: Why is autouse being used here?  Seems like you can just use the standard `unittest.TestCase` here right?

kyleknap: So I pass in the `app_policy` fixture in order to reduce the extra arguments in the test methods by putting it in a `setup()` method. Originally I tried to use the `setup_method` which is the same as `unittest.TestCase.setUp` for pytest, but these methods get called before the fixtures are applied/injected so you get errors like:
```
TypeError: setUp() takes exactly 2 arguments (1 given)
```
when I try to supply the `app_policy` fixture.

Therefore by making it an `autouse=True`, I can pass in any fixture to this `setup` method allowing me to reduce the number of fixtures I have to define in the test methods and have the `setup()` be ran for every test.

jamesls: Ah then please call it something else.  I think people seeing a setup() function in a test class would expect it to behave like unittest.TestCase and wonder why that's not being used instead.  Might be helpful to add a comment as well.

kyleknap: Sounds good.

kyleknap:
@jamesls I added comments and renamed the method fixture. Will probably merge after the travis build passes.

jamesls:
👍

PR #358, opened 2017-05-25
Add documentation for the package command

jamesls:
This was the last action item to close out https://github.com/awslabs/chalice/issues/183

cc @kyleknap @stealthycoin

### DIFF
--- a/docs/source/topics/cfn.rst
+++ b/docs/source/topics/cfn.rst
@@ -1,0 +1,123 @@
+AWS CloudFormation Support
+==========================
+
+When you run ``chalice deploy``, chalice will deploy your application using the
+`AWS SDK for Python <http://boto3.readthedocs.org/>`__).  Chalice also provides
+functionality that allows you to manage deployments yourself using
+cloudformation.  This is provided via the ``chalice package`` command.
+
+When you run this command, chalice will generate the AWS Lambda deployment
+package that contains your application as well as a `Serverless Application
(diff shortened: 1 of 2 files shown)


jamesls:
>  Do you think it would be worthwhile to add a small helper script to chalice to make SAM deployments a bit less tedious?

Personally no.  This is more of an advanced feature if you have existing cfn tooling you prefer to use *or* if you want to integrate with codepipeline (in which case the generate-pipeline command has helper stuff for you).

If we were going to go down that route, I'd prefer we spend the time to just have a cfn based backend, but it's not something I think we should do right at this moment.

stealthycoin:
You mention pipeline support but don't link to or elaborate on it, I think if it is mentioned in the header of the section it should be elaborated upon at some point or linked to another section at least.

jamesls:
@stealthycoin Yeah that's because there is no pipeline topic guide yet.  I think that's something we should add but out of scope for this PR.

docs/source/topics/cfn.rst
```diff
@@ -0,0 +1,123 @@
...
+    $ . /tmp/venv/bin/activate
+    $ pip install chalice awscli
+    $ chalice new-project test-cfn-deploy
+    $ cd test-cfn-deploy
+
+At this point we've installed chalice and the AWS CLI and we have
+a basic app created locally.  Next we'll run the ``package`` command
+and look at its contents::
+
+    $ $ chalice package /tmp/packaged-app/
```
pickfire: $ $

PR #363, opened 2017-05-31
Fix issue in updating iam role arn for deployments

kyleknap:
This fixes two issues related to providing an iam role:

* https://github.com/awslabs/chalice/issues/339: Where am `iam_role_arn` is added after the first deployment and is not used in the update.
* For the package command, the `iam_role_arn` was straight up not even accounted for when it is a valid value that can be passed to the SAM template via the [Role property](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#properties).


chalice/awsclient.py
```diff
@@ -90,22 +90,27 @@ def create_function(self,
             kwargs['Timeout'] = timeout
         if memory_size is not None:
             kwargs['MemorySize'] = memory_size
+        return self._create_or_update_function_with_retries(
+            'create_function', kwargs)['FunctionArn']
```
jamesls: * You can technically pass any client method here so the name seems too restrictive.
* We don't need getattr logic, it's always going to be a method we know statically.  I'd prefer to update this to just pass the method obj directly.  (`self._client('lambda').create_function`)

jamesls:
Looks good, just had a comment about `_create_or_update_function_with_retries`.

chalice/awsclient.py
```diff
@@ -90,22 +90,27 @@ def create_function(self,
             kwargs['Timeout'] = timeout
         if memory_size is not None:
             kwargs['MemorySize'] = memory_size
+        return self._call_client_method_with_retries(
+            self._client('lambda').create_function, kwargs)['FunctionArn']
+
+    def _call_client_method_with_retries(self, method, kwargs):
+        # type: (Callable[..., Any], Dict[str, Any]) -> Dict[str, Any]
```
jamesls: Can we tighten up the signature here?  Callable that only takes kwargs and returns a dict of `str, Any` maybe?  http://mypy.readthedocs.io/en/latest/kinds_of_types.html#extended-callable-types

kyleknap: Right. Clients will always return a dict.

kyleknap:
@jamesls I updated to make stricter type checking on the client method. Decided not to use the KwArg types as that would require pulling in mypy_extensions as a dependency.

chalice/awsclient.py
```diff
@@ -154,8 +155,8 @@ def update_function(self,
             kwargs['Role'] = role_arn
         if kwargs:
             kwargs['FunctionName'] = function_name
-            self._create_or_update_function_with_retries(
-                'update_function_configuration', kwargs)
+            self._call_client_method_with_retries(
+                self._client('lambda').update_function_configuration, kwargs)
```
jamesls: You already create a lambda_client in the first line of this function, you shouldn't need to create it again here.

kyleknap:
Updated to use already instantiated client. Travis passed as well. Merging.

PR #364, opened 2017-06-01
Check pip exit code in create_deployment_package

kislyuk:
Is there a reason why pip failures are ignored?

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -42,6 +42,7 @@ class LambdaDeploymentPackager(object):
             p = subprocess.Popen([pip_exe, 'install', '-r', requirements_file],
                                  stdout=subprocess.PIPE)
             p.communicate()
+            assert p.returncode == os.EX_OK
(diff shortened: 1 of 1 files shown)


jamesls:
Good catch, we should be failing if this is the case.  I'll add a test for this and get it merged.  Thanks for the pull request.

stealthycoin:
This piece of code has been completely re-written and does handle relevant pip error codes now so I'm going to close this PR. Please open issues for any bugs you find with the new deployment system.

kislyuk:
I believe this bug is still there, although now it has a different scope of consequences (the wheel dependency graph scanning process can be silently truncated).

https://github.com/aws/chalice/blob/3c758919063a8e21424fb898bd2bffc2062b2c16/chalice/deploy/packager.py#L587

While it appears as though the error code is passed on, in fact pip errors are ignored (other than gross ones like if pip is missing entirely). To fix this, you'd need to wrap `pip.main()` in `sys.exit(pip_main())`.

kislyuk:
Filed https://github.com/aws/chalice/issues/497

PR #367, opened 2017-06-05
Changed autogen_policy default value from True to None

### LINKED ISSUE #354: autogen_policy from config.json is ignored when doing a chalice package
When we do a "chalice package", we noticed that the autogen_policy was defaulting to true, completely ignoring the '"autogen_policy": false' in the config.json. After debugging, we found that the "create_config_obj" function (called in the 'package' cli command) defaulting autogen_policy to true.

stealthycoin:
Fixes #354

cc @kyleknap @jamesls 

closes #281


jamesls:
There's also another PR for this: https://github.com/awslabs/chalice/pull/281/files

It removes the default of `True` in the CLI command.  Do we need that here as well?

CHANGELOG.rst
```diff
@@ -9,6 +9,9 @@ Next Release (TBD)
   redeployments of chalice applications and in the CloudFormation template
   generated by ``chalice package``
   (`#339 <https://github.com/awslabs/chalice/issues/339>`__)
+* Fix autogen_policy in config being ignored
```
kyleknap: Probably want to make ``autogen_policy`` be a code literal in the changelog.

stealthycoin:
Added those tests cases.

CHANGELOG.rst
```diff
@@ -9,7 +9,7 @@ Next Release (TBD)
   redeployments of chalice applications and in the CloudFormation template
   generated by ``chalice package``
   (`#339 <https://github.com/awslabs/chalice/issues/339>`__)
-* Fix autogen_policy in config being ignored
+* Fix `autogen_policy` in config being ignored
```
kyleknap: Oh it's rst not markdown. That gets me as well. Probably want to use `` instead of `

stealthycoin: Oops

PR #372, opened 2017-06-06
Update interfaces to multi lambda support

jamesls:
* Don't hard code the lambda handler string
* Replace anything that accepted a single lambda arn with a `deployed_resources` dict.

This updates the cases where the deployer accepts a single lambda
ARN.  This is going to have with multi lambda function, so this plumbs
through the change to instead accept a dict.  This gives some
flexibility because we change the dict keys if needed without having
to update all these interfaces again.

Should make rebasing/shared branches easier.

No function changes.

### DIFF
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -198,19 +198,20 @@ class Deployer(object):
         existing_resources = config.deployed_resources(chalice_stage_name)
         deployed_values = self._lambda_deploy.deploy(
             config, existing_resources, chalice_stage_name)
+        deployed_values.update({
+            'backend': self.BACKEND_NAME,
+            'chalice_version': chalice_version,
+        })
         rest_api_id, region_name, apig_stage = self._apigateway_deploy.deploy(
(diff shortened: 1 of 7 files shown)


chalice/deploy/deployer.py
```diff
@@ -198,19 +198,20 @@ def deploy(self, config, chalice_stage_name=DEFAULT_STAGE_NAME):
         existing_resources = config.deployed_resources(chalice_stage_name)
         deployed_values = self._lambda_deploy.deploy(
             config, existing_resources, chalice_stage_name)
+        deployed_values.update({
```
kyleknap: What was the reason for needing to insert these before the api gateway deploy as opposed to adding them at the end?

jamesls: Not needed for a specific feature, just general cleanup.  Fill in all the values we now up front, and only update the deployed values with whatever the current "deployer" creates.

kyleknap: Ok cool. Sounds good.

PR #373, opened 2017-06-06
Prevent creation of invalid pipeline templates.

stealthycoin:
Currently CodeBuild does not support python 3.6, Chalice projects
that use generate-pipeline will create a pipeline that will result in a
python 2.7 lambda function. Until support is added in CodeBuild trying
to create a template in Chalice should result in an error.

cc @jamesls @kyleknap

### DIFF
--- a/chalice/pipeline.py
+++ b/chalice/pipeline.py
@@ -9,10 +9,23 @@ from chalice import constants
 def create_pipeline_template(config):
     # type: (Config) -> Dict[str, Any]
     pipeline = CreatePipelineTemplate()
-    return pipeline.create_template(config.app_name)
+    return pipeline.create_template(config.app_name,
+                                    config.lambda_python_version)
+
+
+class InvalidCodeBuildPythonVersion(Exception):
+    def __init__(self, version):
+        # type: (str) -> None
+        super(InvalidCodeBuildPythonVersion, self).__init__(
+            'CodeBuild does not yet support python version %s.' % version
+        )
 
 
 class CreatePipelineTemplate(object):
+
+    _CODEBUILD_IMAGE = {
+        'python2.7': 'python:2.7.12',
+    }
 
     _BASE_TEMPLATE = {
         "AWSTemplateFormatVersion": "2010-09-09",
@@ -21,6 +34,11 @@ class CreatePipelineTemplate(object):
                 "Default": "ChaliceApp",
                 "Type": "String",
                 "Description": "Enter the name of your application"
+            },
+            "CodeBuildImage": {
+                "Default": "python:2.7.12",
+                "Type": "String",
+                "Description": "Name of codebuild image to use."
             }
         },
         "Resources": {},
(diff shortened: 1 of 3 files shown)


chalice/pipeline.py
```diff
@@ -9,18 +9,36 @@
...
+
     _BASE_TEMPLATE = {
         "AWSTemplateFormatVersion": "2010-09-09",
         "Parameters": {
             "ApplicationName": {
                 "Default": "ChaliceApp",
                 "Type": "String",
                 "Description": "Enter the name of your application"
+            },
+            "CodeBuildImage": {
```
jamesls: If we're going to parameterize the actual code build image anyways, shouldn't we let users also specify custom build images as well?  In other words, maybe we require the entire image identifier instead of always assuming `aws/codebuild/<param>`.

chalice/pipeline.py
```diff
@@ -31,10 +49,21 @@ def __init__(self):
...
-        # type: (str) -> Dict[str, Any]
+    def _codebuild_image(self, lambda_python_version):
+        # type: (str) -> str
+        try:
+            image = self._CODEBUILD_IMAGE[lambda_python_version]
+            return image
+        except KeyError as e:
+            raise InvalidCodeBuildPythonVersion(str(e))
+
+    def create_template(self, app_name, python_lambda_version):
```
jamesls: This is fine for now, but one of the things that I've noticed is that these arguments tend to grow over time as more things are supported.  They eventually just end up taking `config`.  Just something to keep in mind for future cases.

chalice/pipeline.py
stealthycoin: Yea thats a good point. I could pass through some cli argument like `--codebuild-image` which would override whatever the auto selected one. Which would be nice in case a user has their own python 3.6 image. Although the only use case I can think of for this is to use your own image when CodeBuild hasn't caught up to Lambda, so the reason for this flag's existence would have to be documented well.

jamesls: Well I was just thinking even if they're deploying the cfn template manually, it's still something they can parameterize themselves (say using `aws cloudformation deploy`).  I wouldn't think we would need to add ` --codebuild-image` ourselves.

chalice/pipeline.py
```diff
@@ -9,18 +9,36 @@
...
     _BASE_TEMPLATE = {
         "AWSTemplateFormatVersion": "2010-09-09",
         "Parameters": {
             "ApplicationName": {
                 "Default": "ChaliceApp",
                 "Type": "String",
                 "Description": "Enter the name of your application"
+            },
+            "CodeBuildImage": {
+                "Default": "python:2.7.12",
```
jamesls: Don't you need to change this default to `aws/codebuild/python:2.7.12`  now that the `aws/codebuild` isn't used in the template?

stealthycoin: To be consistent yes I should do that. Although that always gets replaced so it won't really have any affect.

PR #375, opened 2017-06-07
Allow view functions to share route url and differ in HTTP method

kyleknap:
This adds support for defining multiple view functions that share the same route path but only differ by HTTP method. Also added support for comparing CORSConfig objects for equality.

Implements https://github.com/awslabs/chalice/issues/81

cc @jamesls @stealthycoin

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -426,9 +426,30 @@ specified resource.  For example:
         pass
 
 The above view function will be called when either an HTTP POST or
-PUT is sent to ``/myview``.  In the next section we'll go over
-how you can introspect the given request in order to differentiate between
-various HTTP methods.
+PUT is sent to ``/myview``.
+
+Alternatively if you do not want to share the same view function across
+multiple HTTP methods for the same route url, you may define separate view
+functions to the same route url but have the view functions differ by
+HTTP method. For example:
+
+.. code-block:: python
+
+    @app.route('/myview', methods=['POST'])
+    def myview_post():
+        pass
+
+    @app.route('/myview', methods=['PUT'])
+    def myview_put():
+        pass
+
+This setup will route all HTTP POST's to ``/myview`` to the ``myview_post()``
+view function and route all HTTP PUT's to ``/myview`` to the ``myview_put()``
+view function.
+
+In the next section we'll go over how you can introspect the given request
+in order to differentiate between various HTTP methods.
+
 
 Tutorial: Request Metadata
 ==========================
--- a/README.rst
+++ b/README.rst
@@ -445,7 +445,9 @@ HTTP method. For example:
 
 This setup will route all HTTP POST's to ``/myview`` to the ``myview_post()``
 view function and route all HTTP PUT's to ``/myview`` to the ``myview_put()``
-view function.
+view function. It is also important to note that the view functions
+**must** have unique names. For example, both view functions cannot be
+named ``myview()``.
 
 In the next section we'll go over how you can introspect the given request
 in order to differentiate between various HTTP methods.
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -486,8 +486,12 @@ class Chalice(object):
             if method in self.routes[path]:
                 raise ValueError(
                     "Duplicate method: '%s' detected for route: '%s'\n"
-                    "A specific method may only be specified once for "
-                    "a particular path." % (method, path))
+                    "between view functions: \"%s\" and \"%s\". A specific "
+                    "method may only be specified once for "
+                    "a particular path." % (
+                        method, path, self.routes[path][method].view_name,
+                        name)
+                )
             entry = RouteEntry(view_func, name, path, method,
                                authorizer_name, api_key_required,
                                content_types, cors, authorizer)
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -7,7 +7,7 @@ import traceback
 import decimal
 import warnings
 import base64
-from collections import Mapping
+from collections import defaultdict, Mapping
 
 # Implementation note:  This file is intended to be a standalone file
 # that gets copied into the lambda deployment package.  It has no dependencies
@@ -232,6 +232,11 @@ class CORSConfig(object):
 
         return headers
 
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self.get_access_control_headers() == \
+                other.get_access_control_headers()
+
 
 class Request(object):
     """The current request from API gateway."""
@@ -333,14 +338,14 @@ class Response(object):
 
 class RouteEntry(object):
 
-    def __init__(self, view_function, view_name, path, methods,
+    def __init__(self, view_function, view_name, path, method,
                  authorizer_name=None,
                  api_key_required=None, content_types=None,
                  cors=False, authorizer=None):
         self.view_function = view_function
         self.view_name = view_name
         self.uri_pattern = path
-        self.methods = methods
+        self.method = method
         self.authorizer_name = authorizer_name
         self.api_key_required = api_key_required
         #: A list of names to extract from path:
@@ -406,7 +411,7 @@ class Chalice(object):
     def __init__(self, app_name, configure_logs=True):
         self.app_name = app_name
         self.api = APIGateway()
-        self.routes = {}
+        self.routes = defaultdict(dict)
         self.current_request = None
         self.debug = False
         self.configure_logs = configure_logs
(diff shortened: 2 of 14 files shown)


tests/unit/deploy/test_deployer.py
```diff
@@ -603,14 +604,70 @@ def test_lambda_deployer_initial_deploy(app_policy, sample_app):
...
+            pass
+
+        @sample_app.route('/cors', methods=['PUT'], cors=custom_cors)
+        def different_cors():
+            pass
+
+        with pytest.raises(ValueError):
+            validate_routes(sample_app.routes)
+
+    def test_can_have_same_cors_configurations(self, sample_app):
```
stealthycoin: It would be nice to have another test like this one that explicitly tests custom CORSConfig objects that are `__eq__`.

chalice/app.py
```diff
@@ -232,6 +232,11 @@ def get_access_control_headers(self):
 
         return headers
 
+    def __eq__(self, other):
```
stealthycoin: I wonder if `__eq__` or an identity check would be better here. The equality check is sufficient for validating that any configuration that you can setup in Chalice will be compatible with API Gateways handling of CORS preflight requests. However we could take it a step further and do identity checks on the CORSConfig objects, which would give the benefit of forcing users to design their APIs in a way that is logically consistent with API Gateway.

For example consider the following chalice app using this branch:
```python
from chalice import Chalice, CORSConfig

app = Chalice(app_name='foobar')

cors_a = CORSConfig()
cors_b = CORSConfig()

@app.route('/a', methods=['POST'], cors=cors_a)
def route_a_post():
    pass


@app.route('/a', methods=['GET'], cors=cors_a)
def route_a_get():
    pass


@app.route('/b', methods=['POST'], cors=cors_a)
def route_b_post():
    pass


@app.route('/b', methods=['GET'], cors=cors_b)
def route_b_get():
    pass
```

We are logically associating `cors_a` with route `/a` and `cors_b` with `/b`, since that is what API Gateway requires. But we made a typo and put `cors_a` on the `route_b_post` handler. With `__eq__` validation this will pass the validation step and get deployed and everything will work until `cors_a` and `cors_b` diverge. There error message that you get is pretty clear so it should be quick to correct the symptom, but maybe your design didn't take into account this limitation and now you are stuck redesigning some portion of your app's cors handling. 

With identity checking the above example code would throw a validation error and raise this issue with the designer earlier on so that they can plan around it. If they need more control over how the CORS requests are handled, a separate handler can be created that accepts OPTIONS requests and checks the `Access-Control-Request-Method` header to figure out which method is about to be called, and set the CORS headers in that response individually.

So its a relatively minor point but in my opinion identity check would be slightly better. We could change the `cors=True` argument to pass in a pre-created CORSConfig object (or cache it the first time it is created) so that it would pass identity checks.

chalice/deploy/swagger.py
```diff
@@ -176,10 +187,9 @@ def _add_view_args(self, apig_integ, view_args):
             for name in view_args
         ]
 
-    def _add_preflight_request(self, view, swagger_for_path):
-        # type: (RouteEntry, Dict[str, Any]) -> None
-        cors = view.cors
-        methods = view.methods + ['OPTIONS']
+    def _add_preflight_request(self, cors, methods, swagger_for_path):
+        # type: (CORSConfig, List[str], Dict[str, Any]) -> None
+        methods = methods + ['OPTIONS']
```
stealthycoin: Maybe since its not being extracted from the view anymore `+=` or `.append()`?

README.rst
```diff
@@ -426,9 +426,30 @@ specified resource.  For example:
...
+
+    @app.route('/myview', methods=['POST'])
+    def myview_post():
+        pass
+
+    @app.route('/myview', methods=['PUT'])
+    def myview_put():
+        pass
+
+This setup will route all HTTP POST's to ``/myview`` to the ``myview_post()``
```
jamesls: Can you make a note that the functions must have unique names?  Just want to make sure we're explicit here.

README.rst
```diff
@@ -426,9 +426,30 @@ specified resource.  For example:
         pass
 
 The above view function will be called when either an HTTP POST or
-PUT is sent to ``/myview``.  In the next section we'll go over
-how you can introspect the given request in order to differentiate between
-various HTTP methods.
+PUT is sent to ``/myview``.
+
+Alternatively if you do not want to share the same view function across
```
jamesls: Can you update the topic guide (`docs/source/topics/views.rst`) on views with the details?  For example, the various validation we'll perform, the edge cases with CORs, etc.

chalice/app.py
```diff
@@ -477,14 +482,16 @@ def _add_route(self, path, view_func, **kwargs):
...
-        entry = RouteEntry(view_func, name, path, methods,
-                           authorizer_name, api_key_required,
-                           content_types, cors, authorizer)
-        self.routes[path] = entry
+        for method in methods:
+            if method in self.routes[path]:
+                raise ValueError(
+                    "Duplicate method: '%s' detected for route: '%s'\n"
+                    "A specific method may only be specified once for "
+                    "a particular path." % (method, path))
```
jamesls: What do you think about actually printing the function name here to make debugging even easier?

chalice/deploy/deployer.py
```diff
@@ -110,36 +105,53 @@ def validate_python_version(config, actual_py_version=None):
...
+        raise ValueError(
+            'In view function "%s", the content_types %s support binary '
+            'and %s do not. All content_types must be consistent in their '
+            'binary support.' % (route_entry.view_name, binary, non_binary))
+
+
+def _validate_cors_for_route(route_url, route_methods):
+    # type: (str, Dict[str, app.RouteEntry]) -> None
+    entries_with_cors = [
+        entry for _, entry in route_methods.items() if entry.cors
```
jamesls: If you don't need the key you can just use `.values()`.

chalice/deploy/deployer.py
```diff
@@ -110,36 +105,53 @@ def validate_python_version(config, actual_py_version=None):
...
                 "Route entry cannot have both cors=True and "
                 "methods=['OPTIONS', ...] configured.  When "
                 "CORS is enabled, an OPTIONS method is automatically "
                 "added for you.  Please remove 'OPTIONS' from the list of "
                 "configured HTTP methods for: %s" % route_url)
 
+        if not all(entries_with_cors[0].cors == entry.cors for entry in
+                   entries_with_cors):
+            raise ValueError(
+                "Route may not have mulitple differing CORS configurations. "
```
jamesls: typo "mulitple"

tests/unit/test_app.py
```diff
@@ -715,3 +738,21 @@ def test_can_serialize_custom_authorizer():
             'authorizerResultTtlInSeconds': 10,
         }
     }
+
+
+class TestCORSConfig(object):
```
jamesls: Can you add a test where `CORSConfig` objects are equal with non-default params? (add  `allow_headers,expose_header,max_age,allow_credentials`).

chalice/app.py
jamesls: -0 for me.  Given this is just a value object, equality is pretty reasonable to me.  For a more complicated app it might make it easier to not have to track around the exact object you want to apply for a set of views.

tests/unit/deploy/test_deployer.py
kyleknap: Sounds good. I can add that.

chalice/deploy/swagger.py
kyleknap: I would be hesitant to do append as it may cause unintentional side effects if the methods list provided is used later on. I can do `+=` though.

README.rst
kyleknap: Sounds good. I can do that. Should I just add an additional section on multi HTTP method support in the `docs/source/topics/views.rst` that is a copy of this but is more thorough?

README.rst
kyleknap: I can add that as well.

README.rst
jamesls: Yeah that sounds good.  Make sure to include examples as well.

chalice/app.py
kyleknap: Sure. I should be able to have access to both view function names. So it should be feasible.

tests/unit/test_app.py
kyleknap: Yep. I can add that.

README.rst
kyleknap: Alright will do.

chalice/app.py
kyleknap: Yeah I think I want to stick with equality for now. Mainly because it may be a bit of pain to pass the same object around for all of the routes when I should just be able to create two objects that are exactly the same. I think it is important that this is documented at least so users are aware of this behavior when designing their API.

chalice/deploy/swagger.py
jamesls: `+=` will also mutate the list if that's your concern.

kyleknap: Right... I am just going to stick with what was there originally.

kyleknap:
@jamesls @stealthycoin I think I addressed all of the feedback. Let me know what you think.

PR #379, opened 2017-06-13
Fix integration tests related to CORS

kyleknap:
Was making assumption that the allow methods would retain the same order. This caused test to potentially fail on Python 2 as dictionaries were unordered until Python 3.6.

### DIFF
--- a/tests/integration/test_features.py
+++ b/tests/integration/test_features.py
@@ -246,6 +246,13 @@ def test_can_round_trip_binary_custom_content_type(smoke_test_app):
     assert response.content == bin_data
 
 
+def _assert_contains_access_control_allow_methods(headers, methods):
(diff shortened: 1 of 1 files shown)


tests/integration/test_features.py
```diff
@@ -246,6 +246,16 @@ def test_can_round_trip_binary_custom_content_type(smoke_test_app):
     assert response.content == bin_data
 
 
+def _assert_contains_access_control_allow_methods(headers, methods):
+    actual_methods = headers['Access-Control-Allow-Methods'].split(',')
```
jamesls: Why not just `assert sorted(actual_methods) == sorted(methods)`?  You'd also get a better error message so you could see what the full list is instead of just which one failed.

kyleknap: Right. I updated code.

Issue #330, opened 2017-05-07
Error on chalice deploy: "The write operation timed out"


ehammond:
chalice (0.8.0?) had been working for me a week ago, but when I tried again with a fresh install (0.8.1) today, I'm now getting timeouts on deploy.

    $ virtualenv --python $(which python3.6) $HOME/.virtualenvs/chalice-python3.6
    $ source $HOME/.virtualenvs/chalice-python3.6/bin/activate
    $ pip install chalice
    $ python --version
    Python 3.6.1
    $ chalice --version
    chalice 0.8.1

    $ chalice new-project helloworld
    $ cd helloworld
    $ time chalice deploy
    Initial creation of lambda function.
    Updating IAM policy.
    Creating deployment package.
    ('Connection aborted.', timeout('The write operation timed out',))
    
    real	5m55.284s
    user	0m3.388s
    sys	0m0.384s

It spends over five minutes uploading something at a fairly solid 0.5Mbps. How much data needs to be uploaded for a bare deploy?

Is it possible the upload is failing for some reason and retrying over and over? Is there a way to get more verbose debug messages?

I tried this with two distinct Internet connection paths to help reduce that as a suspect.

I'm using Ubuntu 17.04 Zesty.

jamesls:
I tried a few different options, but I'm not seeing any timeouts:

```
$ mktmpenv --python $(which python3.6)
Running virtualenv with interpreter /usr/local/bin/python3.6
...
(tmp-db699489363cc6a7) $ pip install chalice
(tmp-db699489363cc6a7) $ python --version
Python 3.6.1
(tmp-db699489363cc6a7) $ chalice --version
chalice 0.8.1
(tmp-db699489363cc6a7) $ chalice new-project helloworld081
(tmp-db699489363cc6a7) $ cd helloworld081/
(tmp-db699489363cc6a7) $ time chalice deploy
Initial creation of lambda function.
Creating role
Creating deployment package.
Initiating first time deployment...
Deploying to: dev
https://9q3tdyldk7.execute-api.us-west-2.amazonaws.com/dev/

real	0m22.901s
user	0m2.596s
sys	0m0.641s
```

Could you share the output of `chalice --debug deploy`?  That should give us a better idea of what's going wrong

ehammond:
Thanks for investigating. Some hours later on a third Internet connection, it started working, and it only took a short time at around the same 0.5 Mbps. There was no change to the code, so it may have been the network or something on the AWS side that was causing the problem.

[...]

PR #380, opened 2017-06-14
Add better errors for too large deployments

kyleknap:
The better errors consistent of:
* Detecting errors that were likely because of chalice applications
  that were too large.
* Indicating that it occurred during the lambda deployment phase
* Indicating suggestions on how to fix it

In addition, added a unified ChaliceDeploymentError that can be
used to improve the error message when deploying a chalice
application for future enhancements.

Fixes https://github.com/awslabs/chalice/issues/246

Here are some examples for the error messages:
1) Connection getting aborted as chalice is sending data while Lambda cuts off the connection.
```
DEPLOYMENT ERROR - While sending your chalice handler code to Lambda, received the following error: ('Connection aborted.', BrokenPipeError(32, 'Broken pipe')). This is likely because the deployment package is 58.2 MB. Lambda only allows deployment packages that are 50.0 MB or less in size. To avoid this error, decrease the size of your chalice application by removing code or removing dependencies from your chalice application.
```
2) Lambda returns an error saying the zipped deployment package is too large.
```
DEPLOYMENT ERROR - While sending your chalice handler code to Lambda, received the following error: An error occurred (RequestEntityTooLargeException) when calling the CreateFunction operation: Request must be smaller than 69905067 bytes for the CreateFunction operation. To avoid this error, decrease the size of your chalice application by removing code or removing dependencies from your chalice application.
```
3) Lambda returns an error saying the contents of the unzipped deployment package is too large.
```
DEPLOYMENT ERROR - While sending your chalice handler code to Lambda, received the following error: An error occurred (InvalidParameterValueException) when calling the CreateFunction operation: Unzipped size must be smaller than 262144000 bytes. To avoid this error, decrease the size of your chalice application by removing code or removing dependencies from your chalice application.
```


jamesls:
Haven't had a chance to look at the code yet, but some questions/feedback on the output:

1. What do the error messages look like when it's not an exception caused by one of the cases you've explicitly accounted for?  For example, write timeouts as shown here: https://github.com/awslabs/chalice/issues/330.  Do we still say you should reduce your deployment package size?
2. I think it's more common to just have `ERROR - then your message`.  The deployment context is inferred given you run `chalice deploy`.
3. I'd prefer something other than "chalice handler", it's vague to me.  Maybe something like "chalice API handler".  That way builtin auth could say "chalice auth handler".
4. The line lengths are really long (you have to scroll horizontally to read the only thing).  I'd prefer if we added line breaks and added a nice multi line error message.
5. Can we clean up the first error message?  I realize this comes from requests, but it just seems odd that only the first error message has this tuple format: `received the following error: ('Connection aborted.', BrokenPipeError(32, 'Broken pipe')).`

I'll try to look at the code tomorrow.

kyleknap:
Just answering some questions:

**What do the error messages look like when it's not an exception caused by one of the cases you've explicitly accounted for? For example, write timeouts as shown here: #330. Do we still say you should reduce your deployment package size?**
If there is no context for the error, it will currently show up as the string of the raised exception for any deployment error:
```
DEPLOYMENT ERROR - <str(Exception)>
```
In terms of write timeouts, receiving that message is entirely determined on if the deployment package is larger than 50MB. If it is, you get an error message similar to the first option. If not you will get something like this:
```
DEPLOYMENT ERROR - ('Connection aborted.', timeout('The write operation timed out',))
```

**I think it's more common to just have ERROR - then your message. The deployment context is inferred given you run chalice deploy.**
That's fine with me.

**I'd prefer something other than "chalice handler", it's vague to me. Maybe something like "chalice API handler". That way builtin auth could say "chalice auth handler".**
That's fair. I was phrasing it based on the current feature state of chalice as there is no mult-handler support currently. I could get that to work currently, but I would be interested to see how it will all fit in when auth handlers get introduced. There will probably need to be reworking of where the errors get thrown as the deployment errors are raised from the ``Deployer.deploy`` method.

**The line lengths are really long (you have to scroll horizontally to read the only thing). I'd prefer if we added line breaks and added a nice multi line error message.**
I was thinking about that too. It is not too bad on the terminal as it wraps around. It just does not look great on GitHub. How would this format sound:
```
ERROR - 
   LOCATION: Deploying the chalice API handler
   EXCEPTION RAISED: <str(Exception)>
   SUGGESTION: To resolve this issue, do the following...
```
I could even have it print out in red or something to make it easier to understand all of the related information about the exception.

**Can we clean up the first error message? I realize this comes from requests, but it just seems odd that only the first error message has this tuple format: received the following error: ('Connection aborted.', BrokenPipeError(32, 'Broken pipe')).**
It may be tricky to see if there is any attributes on `requests.ConnectionError` that I can use. The ConnectionError subclasses from an IOError. Based on the [Python docs](https://docs.python.org/2/library/exceptions.html#exceptions.BaseException) I could maybe inspect the `args` property and try to pull information out of the tuple but I am not sure how well I will be able to reformat into a better looking looking sentence or if I would be making inappropriate assumptions in what the args will be in the tuple. The second option would be to provide your own error message for ConnectionErrors. I am hesitant to do that now as it is nice to see what the exact error we are getting back so it is possible to trace the source of the issue. If we had a `--debug` mode, I think it would be easier to make the case for providing your own message for a connection error, but until then I am not sure if we would want to do that.

kyleknap:
@stealthycoin 

To answer your questions:

**What do these errors look like in a short terminal?**
The text just wraps to make a taller paragraph.

**Can we clean up the request error with the tuple in it a little bit?**
See my last paragraph in the previous response. I feel that if we move to the format that I proposed in the answering 4th question, it might remove the need to clean it up as we would be preserving the exact exception that got raised.

jamesls:
@kyleknap I don't think preserving the original exception is useful from a user's perspective.  Sure we can keep it in debug logs once we add those, but seeing ` ('Connection aborted.', BrokenPipeError(32, 'Broken pipe')).` is confusing for end users.  We've had so many people (not just in chalice) asking about these broken pipe errors that I'm convinced it does more harm than good.

The multi-line format you suggest seems a bit much.  I was just asking you to put line breaks so it's not a super long single line error message.

Also, for the first question, it looks like there's a missing gap.  Saying there's a write timeout without actually saying *what* we were writing when we timed out is going to confuse users as see in the issue I linked.

kyleknap:
@jamesls One thing that I can do to break it up is add a newline between the exception and the suggestion on how to fix it. Would that help? There is really only three possible logical components of the error: the location of where the error occurred, the error message, and the suggestion on how to resolve. Currently the location and the error message are combined into one sentence.


On the:  We've had so many people (not just in chalice) asking about these broken pipe errors that I'm convinced it does more harm than good.

Does it make sense to add improved messaging to botocore then? If I was to do this in chalice, I would prefer we improve the error message for any client call by at least adding information on what operation threw the ConnectionError as a ConnectionError technically could happen for any client call and the error message will still be very bad. Also currently, it is only somewhat manageable/understandable for botocore/boto3/awscli because we have logging and tracebacks enabled.

kyleknap:
I'm still looking on how to improve the connection error message part, but how about something like this?
screen shot 2017-06-15 at 12 53 54 pm

jamesls:
Please avoid using colors for now.  I'm not against it but I don't think we should start with this change.  Let's just focus on improving the error messages themselves for now.

jamesls:
> Does it make sense to add improved messaging to botocore then?

No I don't think so.  Going into botocore would require we could generalize a broken pipe error and why it happened.  We can only do that here because we know the exact API call and common reasons why it can fail.

I strongly suggest to just keep the scope small.  This is just about improving error messages, and I'd rather move on to other features once we've done that.

chalice/deploy/deployer.py
```diff
@@ -229,6 +251,22 @@ def deploy(self, config, chalice_stage_name=DEFAULT_STAGE_NAME):
             chalice_stage_name: deployed_values
         }
 
+    def _get_chalice_deployment_error(self, error):
+        # type: (Any) -> ChaliceDeploymentError
+        suggestion = None
+        where = None
+        if isinstance(error, DeploymentPackageTooLargeError):
```
jamesls: Let's move this out into the `ChaliceDeploymentError` class.  We should want the deployer to only be about deploying resources and move the logic of helpful error messages out to a separate class.

chalice/awsclient.py
```diff
@@ -102,18 +122,60 @@ def _call_client_method_with_retries(self, method, kwargs):
...
+        error_message = str(error)
+        should_raise = False
+        additional_details = None
+        if (isinstance(error, RequestsConnectionError) and
+                deployment_size > self.MAX_LAMBDA_DEPLOYMENT_SIZE):
+            # When the zip deployment package is too large and Lambda
+            # aborts the connection as chalice is still sending it
+            # data
+            additional_details = (
+                'This is likely because the deployment package is %s. '
```
jamesls: Lets move this up the stack and have stuff in awsclient just collect all the raw data, deployment package size, original exceptions, etc.   We'd then have the Exception object assemble the suggestions and pretty print the error messages to the user.

chalice/awsclient.py
```diff
@@ -92,8 +108,12 @@ def create_function(self,
             kwargs['Timeout'] = timeout
         if memory_size is not None:
             kwargs['MemorySize'] = memory_size
-        return self._call_client_method_with_retries(
-            self._client('lambda').create_function, kwargs)['FunctionArn']
+        try:
+            return self._call_client_method_with_retries(
+                self._client('lambda').create_function, kwargs)['FunctionArn']
+        except Exception as e:
+            self._raise_large_deployment_related_errors(e, len(zip_contents))
```
jamesls: Consider creating some lambda specific error so pieces higher up the stack know that we failed in the lambda function.

chalice/awsclient.py
```diff
@@ -102,18 +122,60 @@ def _call_client_method_with_retries(self, method, kwargs):
...
                 # We're assuming that if we receive an
                 # InvalidParameterValueException, it's because
                 # the role we just created can't be used by
                 # Lambda so retry until it can be.
                 self._sleep(self.DELAY_TIME)
                 attempts += 1
-                if attempts >= self.LAMBDA_CREATE_ATTEMPTS:
+                message = e.response['Error'].get('Message', '')
+                retryable_message = (
+                    'The role defined for the function cannot be assumed '
```
kyleknap: I am planning to just check to see if the string 'role' is included in the error message if that is fine?

jamesls: I'd prefer something more robust than that.  If the function name can ever be in the error message and that name contains `role` you'll get a false positive.

kyleknap: So the literal error message, you will get for both create and update is:
```
The role defined for the function cannot be assumed by Lambda.
```
It does not include the name of the function currently. I can check for 'role' and 'assumed by Lambda'? Or just 'assumed by Lambda'?

jamesls: No, I mean in general for InvalidParameterValueException, which is the code block this is in. This a) applies to more than just the role not being able to be assumed and b) has no guarantees about the message changing in the future.  Something like `role.*cannot be assumed` seems reasonable to me.

kyleknap: Alright I can do that.

kyleknap:
@jamesls I updated. I think I got everything we talked about. Should be good to look at again.

jamesls:
@kyleknap Could you show what the updated sample error messages look like?

kyleknap:
@jamesls Here are some examples:

Too big when unzipped:
```
ERROR - While sending your chalice handler code to Lambda function "bigapp-
dev", received the following error:

 An error occurred (InvalidParameterValueException) when calling the 
 CreateFunction operation: Unzipped size must be smaller than 262144000 bytes

To avoid this error, decrease the size of your chalice application by removing 
code or removing dependencies from your chalice application.
```

Connection aborted:
```
ERROR - While sending your chalice handler code to Lambda function 
"bigappv2-dev", received the following error:

 Connection aborted. Lambda closed the connection before chalice finished 
 sending all of the data.

This is likely because the deployment package is 58.2 MB. Lambda only allows 
deployment packages that are 50.0 MB or less in size. To avoid this error, 
decrease the size of your chalice application by removing code or removing 
dependencies from your chalice application.
```

Timeout:
```
ERROR - While sending your chalice handler code to Lambda function 
"bigappv2-dev", received the following error:

 Connection aborted. Timed out sending your app to Lambda.

This is likely because the deployment package is 58.2 MB. Lambda only allows 
deployment packages that are 50.0 MB or less in size. To avoid this error, 
decrease the size of your chalice application by removing code or removing 
dependencies from your chalice application.
```

Zip too large (connection not aborted)
```
ERROR - While sending your chalice handler code to Lambda function 
"bigappv3-dev", received the following error:

 An error occurred (RequestEntityTooLargeException) when calling the 
 CreateFunction operation: Request must be smaller than 69905067 bytes for the 
 CreateFunction operation

This is likely because the deployment package is 50.9 MB. Lambda only allows 
deployment packages that are 50.0 MB or less in size. To avoid this error, 
decrease the size of your chalice application by removing code or removing 
dependencies from your chalice application.
```

jamesls:
Cool, looks better.  The first sentence is a little confusing, we'll not really sending our code to lambda function '{our function}'.  Something like "while sending your chalice handler code to Lambda to create the function "function-name", we received the following error.  Or something along those lines.

Looking over the code now.

chalice/awsclient.py
```diff
@@ -37,6 +42,28 @@ class ResourceDoesNotExistError(Exception):
     pass
 
 
+class LambdaClientError(Exception):
+    def __init__(self, original_error, context):
+        # type: (Any, LambdaErrorContext) -> None
```
jamesls: I try to avoid `Any` unless there's really not a better way.  Can this not be `Exception` or something more specific?

chalice/awsclient.py
```diff
@@ -92,8 +119,12 @@ def create_function(self,
             kwargs['Timeout'] = timeout
         if memory_size is not None:
             kwargs['MemorySize'] = memory_size
-        return self._call_client_method_with_retries(
-            self._client('lambda').create_function, kwargs)['FunctionArn']
+        try:
+            return self._call_client_method_with_retries(
+                self._client('lambda').create_function, kwargs)['FunctionArn']
+        except Exception as e:
```
jamesls: I still don't think this is right.  We're catching all exceptions and casting them down into a more specific exception (`LambdaClientError`).  If there is actually a bug or something exceptional that happens in the code I'd want the original stack trace to propagate up, not be masked by `LambdaClientError`.

chalice/awsclient.py
```diff
@@ -102,18 +133,45 @@ def _call_client_method_with_retries(self, method, kwargs):
...
+            except client.exceptions.InvalidParameterValueException as e:
                 # We're assuming that if we receive an
                 # InvalidParameterValueException, it's because
                 # the role we just created can't be used by
                 # Lambda so retry until it can be.
                 self._sleep(self.DELAY_TIME)
                 attempts += 1
-                if attempts >= self.LAMBDA_CREATE_ATTEMPTS:
+                message = e.response['Error'].get('Message', '')
+                # Do not retry if exceeded retries or does not have anything
```
jamesls: So a style nitpick, but I have said this several times before.  Rather than write a comment explaining *what* the code is doing, just put it in a separate function and omit the comment:

```
if self._is_iam_role_related_error(e):
    raise
```

chalice/awsclient.py
```diff
@@ -102,18 +133,45 @@ def _call_client_method_with_retries(self, method, kwargs):
...
+        # type: (Any, LambdaErrorContext) -> LambdaClientError
+        error_cls = LambdaClientError
+        if (isinstance(error, RequestsConnectionError) and
+                context.deployment_size > MAX_LAMBDA_DEPLOYMENT_SIZE):
+            # When the zip deployment package is too large and Lambda
+            # aborts the connection as chalice is still sending it
+            # data
+            error_cls = DeploymentPackageTooLargeError
+        elif isinstance(error, ClientError):
+            code = error.response['Error'].get('Code')
```
jamesls: These will default to `None` which will crash in your `__contains__` call on L169.  These should default to `str` types, probably just the empty string.

chalice/awsclient.py
```diff
@@ -140,8 +198,12 @@ def update_function(self,
         the targeted lambda function.
         """
         lambda_client = self._client('lambda')
-        return_value = lambda_client.update_function_code(
-            FunctionName=function_name, ZipFile=zip_contents)
+        try:
+            return_value = lambda_client.update_function_code(
+                FunctionName=function_name, ZipFile=zip_contents)
+        except Exception as e:
```
jamesls: Same thing, we should really only have `except Exceptions` and the top most layer of the code.  Restrict this to the subset of exceptions we care about.

chalice/deploy/deployer.py
```diff
@@ -169,6 +177,104 @@ def _validate_manage_iam_role(config):
             )
 
 
+class ChaliceDeploymentError(Exception):
+    def __init__(self, error):
+        # type: (Any) -> None
```
jamesls: Prefer to avoid `Any` unless it's really not practical.  This applies to all the `Any`'s in this class.

chalice/deploy/deployer.py
```diff
@@ -169,6 +177,104 @@ def _validate_manage_iam_role(config):
...
+        #         'Connection aborted.', <SomeException>)
+        # )
+        message = connection_error.args[0].args[0]
+        underlying_error = connection_error.args[0].args[1]
+
+        # In python3, this is a BrokenPipeError. However in python2, this
+        # is a socket.error that has the message 'Broken pipe' in it. So we
+        # don't want to be assuming all socket.error are broken pipes so just
+        # check if the message has 'Broken pipe' in it.
+        if 'Broken pipe' in str(underlying_error):
```
jamesls: This seems more appropriately handled in `compat.py`.  Here we actually have a solid `isinstance` way of checking for broken pipe in python3, and it seems to make sense to use it if possible rather than rely on `__contains__`.

chalice/deploy/deployer.py
```diff
@@ -206,6 +312,13 @@ def deploy(self, config, chalice_stage_name=DEFAULT_STAGE_NAME):
             rest api, lambda function, role, etc.
 
         """
+        try:
+            return self._do_deploy(config, chalice_stage_name)
+        except Exception as error:
```
jamesls: Same thing about `except Exception`.  This would be fine in say `cli/__init__.py`, but not in the deployer module.

chalice/deploy/deployer.py
```diff
@@ -169,6 +177,104 @@ def _validate_manage_iam_role(config):
...
+        # type: (Any) -> None
+        where = self._get_error_location(error)
+        msg = self._wrap_text(
+            'ERROR - %s, received the following error:' % where
+        )
+        msg += '\n\n'
+        msg += self._wrap_text(self._get_error_message(error), indent=' ')
+        msg += '\n\n'
+        suggestion = self._get_error_suggestion(error)
+        if suggestion:
```
jamesls: `if suggestion is not None`

tests/functional/test_awsclient.py
```diff
@@ -475,6 +491,75 @@ def test_can_provide_tags(self, stubbed_session):
             handler='app.app') == 'arn:12345:name'
         stubbed_session.verify_stubs()
 
+    def test_raises_large_deployment_error_for_connection_error(
+            self, mock_lambda_client):
+        mock_session = mock.Mock(botocore.session.Session)
```
jamesls: Why are we doing this?? There's an existing way to test this, which is using the StubbedSession.

If the necessary abstractions aren't in place in StubbedSession, I think it's more appropriate it to add them rather than work around them bypassing the StubbedSession and dropping down to botocore sessions directly.

chalice/awsclient.py
kyleknap: Offline: Decided to just catch botocore.exceptions.ClientError and RequestsConnectionError for this.

chalice/awsclient.py
kyleknap: Sure

chalice/awsclient.py
kyleknap: Offline: Catch LambdaClientError and botocore.exceptions.ClientError here instead.

chalice/awsclient.py
kyleknap: I will check but I believe it was mypy complaining on accessing attributes that are not on Exception such as context.

chalice/deploy/deployer.py
kyleknap: Sure

tests/functional/test_awsclient.py
kyleknap: Offline: Going to subclass this from the botocore.stub.Stubber and add todo to port back to botocore.

chalice/deploy/deployer.py
kyleknap: Offline: Catch LambdaClientError and botocore.exceptions.ClientError here instead.

chalice/deploy/deployer.py
kyleknap: It should not matter if the check changes here. The output will stay the same, but sure I can update.

kyleknap:
@jamesls I updated based on feedback. It should be good to look at again.

chalice/compat.py
```diff
@@ -1,13 +1,27 @@
...
+import socket
 import six
 
 
 if six.PY3:
     from urllib.parse import urlparse, parse_qs
+
+    def is_broken_pipe_error(error):
+        # type: (Exception) -> bool
+        return isinstance(error, BrokenPipeError)  # noqa
```
jamesls: Why is `# noqa` needed here?

kyleknap: Pylint on python2 yelled at me for having an unknown variable name of BrokenPipeError as it only exists in Python3. Is there a better way to handle this? It seemed that using `#noqa` is the precedent that has been set in previous places for pylint errors due to need for type hinting.

jamesls: Are you sure that's the case?  Pylint doesn't honor `# noqa` AFAIK.  That's for pyflakes/flake8.

kyleknap: No you are right. I misspoke. It was for flake8.

PR #381, opened 2017-06-15
Add support for builtin API Gateway authorizers

jamesls:
Chalice currently supports IAM, Cognito, and Custom authorizers.
The custom authorizer allows you to define a lambda function which
contains custom authorization logic specific for your use case.
Previously, you'd have to create your lambda authorizer function out
of band from your chalice app.  We just provided an API to associated
your custom authorizer with certain views.

This change adds support for a "built in authorizer", which allows you
to create custom authorizers that are defined within your chalice app.
When you run "chalice deploy" we'll automatically create and associate
the authorizer for you, but we'll also create the lambda function
associated with the authorizer.

This introduces new concepts in chalice including multi lambda function
support.  We now have an API handler (the existing lambda function) as
well as auth handlers.

There's still a few missing pieces that need to get ironed out:

* Need to work out how we want to configure the auth functions
  in config.json (timeouts, memory, etc)
* Need to work out how to deal with "chalice package".  It appears
  SAM doesn't support giving implicit permission to invoke the lambda
  function in the same way the event type "Api" works.
* Docs (in progress, should be coming shortly).

Closes https://github.com/awslabs/chalice/issues/356


tests/unit/test_app.py
```diff
@@ -773,3 +781,192 @@ def test_eq_non_default_configurations(self):
...
+
+
+def test_can_use_auth_routes_instead_of_strings(auth_request):
+    expected = [
+        "arn:aws:execute-api:us-west-2:123:rest-api-id/dev/GET/a",
+        "arn:aws:execute-api:us-west-2:123:rest-api-id/dev/GET/a/b",
+        "arn:aws:execute-api:us-west-2:123:rest-api-id/dev/POST/a/b",
+    ]
+    response = app.AuthResponse(
+        [app.AuthRoute('/a', ['GET']),
```
kyleknap: Do you think it is worth adding a test that mixes string and AuthRoute in the AuthResponse list?

chalice/awsclient.py
```diff
@@ -483,3 +483,29 @@ def _client(self, service_name):
             self._client_cache[service_name] = self._session.create_client(
                 service_name)
         return self._client_cache[service_name]
+
+    def add_permission_for_authorizer(self, rest_api_id, function_arn,
+                                      random_id):
```
jamesls: I think it would.  IIRC correctly it was done this way to make testing easier.  I believe there were issues with the botocore stubber's `ANY` implementation.  Those might be fixed now, I can check.

chalice/awsclient.py
```diff
@@ -483,3 +483,29 @@ def _client(self, service_name):
...
+        # type: (str, str, str) -> None
+        client = self._client('apigateway')
+        authorizers = client.get_authorizers(restApiId=rest_api_id)
```
jamesls: It does not.  I couldn't find in their docs what the page size, but I think this might be a forward compatibility thing in their API.  The max authorizers for an API is 10 (http://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html) so I don't think we'll be affected.  I can add a comment about that though.

chalice/app.py
```diff
@@ -461,6 +462,25 @@ def define_authorizer(self, name, header, auth_type, provider_arns=None):
...
         }
 
+    def authorizer(self, name=None, **kwargs):
+        def _register_authorizer(auth_func):
+            auth_name = name
+            if auth_name is None:
+                auth_name = auth_func.__name__
+            ttl_seconds = kwargs.pop('ttl_seconds', None)
+            execution_role = kwargs.pop('execution_role', None)
+            auth_config = BuiltinAuthConfig(
```
jamesls: I'll update.

chalice/app.py
```diff
@@ -626,3 +646,120 @@ def _add_cors_headers(self, response, cors):
         for name, value in cors.get_access_control_headers().items():
             if name not in response.headers:
                 response.headers[name] = value
+
+
+class BuiltinAuthConfig(object):
```
jamesls: There'a already a custom authorizer in the chalice code.  The nomenclature that made the most sense to be (given what's already there):

* Custom authorizers - using the custom authorizer functionality in API gateway with a lambda function that you manage outside of chalice.  This maps to the pre-existing `CustomAuthorizer` class.
* Builtin authorizers - authorizers that are defined in a chalice app that we manage/configure for you.  Implementation wise it uses the custom authorizer feature of API gateway but is largely abstracted in chalice.

chalice/deploy/deployer.py
```diff
@@ -280,7 +308,48 @@ def deploy(self, config, existing_resources, stage_name):
                 config, function_name, stage_name)
             deployed_values['api_handler_name'] = function_name
         deployed_values['api_handler_arn'] = function_arn
-        return deployed_values
+
+    def _deploy_auth_handlers(self, config, existing_resources, stage_name,
```
jamesls: It will be semi-orphaned in the sense that we won't delete it on deploy, but it will still stay in `deployed.json` so will get removed as part of `chalice delete`.

I agree it does make sense if you delete your auth handler code we should remove the lambda function.  I'll look into how much work that is.

jamesls: Capturing offline:  I'll use either the deployed.json or query lambda based on the expected tags to delete things at the end of the deploy.

chalice/awsclient.py
jamesls: On second thought, I think it would make sense to parameterize it but not completely hide it.  This is used to ensure uniqueness and makes it impossible to do retries at any level above the awsclient level.  I'd propose making this an `Optional[str]`.

chalice/awsclient.py
kyleknap: That works too.

chalice/deploy/deployer.py
```diff
@@ -280,7 +308,48 @@ def deploy(self, config, existing_resources, stage_name):
...
+            return
+        for auth_config in auth_handlers:
+            self._deploy_auth_handler(
+                config, auth_config, stage_name, deployed_values)
+
+    def _deploy_auth_handler(self, config, auth_config,
+                             stage_name, deployed_values):
+        # type: (Config, app.BuiltinAuthConfig, str, Dict[str, Any]) -> None
+        api_handler_name = deployed_values['api_handler_name']
+        role_arn = self._get_or_create_lambda_role_arn(
```
jamesls: Yeah.  That would be an (awesome) enhancement.

I'll add support for configuring the role via the `config.json` similar to what we have now, but the autopolicy stuff will just be for the whole app.py file for now.

chalice/deploy/deployer.py
```diff
@@ -280,7 +308,48 @@ def deploy(self, config, existing_resources, stage_name):
...
+        zip_contents = self._osutils.get_file_contents(
+            self._packager.deployment_package_filename(config.project_dir),
+            binary=True)
+        function_name = api_handler_name + '-' + auth_config.name
+        if self._aws_client.lambda_function_exists(function_name):
+            response = self._update_lambda_function(
+                config, function_name, stage_name)
+            function_arn = response['FunctionArn']
+        else:
+            function_arn = self._aws_client.create_function(
```
jamesls: Forgot to capture this offline discussion:

In general yes, but this might end up being part of a larger refactoring to the deploy code base.  This is already going to change when I add support for configuring the auth handler lambda functions.  I'll see if I can replace consolidate when I add configuration support, but otherwise I might want to hold off for the larger deployment refactoring.

tests/unit/deploy/test_deployer.py
```diff
@@ -749,18 +783,53 @@ def setup_deployer_dependencies(self, app_policy):
...
+            'myapp-dev-myauth': 'arn:auth-function',
+        }
+
+    def test_can_update_auth_handlers(self, sample_app_with_auth):
+        config = self.create_config_obj(sample_app_with_auth)
+        deployer = LambdaDeployer(
+            self.aws_client, self.packager, None, self.osutils,
+            self.app_policy)
+        self.aws_client.lambda_function_exists.return_value = True
+        self.aws_client.update_function.return_value = {
```
jamesls: Yeah I can update this.

tests/unit/test_app.py
jamesls: Yeah I'll update this.

jamesls:
Incorporated feedback so far.  Remaining items:

* Docs
* Configuration for authorizer functions (timeouts/roles/memory_size)
* Delete auth lambda functions when removed from code on redeploy

Let me know if I missed anything outside of those three items.

kyleknap:
Changes look good to me up to this point.

jamesls:
The last piece of functionality missing is how to configure the authorizers.  I'm going to update the original issue with an updated proposal before implementing.

jamesls:
@kyleknap Updated.  This should be all the code related changes.  Just needs docs.  I'm going to try it out a bit more and see if there's any edge cases I missed, but otherwise is ready to look at again.

chalice/config.py
```diff
@@ -79,13 +80,15 @@ class Config(object):
...
+                 function_name=DEFAULT_HANDLER_NAME,
                  user_provided_params=None,
                  config_from_disk=None,
                  default_params=None):
-        # type: (str, StrMap, StrMap, StrMap) -> None
+        # type: (str, str, StrMap, StrMap, StrMap) -> None
         #: Params that a user provided explicitly,
         #: typically via the command line.
         self.chalice_stage = chalice_stage
+        self.function_name = function_name
```
kyleknap: So I think this is fine for now as lambda functions are really the only thing we configure. But how do you see this will expand if we ever support per-resource configurations for other AWS resources besides Lambda? Would we add resource name to the config as well and the config would then be scoped to: stage + lambda function + any other specific managed resource configuration? I just want to make sure there won't be any issues expanding this out

tests/unit/deploy/test_deployer.py
```diff
@@ -854,6 +855,82 @@ def test_can_update_auth_handlers(self, sample_app_with_auth):
             zip_contents=b'package contents',
         )
 
+    def test_can_create_auth_with_different_config(self, sample_app_with_auth):
+        # We're notusing create_config_obj because we want to approximate
```
kyleknap: notusing -> not using

tests/unit/test_config.py
```diff
@@ -89,6 +93,109 @@ def test_can_chain_chalice_stage_values():
...
+        }
+    }
+    c = Config(chalice_stage='dev', config_from_disk=disk_config)
+    new_config = c.scope(chalice_stage=stage_name,
+                         function_name=function_name)
+    assert new_config.environment_variables == {'from': expected}
+
+
+
+def test_can_create_scope_new_stage_only():
```
kyleknap: Unimplemented test

tests/unit/deploy/test_deployer.py
```diff
@@ -854,6 +855,82 @@ def test_can_update_auth_handlers(self, sample_app_with_auth):
...
+                                  'project_dir': '.'}
+        )
+        deployer = LambdaDeployer(
+            self.aws_client, self.packager, None, self.osutils,
+            self.app_policy)
+        self.aws_client.lambda_function_exists.return_value = False
+        self.aws_client.create_function.side_effect = [
+            self.lambda_arn, 'arn:auth-function']
+        deployer.deploy(config, None, stage_name='dev')
+        self.aws_client.create_function.assert_called_with(
```
kyleknap: I wonder if it makes sense to check that the call to create_function to create the api handler used the expected configuration values as well?

tests/unit/test_config.py
```diff
@@ -89,6 +93,109 @@ def test_can_chain_chalice_stage_values():
...
+
+
+@pytest.mark.parametrize('stage_name,function_name,expected', [
+    ('dev', 'api_handler', 'dev-api-handler'),
+    ('dev', 'myauth', 'dev-myauth'),
+    ('beta', 'api_handler', 'beta-api-handler'),
+    ('beta', 'myauth', 'beta-myauth'),
+    ('prod', 'api_handler', 'prod-stage'),
+    ('prod', 'myauth', 'prod-stage'),
+    ('foostage', 'api_handler', 'global'),
```
kyleknap: So a stage or a function name that does not exist in the config does not raise an error? That seems like it could raise some unexpected results down the road if the wrong stage or function name is used.

tests/unit/test_config.py
```diff
@@ -89,6 +93,109 @@ def test_can_chain_chalice_stage_values():
...
+                        'lambda_timeout': 40,
+                    }
+                }
+            }
+        }
+    }
+    c = Config(chalice_stage='dev', config_from_disk=disk_config)
+    new_config = c.scope(chalice_stage='dev',
+                         function_name='myauth')
+    assert new_config.lambda_timeout == 40
```
kyleknap: So I know it may be tedious, but I wonder if it makes sense to have explicit test to make sure all the valid variables can be configured per lambda function? In the tests, `lambda_timeout`, `environment_variables`, and `tags` are used routinely, but variables such as `iam_role_arn`, `lambda_memory_size`, `autogen_policy`, and `iam_policy_file` are not. I feel that it would be good to have for a quick check. If not split this into individual tests, maybe have one large test or general helper assertion that makes sure that each of these values exhibit the per function behavior (i.e. override global/stage values and are separate from other functions)?

chalice/config.py
```diff
@@ -231,13 +242,35 @@ def manage_iam_role(self):
...
+        # type: () -> Dict[str, str]
+        tags = self._chain_merge('tags')
+        tags['aws-chalice'] = 'version=%s:stage=%s:app=%s' % (
+            current_chalice_version, self.chalice_stage, self.app_name)
+        return tags
+
+    def scope(self, chalice_stage, function_name):
+        # type: (str, str) -> Config
+        # Used to create a new config object that's scoped to a different
+        # stage and/or function.  This creates a completely separate copy.
```
jamesls: This API is actually wrong now.  Initially I was doing a `deepcopy` on the `self.__dict__`, but ran into issues that couldn't be serialized because our config obj has a reference to the `chalice_app` which contains all contains of objects.  Now that we're sharing `__dict__`, this doesn't actually work.  I'll get this fixed.

chalice/config.py
jamesls: I would actually see additional resources not using `Config` directly but instead using a resource specific config obj that `Config` can create.  Initially I had this as `Config --references-> LambdaConfig`, but reverted that in favor of this current approach because:

1) It's a bigger interface change across all the deployer code which is likely getting refactored soon anyways
2) Almost all the config in `Config` is actually lambda configuration.

So say if we had ddb tables, you'd say something like:

```
ddb_config = cfg.resource_config(cfg.stage_name, 'dynamodb_tables')
ddb_config['mytable']['name']
ddb_config['mytable']['provisioned_throughput']
... etc ...
```

tests/unit/deploy/test_deployer.py
jamesls: I'm not following.  That's what's happening on lines 900/901 for the timeout and memory size.  Was there more you wanted asserted?

tests/unit/test_config.py
jamesls: We have no way of knowing that unless we want to change the interface to declare/define your stages before you can deploy them, which IMO is a tedious API.

Perhaps, we could add an optional prompt when you deploy a stage for the first time, but that wouldn't need changes to the config file validation.

tests/unit/deploy/test_deployer.py
kyleknap: So right now we are only checking the last client `create_function` call which was for the auth handler. The first `create_function` call is for the api handler which is not checked, but has different parameters such as its lambda timeout will be 10 and its lambda memory size will be the default. The assertion would be on what was called in the first `create_function` call and it would be just a sanity check to make sure the correct scope was used in creating the api handler. If there is already a check for that though, I don't think we need to add an assertion. I may have missed it if there is one as I am only looking at the diffs as you add them, not the entire PR diff each time.

chalice/config.py
kyleknap: Ok cool. I was thinking along the same lines of how you would get resource specific configuration. Makes sense to keep the changes small for now until more configurable managed resources get added as the deployer will need to get refactored.

tests/unit/deploy/test_deployer.py
jamesls: Oh ok yeah I can update the test if that's not covered.

tests/unit/test_config.py
kyleknap: Makes sense. Yeah I think it would be an optional prompt for a deploying to a new stage that is not the very first `chalice deploy`. I feel that the initial `chalice deploy` should work with minimal amount of prompting. I don't think this should be done in this PR though.

kyleknap:
Latest updates look good. Looks like all that is left is rebasing and adding docs then it should be good.

docs/source/topics/authorizers.rst
```diff
@@ -0,0 +1,219 @@
...
+    from chalice import Chalice, AuthResponse
+
+    app = Chalice(app_name='demoauth1')
+
+
+    @app.authorizer()
+    def demo_auth(auth_request):
+        token = auth_request.token
+        # This is just for demo purposes as shown in the API Gateway docs.
+        # Normally you'd call an oauth provider, validae the
```
kyleknap: validae -> validate

jamesls:
Merged via 4d1751d0a40b53aab29a778bcae4fc78344d9a04

PR #382, opened 2017-06-16
Draft of wheel packaging.

stealthycoin:
Pushing this up a little early so that folks can try it out. Testing is not fleshed out, nor are all the error cases/prompts during deployment.

This branch won't pass automated linting/testing yet.

cc @kyleknap @jamesls @dstufft

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -7,13 +7,15 @@ import shutil
 import zipfile
 import tarfile
 import contextlib
-import pip
 import subprocess
 from email.parser import FeedParser
+from email.message import Message  # noqa
 
-from typing import Any, List, Optional, Tuple, Iterable, Callable  # noqa
+import pip
+from typing import Any, Set, List, Optional, Tuple, Iterable, Callable  # noqa
 from chalice.compat import StringIO, lambda_abi
 from chalice.utils import OSUtils
+from chalice.constants import MISSING_DEPENDENCIES_TEMPLATE
 
 import chalice
 from chalice import app
@@ -27,32 +29,33 @@ class LambdaDeploymentPackager(object):
     _CHALICE_LIB_DIR = 'chalicelib'
     _VENDOR_DIR = 'vendor'
 
-    def __init__(self, osutils=None):
-        # type: (Optional[OSUtils]) -> None
+    def __init__(self, osutils=None, dependency_builder=None):
+        # type: (Optional[OSUtils], Optional[DependencyBuilder]) -> None
         if osutils is None:
             osutils = OSUtils()
+        if dependency_builder is None:
+            dependency_builder = DependencyBuilder(osutils)
         self._osutils = osutils
+        self._dependency_builder = dependency_builder
 
     def _get_requirements_file(self, project_dir):
         # type: (str) -> str
         # Gets the path to a requirements.txt file out of a project dir path
         return self._osutils.joinpath(project_dir, 'requirements.txt')
 
-    def create_deployment_package(self, project_dir, package_filename=None,
-                                  dependency_builder=None):
+    def create_deployment_package(self, project_dir, package_filename=None):
         # type: (str, Optional[str]) -> str
         print("Creating deployment package.")
         # Now we need to create a zip file and add in the site-packages
         # dir first, followed by the app_dir contents next.
-        if dependency_builder is None:
-            dependency_builder = DependencyBuilder(OSUtils())
         deployment_package_filename = self.deployment_package_filename(
             project_dir)
         if package_filename is None:
             package_filename = deployment_package_filename
         if not self._osutils.file_exists(package_filename):
-            dependency_builder.build_site_packages(project_dir)
-        site_packages_dir = dependency_builder.site_package_dir(project_dir)
+            self._dependency_builder.build_site_packages(project_dir)
+        site_packages_dir = self._dependency_builder.site_package_dir(
+            project_dir)
         dirname = self._osutils.dirname(
             self._osutils.abspath(package_filename))
         if not self._osutils.directory_exists(dirname):
@@ -63,11 +66,6 @@ class LambdaDeploymentPackager(object):
             self._add_app_files(z, project_dir)
             self._add_vendor_files(z, self._osutils.joinpath(project_dir,
                                                              self._VENDOR_DIR))
-            # TODO here we need to check that all the dependencies caluclated
-            # during the add_app_files phase are included in the package or
-            # warn the user. They can add pre-built amazon linux dependencies
-            # in the vendor folder and rerun the deployment and it should
-            # succeed. That warning/error should be issued here.
         return package_filename
 
     def _add_vendor_files(self, zipped, dirname):
@@ -179,7 +177,7 @@ class LambdaDeploymentPackager(object):
         print("Regen deployment package...")
         tmpzip = deployment_package_filename + '.tmp.zip'
         with zipfile.ZipFile(deployment_package_filename, 'r') as inzip:
-            with zipfile.ZipFile(tmpzip, 'w') as outzip:
+            with zipfile.ZipFile(tmpzip, 'w', zipfile.ZIP_DEFLATED) as outzip:
                 for el in inzip.infolist():
                     if self._needs_latest_version(el.filename):
                         continue
@@ -224,6 +222,7 @@ class DependencyBuilder(object):
     _MANYLINUX_COMPATIBLE_PLAT = {'any', 'linux_x86_64', 'manylinux1_x86_64'}
 
     def __init__(self, osutils, pip_runner=None):
+        # type: (OSUtils, Optional[PipRunner]) -> None
         self._osutils = osutils
         if pip_runner is None:
             pip_runner = PipRunner(pip)
@@ -266,18 +265,17 @@ class DependencyBuilder(object):
                     return True
         return False
 
-    def _download_sdists(self, requirements_file, directory):
+    def _download_all_dependencies(self, requirements_file, directory):
         # type: (str, str) -> set[Package]
-        # Download raw source dependences to get the transitive closure over
-        # the dependency graph. Once all have been download (as .tar.gz, .zip
-        # files) the wheels can be downloaded for the target platform
-        # manylinux1_x86_64 and using the python version from the current
-        # python environment, as it is assumed that is the environment that is
-        # being deployed to lambda.
+        # Download dependencies prefering wheel files but falling back to
+        # raw source dependences to get the transitive closure over
+        # the dependency graph. Return the set of all package objects
+        # which will serve as the master list of dependencies needed to deploy
+        # successfully.
         self._pip.download_all_dependencies(requirements_file, directory)
-        sdists = {Package(directory, filename) for filename
-                  in self._osutils.get_directory_contents(directory)}
-        return sdists
+        deps = {Package(directory, filename) for filename
+                in self._osutils.get_directory_contents(directory)}
+        return deps
 
     def _download_binary_wheels(self, sdists, directory):
         # type: (set[Package], str) -> set[Package]
@@ -289,44 +287,47 @@ class DependencyBuilder(object):
                 if filename.endswith('.whl')}
         return whls
 
-    def _build_missing_wheels(self, sdists, whls, directory):
+    def _build_missing_wheels(self, all_deps, whls, directory):
         # type: (set[Package], set[Package], str) -> None
         # Now that `directory` has all the manylinux1 compatible binary
         # wheels we could get, and all the source distributions. We need to
         # find all the source dists that do not have a matching wheel file,
         # and try to build it to a wheel file.
-        missing_whls = sdists - whls
+        missing_whls = all_deps - whls
         for whl in missing_whls:
-            path_to_sdist = self._osutils.joinpath(directory, whl.filename)
-            self._pip.build_wheel(path_to_sdist, directory)
+            if whl.dist_type == 'sdist':
+                path_to_sdist = self._osutils.joinpath(directory, whl.filename)
+                self._pip.build_wheel(path_to_sdist, directory)
 
     def _categorize_whl_files(self, directory):
-        # type: (str) -> Tuple[List[Package, List[Package]
-        # Final pass through the directory to ensure that all whl files
-        # are present and compatible with the lambda environment.
+        # type: (str) -> Tuple[Set[Package], Set[Package]]
         final_whls = [Package(directory, filename) for filename
                       in self._osutils.get_directory_contents(directory)
                       if filename.endswith('.whl')]
-        valid_whls, invalid_whls = [], []
+        valid_whls, invalid_whls = set(), set()
         for whl in final_whls:
             if self._valid_lambda_whl(whl.filename):
-                valid_whls.append(whl)
+                valid_whls.add(whl)
             else:
-                invalid_whls.append(whl)
+                invalid_whls.add(whl)
         return valid_whls, invalid_whls
 
     def _download_dependencies(self, directory, requirements_file):
-        # type: (str, str) -> List[Package]
-        sdists = self._download_sdists(requirements_file, directory)
-        whls = self._download_binary_wheels(sdists, directory)
-        self._build_missing_wheels(sdists, whls, directory)
-        # TODO make sure that the invalid whls are in package after
-        # vendored stage.
+        # type: (str, str) -> Tuple[Set[Package], Set[Package]]
+        deps = self._download_all_dependencies(requirements_file, directory)
+        valid_whls, invalid_whls = self._categorize_whl_files(directory)
+        sdists = deps - valid_whls - invalid_whls
+        missing_whls = sdists | invalid_whls
+        valid_whls = self._download_binary_wheels(missing_whls, directory)
+        self._build_missing_wheels(deps, valid_whls, directory)
+        # Final pass to find the valid whl files and see if there are any
+        # unmet dependencies left over.
         valid_whls, _ = self._categorize_whl_files(directory)
-        return valid_whls
+        missing_whls = deps - valid_whls
+        return valid_whls, missing_whls
 
     def _install_whls(self, src_dir, dst_dir, whls):
-        # type: (str, str, List[Package]) -> None
+        # type: (str, str, Set[Package]) -> None
         if os.path.isdir(dst_dir):
             shutil.rmtree(dst_dir)
         os.makedirs(dst_dir)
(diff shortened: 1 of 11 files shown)


chalice/deploy/packager.py
```diff
@@ -1,60 +1,72 @@
...
-        if self._has_at_least_one_package(requirements_file) and not \
-                os.path.isfile(package_filename):
-            p = subprocess.Popen([pip_exe, 'install', '-r', requirements_file],
-                                 stdout=subprocess.PIPE)
-            p.communicate()
-        deps_dir = compat.site_packages_dir_in_venv(venv_dir)
-        assert os.path.isdir(deps_dir)
-        # Now we need to create a zip file and add in the site-packages
-        # dir first, followed by the app_dir contents next.
+        if not os.path.isfile(package_filename):
```
kyleknap: Probably would want this built into osutils as a `file_exists()` to be consistent on using the OSUtils abstraction

chalice/deploy/packager.py
```diff
@@ -1,60 +1,72 @@
...
-        # more cases.
-        venv_dir = os.path.join(project_dir, '.chalice', 'venv')
-        self._create_virtualenv(venv_dir)
-        pip_exe = compat.pip_script_in_venv(venv_dir)
-        assert os.path.isfile(pip_exe)
-        # Next install any requirements specified by the app.
-        requirements_file = os.path.join(project_dir, 'requirements.txt')
+        # Now we need to create a zip file and add in the site-packages
+        # dir first, followed by the app_dir contents next.
+        if dependency_builder is None:
```
kyleknap: I think it would make sense to put this in the `__init__` that way the if they do not specify a `DependencyBuilder` the `OSUtils` will at least get plumbed into that if they specified that as well. Or if not we want to be making sure you specify self._osutils when instantiating this.

chalice/deploy/packager.py
```diff
@@ -1,60 +1,72 @@
...
+        # type: (Optional[OSUtils]) -> None
+        if osutils is None:
+            osutils = OSUtils()
+        self._osutils = osutils
 
-    def create_deployment_package(self, project_dir, package_filename=None):
+    def _get_requirements_file(self, project_dir):
+        # type: (str) -> str
+        # Gets the path to a requirements.txt file out of a project dir path
+        return os.path.join(project_dir, 'requirements.txt')
```
kyleknap: We probably want to include this functionality in the OSUtils. In general any call to the `os` library should probably live in the `OSUtils` class.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        for whl in whls:
+            zipfile_path = os.path.join(src_dir, whl.filename)
+            with zipfile.ZipFile(zipfile_path, 'r') as z:
+                z.extractall(dst_dir)
+
+    def build_site_packages(self, project_dir):
+        # type: (str) -> str
+        requirements_file = os.path.join(project_dir, 'requirements.txt')
+        deps_dir = self.site_package_dir(project_dir)
+        print(deps_dir)
```
kyleknap: Looks like a left over print statement

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+            # actually being specified, but those aren't common
+            # cases.
+            for line in f:
+                line = line.strip()
+                if line and not line.startswith('#'):
+                    return True
+        return False
+
+    @contextlib.contextmanager
+    def _tempdir(self):
```
kyleknap: Might make sense to put this in OSUtils as well

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+
+        As an implementation detail this class assumes that pip was imported
+        and uses pip.main. This gives us the correct version of python that
+        the user is intending to deploy for free, but relies on the pip main
+        function not changing, which is very unlikely.
+        """
+        final_command = [command]
+        final_command.extend(args)
+        # Call pip and hide stdout/stderr
+        with self._consume_stdout_and_stderr():
```
kyleknap: Out of curiosity, what does the output look like when you don't capture it? If it does not look good, it is worth not including it but we may want to at least log it in case anything important happened for debugging reasons.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+
+        # Try to get binary whls for each package downloaded above.
+        sdists = {Package(filename) for filename
+                  in self._osutils.get_directory_contents(directory)}
+        self._pip.download_manylinux_whls(
+            [pkg.identifier for pkg in sdists], directory)
+
+        # Now that `directory` has all the manylinux1 compatible binary
+        # wheels we could get, and all the source distributions. We need to
+        # find all the source dists that do not have a mathcing wheel file,
```
kyleknap: mathcing -> matching

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+    @contextlib.contextmanager
+    def _tempdir(self):
+        # type: () -> Any
+        tempdir = tempfile.mkdtemp()
+        try:
+            yield tempdir
+        finally:
+            shutil.rmtree(tempdir)
+
+    def _download_dependencies(self, directory, requirements_file):
```
kyleknap: It might be worth breaking each of these individual steps into separate internal methods so the method is not super long and it makes it a bit easier to read.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        return '%s==%s' % (name, version)
+
+    def _parse_sdist_filename(self, filename):
+        # type: (str) -> Tuple[str, str]
+        # Filenames are in the format {name}-{version}.tar.gz OR
+        # {name}-{version}.zip. These are the only two formats supported by
+        # PyPi for source distributions.
+        for ext in self.PYPI_SDIST_EXTS:
+            if filename.endswith(ext):
+                name, version = filename[:-len(ext)].split('-')
```
dstufft: This is going to make you sad, but sdist filenames cannot be deterministically parsed to a single name/version without prior knowledge of what name you're looking for. For example: ``foo-1-1.tar.gz`` could be either ``("foo", "1-1")`` or ``("foo-1", "1")``.

You can unpack the tarball and ask it though, doing somehting as simple as ``python setup.py --name`` and ``python setup.py --version`` should print the name/version to stdout... although that can fail if someone writes their ``setup.py`` so that it also prints to stdout. You can do it more robustly, but uh, it's really gross to do that. You get to monkeypatch the ``setup.py`` like pip does and then run ``python setup.py egg_info`` and parse the egg-info file. I can point you to code in pip that does that.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+    def build_wheel(self, wheel, directory):
+        # type: (str, str) -> int
+        """This command builds an sdist into a wheel file."""
+        arguments = ['--no-deps', '--wheel-dir', directory, wheel]
+        return self._execute('wheel', arguments)
+
+    def download_all_dependencies(self, requirements_file, directory):
+        # type: (str, str) -> int
+        """This command downloads all dependencies as sdists."""
+        # None of these should fail so we can request non-binary versions of
```
dstufft: I told you that none of these should fail so you should download the sdists, but it just occurs to me that this can still fail, because people can publish wheels to PyPI without publishing a sdist at all. This is pretty rare though, so it's possible we could just say that we don't support that edge case?

chalice/deploy/packager.py
```diff
@@ -208,3 +205,258 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        # all the packages in one call to `pip download`
+        arguments = ['--no-binary=:all:', '-r', requirements_file, '--dest',
```
dstufft: Right now people can influence where pip pulls its dependencies from using environment variables and config files. Do we want to allow that still or should chalice control where pip pulls its deps from and isolate itself from the host system? Either way, I think it would be nice to let people pass in their own repositories for people who don't want their deployment to depend on PyPI.

chalice/deploy/packager.py
stealthycoin: Ok I can get rid of the `--no-binary` part and adjust the rest of the steps minorly to account for that. Should cover all the cases.

chalice/deploy/packager.py
stealthycoin: Hmmm interesting. I guess technically pip is an implementation detail of the requirements builder. Though I think that would be useful to allow people who are getting requirements that are not on pypi.

chalice/deploy/packager.py
stealthycoin: Its pretty chatty, progress bars and warnings as pip installs things. I was thinking of capturing it and we can elevate it to the user if they have `--debug` enabled.

chalice/deploy/packager.py
dstufft: If we want to keep it an implementation detail, I suggest we run pip in isolated mode, and surface our own settings for passing in a repository or find-links style location and pass that into pip ourselves. That way we still control the interface and if we rip out pip then we can just translate those options to whatever the new thing is.

chalice/deploy/packager.py
dstufft: You might try adding varying levels of ``-q`` to the pip call to see if that makes it better enough that it changes your mind on how you want to handle this. Though I'm fine with not exposing that by default too, just mentioning that.

chalice/deploy/packager.py
stealthycoin: updated.

chalice/deploy/packager.py
stealthycoin: Good point, I hadn't considered that.

jamesls:
I haven't had a chance to look at the code yet, but just playing around with the PR, it seems to work well for most of the packages I tried.

I noticed that when I try out `scipy`, the download of the sdist takes *forever*.  It looks like it's actually compiling the package as part of the download.  As someone not familiar with the internals that's a bit surprising to me, as I would expect (regardless of pip's behavior) chalice to not compile something especially when there appears to be a lambda compatible wheel file (https://pypi.python.org/pypi/scipy).

Thoughts?

dstufft:
@jamesls I think that is probably going to be solved for the case of Scipy when @stealthycoin removes the ``--no-binary`` flag from the initial download. More generically though, there's not much that can be done about it if a project is compiling something when running ``python setup.py egg_info`` (even though they should *not* be doing that), but in this specific case the wheels will solve it.

stealthycoin:
Yep was just about to comment the same thing. I think this will be solved when I update the download logic to not force sdists initially.

stealthycoin:
By the way scipy if you download it straight from PyPi is `65mb` (after zipped) so it will fail to upload anyway. It will work if you vendor it yourself after manually stripping the binaries though since that gets it down to around `40mb` after zipped. Then of course you don't have debug symbols.

dstufft:
That reminds me of https://github.com/pypa/pip/issues/3938, for pip I decided not to do anything about it because space is *generally* not a concern for us and stripping binaries isn't universally safe. However for Lambda space is a concern, so maybe it'd make sense to offer binary stripping as part of the deployment process?

stealthycoin:
Downside of that is that we would need to introduce options to strip particular dependencies, and within those its possible that there are multiple binaries that can be safe/unsafe to strip which seems like a granularity that is better suited for just letting them handle it themselves and put what they want in the vendor directory. We would also either have to take a dependency on a pure python ELF parser or do best effort based on what tools are available on the operating system. Not sure it's worth it.

dstufft:
Yea, another option may be just the ability to run a user provided script as part of the deploy process that allows them to script that themselves-- although that's not super relevant to this particular PR.

tests/unit/deploy/test_packager.py
```diff
@@ -0,0 +1,253 @@
...
+        else:
+            with tarfile.open(path, 'w:gz') as tar:
+                tarinfo = tarfile.TarInfo('sdist/setup.py')
+                tarinfo.size = len(setup_py)
+                tar.addfile(tarinfo, io.BytesIO(setup_py.encode()))
+        return directory, filename
+
+    def test_setup_tar_gz(self, osutils, sdist_reader):
+        setup_py = self._SETUP_PY % (
+            self._SETUPTOOLS, 'foo-bar', '1.0-2b'
```
kyleknap: And we are using `'1.0-2b'` and  `'foo-bar'` because those won't work with the naive approach? It might be worth making it clear by having an explicit test name for it. And also have a test for a normal package name and version.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,332 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+
+    def _download_all_dependencies(self, requirements_file, directory):
+        # type: (str, str) -> set[Package]
+        # Download dependencies prefering wheel files but falling back to
+        # raw source dependences to get the transitive closure over
+        # the dependency graph. Return the set of all package objects
+        # which will serve as the master list of dependencies needed to deploy
+        # successfully.
+        self._pip.download_all_dependencies(requirements_file, directory)
+        deps = {Package(directory, filename) for filename
```
dstufft: I'm pretty sure that pip should only ever be downloading a single package per project here. Unless I'm missing something from how we're calling pip here, if it did so that would be a bug.

chalice/deploy/packager.py
```diff
@@ -208,3 +205,332 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        else:
+            info_fetcher = SdistMetadataFetcher()
+            name, version = info_fetcher.get_package_name_and_version(
+                self._directory, self.filename)
+        return '%s==%s' % (name, version)
+
+
+class SdistMetadataFetcher(object):
+    """This is the "correct" way to get name and version from an sdist."""
+    # https://github.com/pypa/pip/blob/master/pip/utils/setuptools_build.py
```
dstufft: It may be a reasonable idea to embed a commit hash in this URL since this file can be renamed, deleted, etc in pip's ``master`` branch which would invalidate this link.

chalice/deploy/packager.py
kyleknap: Ok cool. I made the assumption that the set was being used for duplicates, but after talking offline with @stealthycoin it was just to be able to take advantage of set operations.

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
...
+    def test_can_get_whls_all_manylinux(self, tmpdir, osutils, pip_runner):
+        reqs = ['foo', 'bar']
+        pip, runner = pip_runner
+        pip.add_side_effect('download', [
+            PipSideEffect('foo-1.2-cp36-cp36m-manylinux1_x86_64.whl'),
+            PipSideEffect('bar-1.2-cp36-cp36m-manylinux1_x86_64.whl')
+        ])
+
+        appdir = str(_create_app_structure(tmpdir))
+        self._write_requirements_txt('\n'.join(reqs), appdir)
```
kyleknap: So I wonder if we can make a higher level helper function that creates the app structure but does so by providing the tempdir and a list of requirements (instead of joining them with a newline and passing a string)? It will make it easier to read.

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
...
+        builder = DependencyBuilder(osutils, runner)
+        site_packages = builder.build_site_packages(appdir)
+        installed_packages = os.listdir(site_packages)
+        for req in reqs:
+            assert req in installed_packages
+
+    def test_can_get_whls_mixed_compat(self, tmpdir, osutils, pip_runner):
+        reqs = ['foo', 'bar', 'baz']
+        pip, runner = pip_runner
+        pip.add_side_effect('download', [
```
kyleknap: So it looks like download is really the only command used for side effects. Would it be possible to simplify to something like this:
```python
pip.packages_to_download([
    'foo-1.0-cp36-none-any.whl',
    'bar-1.2-cp36-cp36m-manylinux1_x86_64.whl',
    'baz-1.5-cp36-cp36m-linux_x86_64.whl'
])
```
Ideally I would like to not have to worry about the `PipSideEffect` object

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
...
+        ])
+
+        appdir = str(_create_app_structure(tmpdir))
+        self._write_requirements_txt('\n'.join(reqs), appdir)
+        builder = DependencyBuilder(osutils, runner)
+        with consume_stdout_and_stderr() as (out, _):
+            site_packages = builder.build_site_packages(appdir)
+        installed_packages = os.listdir(site_packages)
+        assert len(installed_packages) == 0
+        assert 'Could not install dependencies:\nbaz==1.5' in out.getvalue()
```
kyleknap: I would think this should be in standard error if it is stating it cannot download a dependency

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
...
+
+        appdir = str(_create_app_structure(tmpdir))
+        self._write_requirements_txt('\n'.join(reqs), appdir)
+
+        builder = DependencyBuilder(osutils, runner)
+        with consume_stdout_and_stderr() as (out, _):
+            site_packages = builder.build_site_packages(appdir)
+        installed_packages = os.listdir(site_packages)
+        assert len(installed_packages) == 0
+        assert 'Could not install dependencies:\nbaz==1.5' in out.getvalue()
```
kyleknap: Thinking about this again. I wonder if there is a better way to represent the packages that failed to install than relying on standard error? I wonder if we should be raising an error or returning the list of failed packages (along with successfully installed packages)?

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
...
+    def test_can_build_sdist(self, tmpdir, osutils, pip_runner):
+        reqs = ['foo', 'bar']
+        pip, runner = pip_runner
+        pip.add_side_effect('download', [
+            PipSideEffect('foo-1.2.zip'),
+            PipSideEffect('bar-1.2-cp36-cp36m-manylinux1_x86_64.whl')
+        ])
+        # Foo is built from and is pure python so it yields a compatible
+        # wheel file.
+        pip.add_side_effect('wheel', [
```
kyleknap: Yeah then for wheels we would have a similar:
```python
pip.packages_to_build_wheels_for(['foo-1.2-cp36-none-any.whl'])
```

setup.py
```diff
@@ -9,9 +9,9 @@
 install_requires = [
     'click==6.6',
     'botocore>=1.5.40,<2.0.0',
-    'virtualenv>=15.0.0,<16.0.0',
     'typing==3.5.3.0',
     'six>=1.10.0,<2.0.0',
+    'pip>=9'
```
dstufft: The ``--platform`` arguments to ``pip download`` he's using are new in pip 9, the original version specifier allowed versions that wouldn't work.

tests/functional/test_package.py
```diff
@@ -23,6 +31,257 @@ def index():
     return app
 
 
+@contextlib.contextmanager
```
stealthycoin: Ah cool I hadn't seen that.

setup.py
kyleknap: Ahh makes sense

tests/functional/test_package.py
stealthycoin: Wheel also uses it so I do need a way to provide that argument.

tests/functional/test_package.py
stealthycoin: In python 2 create_app_structure returns a LocalPath object.

stealthycoin:
Capturing offline discussion. The `PipSideEffect` class will not be invoked directly instead a function off of `FakePip` such as  `packages_to_build_wheels_for` will be called to registere a set of files to be created based on the pip function invoked. There will also be a way to check that the input parameters match some expected set before writing the files.

Issue #227, opened 2017-01-11
Support for egg-links


jleclanche:
When installing a dependency with git that is marked as zip-safe, `egg-link` files are created instead of a folder.

Chalice does not handle this. 

https://github.com/awslabs/chalice/blob/79560c81764f2a341daa729b3511513bd6a944cd/chalice/deployer.py#L397-L426

Zappa supports it:

https://github.com/Miserlou/Zappa/blob/f34beb51c2d7d0932769ccd54498010f47a36041/zappa/zappa.py#L394-L400

jamesls:
#388 added support for this.  We've rewritten how we handle dependencies and egg links are automatically supported now.

PR #388, opened 2017-06-26
Added packaging support for whl files.

stealthycoin:
The algorithm for building a lambda compatible bundle is as follows:

1) Download all packages using pip download without specifying binary
preference. This allows it to download whl files where possible and
sdists otherwise.
2) Try to download lamba-compatible whl files for all sdists
and lambda-incompatible whl files that were downloaded in the first
step. Once this step is completed we have all the whl files are are
going to be able to get from PyPi that are compatible with lambda.
3) Any sdists left over without a corresponding compatible whl file will
now be built to a whl. If this results in an incompatible whl file there
is nothing more we can do to get a compatible one and it is up to the
user to vendor that package themselves.

Now that we have done our best to get a whl file for each dependency, we
do a final pass over the directory and install all compatible whl files
to the lambda package. Any left over dependencies that do not have a
compatible whl file are reported to the user, as they will need to be
manually added to the vendor directory before deployment.

Virtualenv is no longer used so the compat functionality for it is no
longer needed.

Since the packaging process now downloads whl files which vary by python
version the deployment process now rebuilds the deployment package if it
detects that the python version has changed.

Old PR which stopped sycing with my fork branch for some reason: https://github.com/awslabs/chalice/pull/382

closes: #249 #227 #364 

### DIFF
--- a/chalice/constants.py
+++ b/chalice/constants.py
@@ -197,3 +197,15 @@ If you'd like to proceded, then answer the questions
 below.
 
 Please enter the project name"""
+
+
+MISSING_DEPENDENCIES_TEMPLATE = r"""
+Could not install dependencies:
+%s
+You will have to build these yourself and vendor them in
+the chalice vendor folder.
+
+Your deployment will continue but may not work correctly
+if missing dependencies are not present. For more information:
+http://chalice.readthedocs.io/en/latest/topics/packaging.html
+"""
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -689,6 +689,18 @@ class LambdaDeployer(object):
             return DEFAULT_LAMBDA_MEMORY_SIZE
         return config.lambda_memory_size
 
+    def _can_reuse_deployment_package(self, config, lambda_name,
+                                      deployment_package_filename):
+        # type: (Config, str, str) -> bool
+        if not self._osutils.file_exists(deployment_package_filename):
+            return False
+        lambda_config = self._aws_client.get_function_configuration(
+            lambda_name)
+        lambda_python_version = config.lambda_python_version
+        if lambda_config['Runtime'] != lambda_python_version:
+            return False
+        return True
+
     def _update_lambda_function(self, config, lambda_name, stage_name):
         # type: (Config, str, str) -> Dict[str, Any]
         print("Updating lambda function...")
@@ -696,9 +708,10 @@ class LambdaDeployer(object):
         packager = self._packager
         deployment_package_filename = packager.deployment_package_filename(
             project_dir)
-        if self._osutils.file_exists(deployment_package_filename):
-            packager.inject_latest_app(deployment_package_filename,
-                                       project_dir)
+        if self._can_reuse_deployment_package(config, lambda_name,
+                                              deployment_package_filename):
+            packager.inject_latest_app(
+                deployment_package_filename, project_dir)
         else:
             deployment_package_filename = packager.create_deployment_package(
                 project_dir)
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -54,8 +54,8 @@ def create_default_deployer(session, prompter=None):
     aws_client = TypedAWSClient(session)
     api_gateway_deploy = APIGatewayDeployer(aws_client)
 
-    packager = LambdaDeploymentPackager()
     osutils = OSUtils()
+    packager = LambdaDeploymentPackager(osutils=osutils)
     lambda_deploy = LambdaDeployer(
         aws_client, packager, prompter, osutils,
         ApplicationPolicyHandler(
@@ -565,7 +565,8 @@ class LambdaDeployer(object):
         role_arn = self._get_or_create_lambda_role_arn(
             config, api_handler_name)
         zip_contents = self._osutils.get_file_contents(
-            self._packager.deployment_package_filename(config.project_dir),
+            self._packager.deployment_package_filename(
+                config.project_dir, config.lambda_python_version),
             binary=True)
         function_name = api_handler_name + '-' + name
         if self._aws_client.lambda_function_exists(function_name):
@@ -661,7 +662,7 @@ class LambdaDeployer(object):
         print("Initial creation of lambda function.")
         role_arn = self._get_or_create_lambda_role_arn(config, function_name)
         zip_filename = self._packager.create_deployment_package(
-            config.project_dir)
+            config.project_dir, config.lambda_python_version)
         zip_contents = self._osutils.get_file_contents(
             zip_filename, binary=True)
 
@@ -689,32 +690,19 @@ class LambdaDeployer(object):
             return DEFAULT_LAMBDA_MEMORY_SIZE
         return config.lambda_memory_size
 
-    def _can_reuse_deployment_package(self, config, lambda_name,
-                                      deployment_package_filename):
-        # type: (Config, str, str) -> bool
-        if not self._osutils.file_exists(deployment_package_filename):
-            return False
-        lambda_config = self._aws_client.get_function_configuration(
-            lambda_name)
-        lambda_python_version = config.lambda_python_version
-        if lambda_config['Runtime'] != lambda_python_version:
-            return False
-        return True
-
     def _update_lambda_function(self, config, lambda_name, stage_name):
         # type: (Config, str, str) -> Dict[str, Any]
         print("Updating lambda function...")
         project_dir = config.project_dir
         packager = self._packager
         deployment_package_filename = packager.deployment_package_filename(
-            project_dir)
-        if self._can_reuse_deployment_package(config, lambda_name,
-                                              deployment_package_filename):
+            project_dir, config.lambda_python_version)
+        if self._osutils.file_exists(deployment_package_filename):
             packager.inject_latest_app(
                 deployment_package_filename, project_dir)
         else:
             deployment_package_filename = packager.create_deployment_package(
-                project_dir)
+                project_dir, config.lambda_python_version)
         zip_contents = self._osutils.get_file_contents(
             deployment_package_filename, binary=True)
         role_arn = self._get_or_create_lambda_role_arn(config, lambda_name)
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -1,87 +1,94 @@
 from __future__ import print_function
+import sys
 import hashlib
 import inspect
 import os
+import re
 import shutil
-import subprocess
 import zipfile
+import tarfile
+import subprocess
+from email.parser import FeedParser
+from email.message import Message  # noqa
 
-import virtualenv
-from typing import Any, Optional  # noqa
+from typing import Any, Set, List, Optional, Tuple, Iterable, Callable  # noqa
+from chalice.compat import lambda_abi
+from chalice.utils import OSUtils
+from chalice.constants import MISSING_DEPENDENCIES_TEMPLATE
 
 import chalice
-from chalice import compat, app
+from chalice import app
+
+
+class InvalidSourceDistributionNameError(Exception):
+    pass
+
+
+class MissingDependencyError(Exception):
+    def __init__(self, missing):
+        # type: (Set[Package]) -> None
+        self.missing = missing
+
+
+class PipError(Exception):
+    pass
 
 
 class LambdaDeploymentPackager(object):
     _CHALICE_LIB_DIR = 'chalicelib'
     _VENDOR_DIR = 'vendor'
 
-    def _create_virtualenv(self, venv_dir):
-        # type: (str) -> None
-        virtualenv.create_environment(venv_dir)
+    def __init__(self, dependency_builder=None):
+        # type: (Optional[DependencyBuilder]) -> None
+        self._osutils = OSUtils()
+        if dependency_builder is None:
+            dependency_builder = DependencyBuilder(self._osutils)
+        self._dependency_builder = dependency_builder
+
+    def _get_requirements_file(self, project_dir):
+        # type: (str) -> str
+        # Gets the path to a requirements.txt file out of a project dir path
+        return self._osutils.joinpath(project_dir, 'requirements.txt')
 
     def create_deployment_package(self, project_dir, package_filename=None):
         # type: (str, Optional[str]) -> str
         print("Creating deployment package.")
-        # pip install -t doesn't work out of the box with homebrew and
-        # python, so we're using virtualenvs instead which works in
-        # more cases.
-        venv_dir = os.path.join(project_dir, '.chalice', 'venv')
-        self._create_virtualenv(venv_dir)
-        pip_exe = compat.pip_script_in_venv(venv_dir)
-        assert os.path.isfile(pip_exe)
-        # Next install any requirements specified by the app.
-        requirements_file = os.path.join(project_dir, 'requirements.txt')
+        # Now we need to create a zip file and add in the site-packages
+        # dir first, followed by the app_dir contents next.
         deployment_package_filename = self.deployment_package_filename(
             project_dir)
         if package_filename is None:
             package_filename = deployment_package_filename
-        if self._has_at_least_one_package(requirements_file) and not \
-                os.path.isfile(package_filename):
-            p = subprocess.Popen([pip_exe, 'install', '-r', requirements_file],
-                                 stdout=subprocess.PIPE)
-            p.communicate()
-        deps_dir = compat.site_packages_dir_in_venv(venv_dir)
-        assert os.path.isdir(deps_dir)
-        # Now we need to create a zip file and add in the site-packages
-        # dir first, followed by the app_dir contents next.
-        dirname = os.path.dirname(os.path.abspath(package_filename))
-        if not os.path.isdir(dirname):
-            os.makedirs(dirname)
+        try:
+            self._dependency_builder.build_site_packages(project_dir)
+        except MissingDependencyError as e:
+            missing_packages = '\n'.join([p.identifier for p
+                                          in e.missing])
+            print(MISSING_DEPENDENCIES_TEMPLATE % missing_packages)
+        site_packages_dir = self._dependency_builder.site_package_dir(
+            project_dir)
+        dirname = self._osutils.dirname(
+            self._osutils.abspath(package_filename))
+        if not self._osutils.directory_exists(dirname):
+            self._osutils.makedirs(dirname)
         with zipfile.ZipFile(package_filename, 'w',
                              compression=zipfile.ZIP_DEFLATED) as z:
-            self._add_py_deps(z, deps_dir)
+            self._add_py_deps(z, site_packages_dir)
             self._add_app_files(z, project_dir)
-            self._add_vendor_files(z, os.path.join(project_dir,
-                                                   self._VENDOR_DIR))
+            self._add_vendor_files(z, self._osutils.joinpath(project_dir,
+                                                             self._VENDOR_DIR))
         return package_filename
 
     def _add_vendor_files(self, zipped, dirname):
         # type: (zipfile.ZipFile, str) -> None
-        if not os.path.isdir(dirname):
+        if not self._osutils.directory_exists(dirname):
             return
         prefix_len = len(dirname) + 1
-        for root, _, filenames in os.walk(dirname):
+        for root, _, filenames in self._osutils.walk(dirname):
             for filename in filenames:
-                full_path = os.path.join(root, filename)
+                full_path = self._osutils.joinpath(root, filename)
                 zip_path = full_path[prefix_len:]
                 zipped.write(full_path, zip_path)
-
-    def _has_at_least_one_package(self, filename):
-        # type: (str) -> bool
-        if not os.path.isfile(filename):
-            return False
-        with open(filename, 'r') as f:
-            # This is meant to be a best effort attempt.
-            # This can return True and still have no packages
-            # actually being specified, but those aren't common
-            # cases.
-            for line in f:
-                line = line.strip()
-                if line and not line.startswith('#'):
-                    return True
-        return False
 
     def deployment_package_filename(self, project_dir):
         # type: (str) -> str
@@ -89,10 +96,11 @@ class LambdaDeploymentPackager(object):
         # based on a hash of the requirements file.
         # This is done so that we only "pip install -r requirements.txt"
         # when we know there's new dependencies we need to install.
-        requirements_file = os.path.join(project_dir, 'requirements.txt')
+        requirements_file = self._get_requirements_file(project_dir)
         hash_contents = self._hash_project_dir(
-            requirements_file, os.path.join(project_dir, self._VENDOR_DIR))
-        deployment_package_filename = os.path.join(
+            requirements_file, self._osutils.joinpath(project_dir,
+                                                      self._VENDOR_DIR))
+        deployment_package_filename = self._osutils.joinpath(
             project_dir, '.chalice', 'deployments', hash_contents + '.zip')
         return deployment_package_filename
 
@@ -105,7 +113,7 @@ class LambdaDeploymentPackager(object):
                 # what we want to include in _add_app_files.
                 dirnames.remove('chalice')
             for filename in filenames:
-                full_path = os.path.join(root, filename)
+                full_path = self._osutils.joinpath(root, filename)
                 zip_path = full_path[prefix_len:]
                 zip_fileobj.write(full_path, zip_path)
 
@@ -121,7 +129,7 @@ class LambdaDeploymentPackager(object):
             chalice_init = chalice_init[:-1]
         zip_fileobj.write(chalice_init, 'chalice/__init__.py')
 
-        zip_fileobj.write(os.path.join(project_dir, 'app.py'),
+        zip_fileobj.write(self._osutils.joinpath(project_dir, 'app.py'),
                           'app.py')
         self._add_chalice_lib_if_needed(project_dir, zip_fileobj)
 
@@ -180,7 +188,7 @@ class LambdaDeploymentPackager(object):
         print("Regen deployment package...")
         tmpzip = deployment_package_filename + '.tmp.zip'
         with zipfile.ZipFile(deployment_package_filename, 'r') as inzip:
-            with zipfile.ZipFile(tmpzip, 'w') as outzip:
+            with zipfile.ZipFile(tmpzip, 'w', zipfile.ZIP_DEFLATED) as outzip:
                 for el in inzip.infolist():
                     if self._needs_latest_version(el.filename):
                         continue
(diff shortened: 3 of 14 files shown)


chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
                         self._CHALICE_LIB_DIR,
                         fullpath[len(libdir) + 1:])
                     zip_fileobj.write(fullpath, zip_path)
+
+
+class DependencyBuilder(object):
+    """Build site-packages by manually downloading and unpacking whls.
+
+    Pip is used to download all the dependency sdists. Then wheels that
```
stealthycoin: that compatible -> that are compatible

tests/functional/test_package.py
```diff
@@ -23,6 +34,589 @@ def index():
...
+        pip, runner = pip_runner
+        appdir, builder = self._make_appdir_and_dependency_builder(
+            reqs, tmpdir, runner)
+        requirements_file = os.path.join(appdir, 'requirements.txt')
+        pip.packages_to_download(
+            expected_args=['-r', requirements_file, '--dest', mock.ANY],
+            packages=[
+                'foo-1.2-cp36-cp36m-manylinux1_x86_64.whl'
+            ],
+            whl_contents=['{data_dir}/purelib/foo/placeholder']
```
kyleknap: The {data_dir} part is a little confusing for me. What does that represent?

tests/functional/test_package.py
```diff
@@ -23,6 +34,589 @@ def index():
...
+        pip, runner = pip_runner
+        appdir, builder = self._make_appdir_and_dependency_builder(
+            reqs, tmpdir, runner)
+        requirements_file = os.path.join(appdir, 'requirements.txt')
+        pip.packages_to_download(
+            expected_args=['-r', requirements_file, '--dest', mock.ANY],
+            packages=[
+                'foo-1.2-cp36-cp36m-manylinux1_x86_64.whl'
+            ],
+            whl_contents=['{data_dir}/platlib/foo/placeholder']
```
kyleknap: Also I am not very sure what the placeholder part of the string for `whl_contents` represents. Could you elaborate on what the placeholder is for?

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+    def __init__(self, pip):
+        # type: (PipWrapper) -> None
+        self._wrapped_pip = pip
+
+    def _execute(self, command, args):
+        # type: (str, List[str]) -> None
+        """Execute a pip command with the given arguments."""
+        main_args = [command] + args
+        _, err = self._wrapped_pip.main(main_args)
+        if err:
```
stealthycoin: Yes it will fail if it fails to get packages, which is not uncommon for smaller simple pure python packages. That is too be expected and not an error. All I really care about is if pip throws us a stack trace of some kind via stderr, and the rc does not vary for that so I can't sense what kind of error it was without parsing stderr.

I talked to @dstufft about this and if I remember correctly this was what we arrived at as the best solution for catching these types of issues.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        with self._osutils.tempdir() as tempdir:
+            package_dir = self._unpack_sdist_into_dir(sdist_path, tempdir)
+            pkg_info_filepath = self._generate_egg_info(package_dir)
+            metadata = self._parse_pkg_info_file(pkg_info_filepath)
+            name = metadata['Name']
+            version = metadata['Version']
+        return name, version
+
+
+class PipWrapper(object):
```
stealthycoin: I subclassed it for testing purposes as well and wanted to ensure that the type signature contract is not broken by the testing class. This way mypy will catch it if either of them get out of sync. I could also move the logic from the `SubprocessPip` up to `PipWrapper`and just have the testing inherit from `PipWrapper` and override the `main` method. I thought it was clearer what the purpose of everything was this way.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        # unmet dependencies left over. At this point there is nothing we can
+        # do about any missing wheel files.
+        valid_whls, _ = self._categorize_whl_files(directory)
+        missing_whls = deps - valid_whls
+        return valid_whls, missing_whls
+
+    def _install_purelib_and_platlib(self, whl, root):
+        # type: (Package, str) -> None
+        # Take a wheel package and the directory it was just unpacked into and
+        # properly unpackage the purelib and platlib subdirectories if they
```
stealthycoin: I'll reword the comment.

chalice/deploy/deployer.py
```diff
@@ -596,16 +596,29 @@ def _get_lambda_memory_size(self, config):
             return DEFAULT_LAMBDA_MEMORY_SIZE
         return config.lambda_memory_size
 
+    def _can_reuse_deployment_package(self, config, lambda_name,
+                                      deployment_package_filename):
+        # type: (Config, str, str) -> bool
+        if not self._osutils.file_exists(deployment_package_filename):
+            return False
+        lambda_config = self._aws_client.get_function_configuration(
```
stealthycoin: It looked like it would require too much refactoring of things not related to my PR so I didn't but yes that call is made twice which is inefficient.

chalice/utils.py
```diff
@@ -101,6 +104,60 @@ def set_file_contents(self, filename, contents, binary=True):
...
+
+    def joinpath(self, *args):
+        # type: (str) -> str
+        return os.path.join(*args)
+
+    def walk(self, path):
+        # type: (str) -> Iterator[Tuple[str, List[str], List[str]]]
+        return os.walk(path)
+
+    def copytree(self, source, destination):
```
stealthycoin: https://github.com/python/cpython/blob/master/Lib/shutil.py#L315 which fails if the directory already exists.

tests/functional/test_package.py
stealthycoin: I just need a garbage file in the directory so that folder exists. Maybe I'm conflating how zip works with how git works. I didn't try it without the placeholders.

tests/functional/test_package.py
stealthycoin: The data directory in an unpacked wheel file.

tests/functional/test_package.py
stealthycoin: Offline discussion: removed the placeholder, I didn't realize empty folders worked in zip/tar for some reason.

tests/functional/test_package.py
stealthycoin: Offline discussion: Those were there for the implicit case where I do not supply wheel contents and I need to calculate the data directory name anyway so I reused that mechanism. Its clearer to write it out in the explicit tests so I will change it to do that rather than rely on the format string.

chalice/deploy/packager.py
stealthycoin: Looks like the typing doesn't actually help the way I thought it did, so I'm not sure exactly how to accomplish what I was trying for here.

chalice/deploy/packager.py
stealthycoin: Offline discussion: Need to talk with donald about this more in detail. But the initial download case can probably rely on rc, there rest will need to tolerate an rc != 1 so we are stuck parsing the stderr.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+            return False
+        # Verify that the ABI is compatible with lambda. Either none or the
+        # correct type for the python version cp27mu for py27 and cp36m for
+        # py36.
+        if abi == 'none':
+            return True
+        prefix_version = implementation[:3]
+        if prefix_version == 'cp3':
+            # Deploying python 3 function which means we need cp36m abi
+            return abi == 'cp36m'
```
jamesls: Can we just use the `lambda_abi` from `compat.py`?

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        return valid_whls, invalid_whls
+
+    def _download_dependencies(self, directory, requirements_file):
+        # type: (str, str) -> Tuple[Set[Package], Set[Package]]
+        # Download all dependencies we can, letting pip choose what to download
+        deps = self._download_all_dependencies(requirements_file, directory)
+
+        # Sort the downloaded packages into three categories:
+        # - sdists (Pip could not get a wheel so it gave us a sdist)
+        # - valid whls (lambda compatbile wheel files)
```
jamesls: Typo

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        # - invalid whls (lambda incompatible wheel files)
+        valid_whls, invalid_whls = self._categorize_whl_files(directory)
```
jamesls: This section is confusing to me.  Why not use `deps` from `_download_all_dependencies` and categorize accordingly rather than recreating the list of `Package` objects from a directory name?

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+
+    def _download_dependencies(self, directory, requirements_file):
+        # type: (str, str) -> Tuple[Set[Package], Set[Package]]
+        # Download all dependencies we can, letting pip choose what to download
+        deps = self._download_all_dependencies(requirements_file, directory)
+
+        # Sort the downloaded packages into three categories:
+        # - sdists (Pip could not get a wheel so it gave us a sdist)
+        # - valid whls (lambda compatbile wheel files)
+        # - invalid whls (lambda incompatible wheel files)
```
jamesls: Why not just call it that (`lambda_incompatible_wheels`) rather than have a comment about what an "invalid whl" is?

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        valid_whls, invalid_whls = self._categorize_whl_files(directory)
+        sdists = deps - valid_whls - invalid_whls
+
+        # Find which packages we do not yet have a valid whl file for. And
```
jamesls: I think I've made this comment before, but rather than explaining what the code is doing, it would really helpful if you explain *why* you're doing this.  As someone not as familiar with the intricacies of pip, a high level overview of why we're doing this multipass approach, why an invalid wheel might be able to be "converted" to a valid manylinux1 wheel, etc.  Putting a more detailed explanation of your algorithm outlined in the PR description would be better in the code itself IMO.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+                self._osutils.copytree(source, root)
+                # No reason to keep the purelib/platlib source directory around
+                # so we delete it to conserve space in the package.
+                self._osutils.rmtree(source)
+
+    def _install_whls(self, src_dir, dst_dir, whls):
+        # type: (str, str, Set[Package]) -> None
+        if os.path.isdir(dst_dir):
+            shutil.rmtree(dst_dir)
+        os.makedirs(dst_dir)
```
jamesls: If you're introducing `OSUtils` as a dependency for this object, the point of that is it abstracts *all* os operations to allow for granular testing, error conditions, etc.  Using `shutil` and `os` directly in the same object defeats that purpose.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+            metadata = self._parse_pkg_info_file(pkg_info_filepath)
+            name = metadata['Name']
+            version = metadata['Version']
+        return name, version
+
+
+class PipWrapper(object):
+    def main(self, args):
+        # type: (List[str]) -> Tuple[Optional[bytes], Optional[bytes]]
+        raise NotImplementedError('PipWrapper.main')
```
jamesls: I think we usually just use the method name and not the base class.  If a subclass doesn't implement this they'll see the base class's name in the error message which can be confusing.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        arguments = ['--no-deps', '--wheel-dir', directory, wheel]
+        self._execute('wheel', arguments)
+
+    def download_all_dependencies(self, requirements_file, directory):
+        # type: (str, str) -> None
+        """Download all dependencies as sdist or whl."""
+        arguments = ['-r', requirements_file, '--dest', directory]
+        self._execute('download', arguments)
+
+    def download_manylinux_whls(self, packages, directory):
```
jamesls: Nitpick, but it's not 100% consistent in this code if we use "wheel" or "whl" (`build_wheel` vs. `download_manylinux_whls`.  FWIW, I prefer "wheel", it's two extra characters, makes it more readable, and is the name of the concept, whereas "whl" is the file extension.

setup.py
```diff
@@ -9,9 +9,9 @@
 install_requires = [
     'click==6.6',
     'botocore>=1.5.40,<2.0.0',
-    'virtualenv>=15.0.0,<16.0.0',
     'typing==3.5.3.0',
     'six>=1.10.0,<2.0.0',
+    'pip>=9'
```
dstufft: This is a bit tricky to do with pip I think, because while we're not setuptools style, we do rev our versions fairly regularly, but our breaking changes tend to be removing things that few people use rather than completely changing the common paths. So while we could do this, we should expect to need to rev this once or twice a year.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+        # type: (OSUtils) -> None
+        if osutils is None:
+            osutils = OSUtils()
+        self._osutils = osutils
+
+    def _parse_pkg_info_file(self, filepath):
+        # type: (str) -> Message
+        with open(filepath, 'r') as f:
+            data = f.read()
+        parser = FeedParser()
```
dstufft: FWIW that's the API that the stdlib provides for parsing a PKG-INFO file, largely for historical reasons. It could use the pkg-info library though.

chalice/deploy/packager.py
jamesls: Capturing offline discussion: I retract my previous statement.

chalice/deploy/packager.py
```diff
@@ -190,21 +217,416 @@ def inject_latest_app(self, deployment_package_filename, project_dir):
...
+        p = subprocess.Popen(invoke_pip,
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        _, err = p.communicate()
+        rc = p.returncode
+        return rc, err
+
+
+class PipRunner(object):
+    """Wrapper around pip calls used by chalice."""
+
```
JordonPhillips: Sorry to nitpick, but this is the only place you've got a space between the class doc string and the first method.

chalice/utils.py
```diff
@@ -101,6 +116,84 @@ def set_file_contents(self, filename, contents, binary=True):
...
+            yield z
+        finally:
+            z.close()
+
+    def unpack_tarfile(self, tarfile_path, unpack_dir):
+        # type: (str, str) -> None
+        with tarfile.open(tarfile_path, 'r:gz') as tar:
+            tar.extractall(unpack_dir)
+
+    def directory_exists(self, path):
```
JordonPhillips: These should probably stay close to the names of the functions they're aliasing.

chalice/utils.py
JordonPhillips: Yeah copytree is pretty annoying like that.

chalice/deploy/packager.py
```diff
@@ -190,21 +217,416 @@ def inject_latest_app(self, deployment_package_filename, project_dir):
...
+        # type: (str, str) -> str
+        if sdist_path.endswith('.zip'):
+            self._osutils.unpack_zipfile(sdist_path, unpack_dir)
+        elif sdist_path.endswith('.tar.gz'):
+            self._osutils.unpack_tarfile(sdist_path, unpack_dir)
+        else:
+            raise InvalidSourceDistributionNameError(sdist_path)
+        # There should only be one directory unpacked.
+        contents = self._osutils.get_directory_contents(unpack_dir)
+        assert len(contents) == 1
```
stealthycoin: There should be no case where there is any customer impact from this. The line below the assert operates under the assumption that the upstream code is generating a directory with exactly one entry. This assertion acts as a dev-time safeguard since the tests may pass by happenstance depending on which file in the directory is opened with `[0]`, and it is not a useful thing to test since it is an implementation detail. 

If anyone violates that assumption during development of a feature or refactoring the assertion will explode and point them to what and where that assumption was. Then if it's no longer valid they can remove it and modify the code below it to work with their upstream code. Or they can fix their code to not break the assumption.

chalice/deploy/packager.py
```diff
@@ -208,3 +216,373 @@ def _add_chalice_lib_if_needed(self, project_dir, zip_fileobj):
...
+
+    def __str__(self):
+        # type: () -> str
+        return '%s(%s)' % (self.identifier, self.dist_type)
+
+    def __repr__(self):
+        # type: () -> str
+        return str(self)
+
+    def __eq__(self, other):
```
stealthycoin: Not equal doesn't really make sense in this context and the equal set and unequal set are not disjoint so I don't think its worth implementing as it should not be used and currently isn't.

chalice/utils.py
```diff
@@ -92,6 +98,15 @@ def get_file_contents(self, filename, binary=True):
         with open(filename, mode) as f:
             return f.read()
 
+    @contextlib.contextmanager
+    def get_file_context(self, filename, mode='r'):
```
jamesls: Why is this needed?  Is there a reason you can't use `OSUtils.open` directly?

chalice/utils.py
```diff
@@ -101,6 +116,84 @@ def set_file_contents(self, filename, contents, binary=True):
         with open(filename, mode) as f:
             f.write(contents)
 
+    def unpack_zipfile(self, zipfile_path, unpack_dir):
```
jamesls: This is picky, but I think we should use verbs that are commonly associated with whatever the noun is, so for a zipfile, I would expect "unzip", or "extract" but not "unpack".

chalice/utils.py
```diff
@@ -101,6 +116,84 @@ def set_file_contents(self, filename, contents, binary=True):
...
+        # type: (str, str) -> None
+        with zipfile.ZipFile(zipfile_path, 'r') as z:
+            z.extractall(unpack_dir)
+
+    @contextlib.contextmanager
+    def zipfile_context(self, filename, mode='r',
```
jamesls: Why do we need to wrap the `zipfile` with its own context manager?  This is already a capability of `zipfile.ZipFile` so it seems unnecessary.

chalice/utils.py
```diff
@@ -101,6 +116,84 @@ def set_file_contents(self, filename, contents, binary=True):
...
+                        compression=zipfile.ZIP_STORED):
+        # type: (str, str, int) -> Iterator[zipfile.ZipFile]
+        try:
+            z = zipfile.ZipFile(filename, mode=mode, compression=compression)
+            yield z
+        finally:
+            z.close()
+
+    def unpack_tarfile(self, tarfile_path, unpack_dir):
```
jamesls: Same comment as zip.  Maybe "untar" or "extract"

setup.py
jamesls: Once or twice a year seems reasonable, I'm just concerned that it's still a possibility within semver to break a common API (even if it's rare), and people on older versions of chalice won't be guarded against this.

chalice/deploy/packager.py
jamesls: I think I may have either added this or a similar assertion somewhere in the packager code.  I would also suggest removing these assertions.  The only reason I added them previously was during the initial POC and they just never got removed.  They generate unhelpful error messages.  If this is a case a customer can run into, raise a specific exception, otherwise remove the assertion.

chalice/deploy/packager.py
```diff
@@ -1,116 +1,142 @@
...
-        dirname = os.path.dirname(os.path.abspath(package_filename))
-        if not os.path.isdir(dirname):
-            os.makedirs(dirname)
-        with zipfile.ZipFile(package_filename, 'w',
-                             compression=zipfile.ZIP_DEFLATED) as z:
-            self._add_py_deps(z, deps_dir)
+        try:
+            requirements_filepath = self._get_requirements_filename(
+                project_dir)
+            site_packages_dir = self._osutils.joinpath(
```
jamesls: Does this directory not need to be python version aware?  How does it handle the case where the python version is switched and different requirements are specified?  Wouldn't it be safer to just have a separate py-specific site-packages directory to avoid any possibility of invalid packages?

stealthycoin: Site packages are not reused, I was considering deleting it, the only reason its not deleted was for debugging purposes so I could see what the folder looked like, its convenient if something goes wrong to be able to see what was packaged before and after vendoring.

jamesls: Ah, I missed that that happens in `_install_wheels`.  While it seems a little surprising that a method like `build_site_packages` will wipe out the `target_directory` if it exists (arguably even dangerous), seems like my original concern isn't an issue.

stealthycoin: I suppose there are three options:
* Delete it first and rebuild which is safe with regards to building a working site-packages directory. Not safe if contents was pre-merged in rather than post-merged in like we do currently.
* Merge content. Not safe since python versions can change or get removed.
* Error out. 


So either:
* Leave the code as is.
* I leave the directory there after its built, but delete it in the caller of `build_site_packages` and have the it error out if the directory exists.
* Delete it after the package finishes building, and have `build_site_packages` error out  if the directory already exists.

My preference is option 1 or 2 since I like having directory there "just in case" you need to investigate it for some reason.

chalice/constants.py
```diff
@@ -197,3 +197,15 @@ def index():
...
+
+MISSING_DEPENDENCIES_TEMPLATE = r"""
+Could not install dependencies:
+%s
+You will have to build these yourself and vendor them in
+the chalice vendor folder.
+
+Your deployment will continue but may not work correctly
+if missing dependencies are not present. For more information:
+http://chalice.readthedocs.io/en/latest/topics/packaging.html
```
stealthycoin: Got a draft up.

docs/source/topics/packaging.rst
```diff
@@ -35,9 +35,9 @@ write yourself.
 There are two options for handling python package dependencies:
 
 * **requirements.txt** - During the packaging process, chalice will
-  run ``pip install -r requirements.txt`` in a virtual environment
-  and automatically install 3rd party python packages into the deployment
-  package.
+  install any packages it find or build compatible wheels for. Specifically
```
jamesls: I think the wording is off here.  Did you mean "...install any packages it finds or builds compatible wheels for" maybe?

docs/source/topics/packaging.rst
```diff
@@ -35,9 +35,9 @@ write yourself.
...
+  all pure python projects as well as all projects that upload wheel files for
```
jamesls: What about packages instead of projects?

chalice/deploy/packager.py
jamesls: I'm ok with option 2.  That allows `build_site_packages` to just focus on populating the directory and it's the caller's responsibility to put it into whatever state it needs to be (empty in this case).

Though my personal preference would be to just delete the directory.  It's of no use to chalice users (we don't reuse the directory across deploys) and it just takes up space.  If we want dev debugging hooks, we can always add them later and explicitly turn them on (or ask users to run with debugging mode enabled).

stealthycoin: I can just nuke the directory and we can allow debugging options later. The debug hooks would need to give you access to it rather than just logging it though since sometimes you need to go look in files and go pretty deep into nested directories, more than a simple log will give.

jamesls: Ok that seems fine.  Though in that case you can just use a `OSUtils.tempdir` so it can handle cleaning up the directory when you're done.

stealthycoin: Yep thats the plan. Ill just make the debug hook copy it out somewhere before it gets erased.

tests/functional/test_package.py
```diff
@@ -23,6 +33,585 @@ def index():
...
+                for content_path in self._whl_contents:
+                    z.writestr(content_path.format(
+                        package_name=self._package_name,
+                        data_dir=package.data_dir
+                    ), b'')
+
+    def _build_fake_sdist(self, filepath):
+        # tar.gz is the same no reason to test it here as it is tested in
+        # unit.deploy.TestSdistMetadataFetcher
+        assert filepath.endswith('zip')
```
JordonPhillips: `.zip`

tests/functional/test_package.py
```diff
@@ -23,6 +33,585 @@ def index():
...
+            ],
+            whl_contents=['foo-1.2.data/purelib/foo/']
+        )
+
+        site_packages = os.path.join(appdir, '.chalice.', 'site-packages')
+        builder.build_site_packages(requirements_file, site_packages)
+        installed_packages = os.listdir(site_packages)
+
+        pip.validate()
+        for req in reqs:
```
JordonPhillips: There's only one of these so it's a bit odd to iterate. This happens in a few other places as well.

docs/source/topics/packaging.rst
```diff
@@ -94,54 +95,82 @@ This directory structure is then zipped up and sent to AWS Lambda during the
...
-Psycopg2 Example
-----------------
+Cryptography Example
+--------------------
+
+Below shows an example of how to use the
+`cryptography <https://pypi.python.org/pypi/cryptography>`__ package in a
+Chalice app for the ``python3.6`` lambda environment.
+
+We're going to leverage the ``vendor/`` directory in order to use this package
```
kyleknap: It might be worth making the mention that since chalice will try to build the wheel for you and you are on the right system, it may be able to work via the `requirements.txt`. I am only suggesting it because it came off to me as `cryptography` **will** require additional steps to be part of the deployment package.

tests/functional/test_package.py
stealthycoin: Ah, thought I caught all those.

tests/functional/test_package.py
stealthycoin: Makes it easy to update the tests if I need to.

docs/source/topics/packaging.rst
jamesls: What about just saying something like "Suppose we're on a Mac and we want to deploy an app that uses `cryptography`".  That keeps it more user-centric and less about implementation details of chalice building wheels.  You could also say that this wouldn't be an issue if we were on the same platform that Lambda uses.

kyleknap: That would be fine as well. I just want to make sure a user sees this and does not immediately start trying to follow the steps when it actually would have worked for them to start.

tests/functional/test_package.py
JordonPhillips: Yeah I'm -0 on keeping them. This is not a PR blocker.

PR #393, opened 2017-06-29
Add support for scheduled events

jamesls:
Proposal: https://github.com/awslabs/chalice/issues/390

Sample app:

```python
from chalice import Chalice
import datetime

app = Chalice(app_name='testevents')


@app.route('/')
def index():
    return {'hello': 'world'}


@app.schedule('rate(1 minute)')
def every_minute(event):
    print(event.to_dict())

```

The implementation piggy backs off of the multi lambda support from the built in authorizers.  The only unexpected change was an update to the deployed.json file from:

```
    "lambda_functions": {
      "testevents-dev-every_minute": "arn:lambda-foo"
    },

```
to:

```
    "lambda_functions": {
      "testevents-dev-every_minute": {
        "type": "scheduled_event",
        "arn": "arn:lambda-foo"
      }
    },

```

So we can differentiate the different types of lambda functions.

The other caveat is that right now you still need to deploy an API gateway API.  There isn't support yet for only deploying scheduled events.  That will be in a follow up PR.

**Review Notes**

The deployer module has some duplication that I'm working on removing.  The logic itself won't change but I'm refactoring the code.  Everything else (`app.py`, `awsclient.py`, all the tests) are ready for review.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -491,7 +491,7 @@ class Chalice(object):
                 execution_role=execution_role,
             )
             self.builtin_auth_handlers.append(auth_config)
-            return ChaliceAuthorizer(name, auth_func, auth_config)
+            return ChaliceAuthorizer(auth_name, auth_func, auth_config)
         return _register_authorizer
 
     def schedule(self, expression, name=None):
@@ -499,12 +499,12 @@ class Chalice(object):
             handler_name = name
             if handler_name is None:
                 handler_name = event_func.__name__
-            event_source = EventSource(
+            event_source = CloudWatchEventSource(
                 name=handler_name,
                 schedule_expression=expression,
                 handler_string='app.%s' % event_func.__name__)
             self.event_sources.append(event_source)
-            return ScheduluedEventHandler(event_func)
+            return ScheduledEventHandler(event_func)
         return _register_schedule
 
     def route(self, path, **kwargs):
@@ -792,10 +792,15 @@ class AuthRoute(object):
 
 
 class EventSource(object):
-    def __init__(self, name, schedule_expression, handler_string):
+    def __init__(self, name, handler_string):
         self.name = name
-        self.schedule_expression = schedule_expression
         self.handler_string = handler_string
+
+
+class CloudWatchEventSource(EventSource):
+    def __init__(self, name, handler_string, schedule_expression):
+        super(CloudWatchEventSource, self).__init__(name, handler_string)
+        self.schedule_expression = schedule_expression
 
 
 class ScheduleExpression(object):
@@ -841,7 +846,7 @@ class Cron(ScheduleExpression):
         )
 
 
-class ScheduluedEventHandler(object):
+class ScheduledEventHandler(object):
     def __init__(self, func):
         self.func = func
 
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -428,6 +428,7 @@ class Chalice(object):
         self.log = logging.getLogger(self.app_name)
         self._authorizers = {}
         self.builtin_auth_handlers = []
+        self.event_sources = []
         if self.configure_logs:
             self._configure_logging()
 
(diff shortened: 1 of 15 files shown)


chalice/app.py
```diff
@@ -765,3 +779,82 @@ class AuthRoute(object):
...
+        event_obj = self._convert_to_obj(event)
+        return self.func(event_obj)
+
+    def _convert_to_obj(self, event_dict):
+        return CloudWatchEvent(event_dict)
+
+
+class CloudWatchEvent(object):
+    def __init__(self, event_dict):
+        self.version = event_dict['version']
```
jamesls: Yes.

chalice/app.pyi
```diff
@@ -148,3 +149,31 @@ class AuthResponse(object):
...
+    schedule_expression = ...  # type: Union[str, ScheduleExpression]
+    handler_string = ...  # type: str
+
+
+class ScheduleExpression(object):
+    def to_string(self) -> str: ...
+
+
+class Rate(ScheduleExpression):
+    unit = ... # type: str
```
jamesls: That should be an integer, I'll update.

chalice/app.py
```diff
@@ -765,3 +779,82 @@ class AuthRoute(object):
     def __init__(self, path, methods):
         self.path = path
         self.methods = methods
+
+
+class EventSource(object):
```
jamesls: It's intentional to have an `EventSource` type.  The idea would be this is used for all event types as support more types is added.  I don't envision adding a new public attr to `chalice.Chalice` for every new event source, that wouldn't be maintainable.  I can extract out a separate `CloudWatchEventSource` class with `EventSource` as its base type.  Seemed like overkill for now, but I don't mind adding it.

kyleknap: Makes sense. I like the idea of having `EventSource` be the base class and `CloudWatchEventSource` subclass from it to include any of the cloudwatch event specific logic.

jamesls: I pushed an update.  Let me know what you think.

tests/functional/test_awsclient.py
```diff
@@ -906,17 +906,48 @@ def test_can_add_permission_for_apigateway_not_needed(self, stubbed_session):
...
+                 'Principal': {'Service': 'apigateway.amazonaws.com'},
+                 'Resource': 'arn:aws:lambda:us-west-2:account_id:function:name',
+                 'Sid': 'e4755709-067e-4254-b6ec-e7f9639e6f7b'},
+                {'Action': 'lambda:InvokeFunction',
+                 'Condition': {
+                     'ArnLike': {
+                         'AWS:SourceArn': source_arn,
+                     }
+                 },
+                 'Effect': 'Allow',
```
kyleknap: Without looking at what got removed in the diff, it is not very easy to tell what portion of the policy needs to be present for it to be skipped. Maybe separate this into tests where one tests that a specific policy entry skips over the `add_permission` call but the other entries causes `add_permission` to be triggered?

chalice/app.py
```diff
@@ -765,3 +779,87 @@ class AuthRoute(object):
...
+    def to_string(self):
+        unit = self.unit.lower()
+        if self.value == 1:
+            # Remove the 's' from the end if it's singular.
+            # This is required by the cloudwatch events API.
+            unit = unit[:-1]
+        return 'rate(%s %s)' % (self.value, unit)
+
+
+class Cron(ScheduleExpression):
```
kyleknap: Did we want to add `Cron` and `Rate` to the `__init__`?

jamesls: I don't follow what you're asking.  Can you elaborate?

tests/functional/test_awsclient.py
jamesls: I refactored the test a bit to make this more clear.

chalice/app.py
kyleknap: Oh so I have to import those two from `chalice.app` right now, but I'm use to importing all of the user facing classes from `chalice` instead when I right my application.

jamesls: ohhhh, `chalice/__init__.py`.  Gotcha.  Yeah I can add that.

kyleknap:
Cool. Looks good once docs are included, it should be good to merge. I also noticed that there was some missing coverage according to the codecov report. Do you know what that may be?

jamesls:
Yeah it was some of the new code I added for the base class refactoring.  I added a `isinstance` check that's missing a testcase if you don't pass an instance of `CloudWatchEventSource`.  Right now there are no other kinds so I can make one up for the test.

jamesls:
OK I think everything has been addressed, and docs have been added.  Let me know if there's any more feedback.

cc @kyleknap @stealthycoin

docs/source/api.rst
```diff
@@ -118,6 +118,46 @@ Chalice
...
+      This will create a lambda function associated with the decorated
+      function.  It will also schedule the lambda function to be invoked
+      with a scheduled CloudWatch Event.
+
+      See :doc:`events` for more information.
+
+      .. code-block:: python
+
+          @app.schedule('cron(15 10 ? * 6L 2002-2005)')
+          def handler(event):
```
jamesls: You mean not share?

kyleknap: yes not share

jamesls: OK I'll update that before merging.

PR #395, opened 2017-07-03
Fix unicode strings being quoted in python 2.7.

### LINKED ISSUE #262: Passing a unicode type body to Response causes body to be json encoded
https://github.com/awslabs/chalice/blob/a8d68083e8e76f096307cdd992686b658e39f1a2/chalice/app.py#L144-L145

If you pass a unicode typed body variable to the Response object it json encodes the value.

# Steps to reproduce

```

app = Chalice(app_name='example')

@app.route('/')
def index():
    body = u'<h1>test</h1>'
    return Response(body=body,
                    status_code=200,
                    headers={'Content-Type': 'text/html'})
```

# Expected results

```
<h1>test</h1>
```

# Actual results

```
"<h1>test</h1>"
```

# Suggested change

It would make sense to check for `str` or `unicode` strings [here](https://github.com/awslabs/chalice/blob/a8d68083e8e76f096307cdd992686b658e39f1a2/chalice/app.py#L144-L145)

stealthycoin:
Not sure how exactly we want to do the compat layer. Still have no explicit checks against python version.

fixes #262 

cc @jamesls @kyleknap  @dstufft @JordonPhillips

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -17,6 +17,16 @@ from collections import defaultdict, Mapping
 
 _PARAMS = re.compile(r'{\w+}')
 
+try:
+    # In python 2 there is a base class for the string types that
+    # we can check for. It was removed in python 3 so it will cause
+    # a name error.
+    _ANY_STRING = (basestring, bytes)
+except NameError:
+    # In python 3 string and bytes are different so we explicitly check
+    # for both.
+    _ANY_STRING = (str, bytes)
+
 
 def handle_decimals(obj):
     # Lambda will automatically serialize decimals so we need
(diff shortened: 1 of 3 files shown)


chalice/app.py
```diff
@@ -299,7 +309,7 @@ def __init__(self, body, headers=None, status_code=200):
 
     def to_dict(self, binary_types=None):
         body = self.body
-        if not isinstance(body, (str, bytes)):
+        if not isinstance(body, _STRING_TYPES):
```
jamesls: Can you rename this constant?  It took me a while to figure this out because I was assuming this mapped to `six.string_types` which doesn't include the bytes type in either py2 or py3.  mypy using `AnyStr` for this type so maybe that would work?

PR #397, opened 2017-07-04
ISSUE-396 enhances local to utilize environment_variables from config.json

sbraverman:
Issue for this PR:
https://github.com/awslabs/chalice/issues/396

Running `chalice local` with environment variables set inside of `config.json` now works as expected:
```
(venv-chalice) Braverman@local:~/Repos/fix-local-env-vars $ http localhost:8000
HTTP/1.0 200 OK
Content-Length: 78
Date: Tue, 04 Jul 2017 17:42:01 GMT
Server: BaseHTTP/0.3 Python/2.7.10

{
    "env_vars": {
        "py": "con", 
        "rick": "Morty", 
        "test": "True"
    }, 
    "hello": "world"
}

```

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -73,6 +73,16 @@ def cli(ctx, project_dir, debug=False):
     os.chdir(project_dir)
 
 
+def get_env_variables(config):
+    # type: (Dict) -> Dict
+    env_vars_key = 'environment_variables'
+    env_vars = {}  # type: Dict
+    if env_vars_key in config and config[env_vars_key]:
+        env_vars = {key: value
+                    for key, value in config[env_vars_key].items()}
+    return env_vars
+
+
 @cli.command()
 @click.option('--port', default=8000, type=click.INT)
 @click.pass_context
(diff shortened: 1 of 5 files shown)


stealthycoin:
Thanks for the pull request! Two comments:

* Could you update your tests to comply with PEP8
* Remove the usage of `mock.patch`. We don't use patches in Chalice testing as a rule, a linter will be added eventually so it will fail `make prcheck` but hasn't yet. Classes in Chalice have been designed so they are easy to instantiate in a testing environment, so there should be no need for patching out objects. There are also a number of fixtures already available that should construct anything you need.

Thanks again for the contribution!

sbraverman:
@stealthycoin thanks for reviewing!

I had a couple hours tonight to take a look. I removed the patching and made the unit tests abide to PEP8.
After looking into fixture for a bit and studying the codebase, I decided to take a different approach to get the environment variables into the local app.
Please take a look. Let me know if approach and coverage change is acceptable.

I can also squash the commits into one (Wanted to keep both commits for review purposes). 
Let me know how we can proceed from here.

jamesls:
Thanks for the PR.  Taking a look now.

jamesls:
One of the things I'm noticing when trying out this PR is that if any module level code tries to access os.environ, `chalice local` will crash.  To be fair, this needs to be mapped for all commands (`chalice deploy`, `chalice local`, etc) but I wanted to get the process started with the `local` command.

So I took this PR and made a few changes here: https://github.com/awslabs/chalice/pull/411.  Let me know what you think.

sbraverman:
Thanks for picking it up. Closing this PR in favor of:
https://github.com/awslabs/chalice/pull/411

PR #401, opened 2017-07-05
Add support for pure lambda functions

jamesls:
Proposal: https://github.com/awslabs/chalice/issues/400

Sample code:

```python
from chalice import Chalice

app = Chalice(app_name='testpurelambda')

@app.lambda_function()
def foo(event, context):
    return {"name": "foo"}


@app.lambda_function()
def bar(event, context):
    return {"name": "bar"}


@app.route('/')
def index():
    return {'hello': 'world'}
```

This will create 3 lambda functions:

```
 $ aws lambda list-functions --query "Functions[?starts_with(FunctionName, 'testpurelambda')].FunctionName"
[
    "testpurelambda-dev-foo",
    "testpurelambda-dev-bar",
    "testpurelambda-dev"
]
```

### DIFF
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -77,6 +77,7 @@ def validate_configuration(config):
     validate_route_content_types(routes, config.chalice_app.api.binary_types)
     _validate_manage_iam_role(config)
     validate_python_version(config)
+    validate_unique_function_names(config)
 
 
 def validate_routes(routes):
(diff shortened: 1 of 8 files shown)


docs/source/api.rst
```diff
@@ -159,6 +159,18 @@ Chalice
         entire lambda function name.  This parameter is optional.  If it is
         not provided, the name of the python function will be used.
 
+   .. method:: lambda_function(name=None)
+
+      Create a pure lambda function that's not connected to anything.
+
+      See :doc:`purelambda` for more information.
+
+      :param name: The name of the function to use.  This name is combined
```
stealthycoin: Can you say how it is combined. Like: {app}-{stage}-{name} notation or give an example. Or do that in the extended `purelambda` docs.

chalice/deploy/deployer.py
```diff
@@ -421,11 +421,25 @@ def deploy(self, config, existing_resources, stage_name):
...
         self._deploy_event_sources(config, existing_resources, stage_name,
                                    deployed_values)
+        self._deploy_pure_lambda_functions(config, existing_resources,
+                                           stage_name, deployed_values)
         if existing_resources is not None:
             self._cleanup_unreferenced_functions(existing_resources,
                                                  deployed_values)
         return deployed_values
 
+    def _deploy_pure_lambda_functions(self, config, existing_resources,
```
stealthycoin: From looking at this logic it seems I can create two pure lambda functions with the same name, and the deployment will crash out with a ResourceAlreadyExistsException. Is that correct? Can we do some validation like we do on routes here we check that they are unique if this is the case?

docs/source/api.rst
jamesls: I'm hesitant to do that because it's an implementation details and can (and has) changed a few times.  If I do document it I would have to say not to rely on this behavior.

chalice/deploy/deployer.py
jamesls: It will just override the first lambda function (it will trigger the function already exists check which will make it think it's an update instead of an initial create).

I can add validation for this.  It's a little more involved because you have to check uniqueness across all lambda functions we deploy (scheduled events, custom auth, pure lambda).

jamesls:
Feedback incorporated, added validation for dupe names.  Also fixed the travis build.  Let me know if there's anything else.

PR #402, opened 2017-07-05
Add Beta 1 Branch

jamesls:
This starts a tracking `rc1` branch that removes all the deprecated features so far.  It will need to get rebased with master once all the features have been merged, but this review should be independent of any new incoming features to master.


jamesls:
Rebased against latest master and repushed.

jamesls:
Cool, incorporated feedback so far.  Also changed a the pre-alpha trove classifier to be beta (there wasn't an RC one from what I can tell).

jamesls:
I intentionally left that in.  It's still in awslabs, and it is just an RC release.  I think it makes sense to remove the dev preview label once we get a GA release and change the org.

kyleknap:
@jamesls So then what was the reasoning for taking the preview mention out of the prompt in the command line then?

jamesls:
Fair point, I mostly wanted to reduce the wording a bit, but I can add back the bit about it being under developer preview.  In fact, I can just match the text from the README so it'll be easier to find/delete at the 1.0 release.

kyleknap:
Sounds good to me.

CHANGELOG.rst
```diff
@@ -2,8 +2,16 @@
...
-Next Release (TBD)
-==================
+1.0.0b1
+=======
+
+Please read the `upgrade notes for 1.0.0b1
+<http://chalice.readthedocs.io/en/latest/upgrading.html#v1-0-0b1>`__
+for more detailed information about upgrading to this release.
+
+Note: to install this RC version of chalice you must specify
```
kyleknap: Not an RC version anymore. May want to make it beta version?

chalice/constants.py
```diff
@@ -190,6 +190,13 @@ def index():
 you to quickly create and deploy applications using
 Amazon API Gateway and AWS Lambda.
 
+This project is published as a preview project.
+Until the 1.0 GA release, we may introduce backwards
+incompatible changes based on customer feedback we receive.
+As a best practice, we do not recommend launching production
+APIs until 1.0 GA because of potential backward incompatible
```
kyleknap: Or should we recommending to pin to a version if you want to use it in production?

jamesls: No I don't think so.   That warning has been in the README for almost a year now so I'd like to leave it as is.

kyleknap: Ok that's fine. Forgot that came from the README

PR #408, opened 2017-07-09
reduce package size by ignoring __pycache__ folder

Segflow:
This reduced my package size by 60%

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -23,6 +23,7 @@ class InvalidSourceDistributionNameError(Exception):
 
 class MissingDependencyError(Exception):
     """Raised when some dependencies could not be packaged for any reason."""
+
     def __init__(self, missing):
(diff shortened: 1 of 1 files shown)


stealthycoin:
@Segflow Thanks for the submission! I have three bits of feedback:

* We would need tests that cover the new cases you are adding.
* The style changes are not necessary.
* I think the situation is a little more complicated than this. Including the pycache/pyc files can improve your load time a little bit on every cold start. It may not seem like a lot but that will add up over all the cold starts, so hardcoding it to be removed may not give the flexibility some users need, and there may need to be additions to this blacklist of filetypes/names in the future. My feedback would be to make a blacklist of files to be ignored rather than hardcoding __pycache__ in two places. Then when copying files check against the blacklist before including it in the bundle. This will give us more flexibility later.

stealthycoin:
Closing this in favor of #1068. That approach was closer to what I had in mind for this PR.

PR #411, opened 2017-07-11
Set env vars from config.json when running chalice local

jamesls:
This builds on #397 and pulls up setting up the env vars up to cli/__init__.py.  We may need to pull this up even higher because we need to make sure env vars are set before ever importing the chalice app.  Otherwise we'll crash with KeyErrors on import is module level code accesses `os.environ`.

cc @stealthycoin @kyleknap

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -12,7 +12,7 @@ import shutil
 
 import botocore.exceptions
 import click
-from typing import Dict, Any, Optional  # noqa
+from typing import Dict, Any, Optional, MutableMapping  # noqa
 
 from chalice import __version__ as chalice_version
 from chalice.app import Chalice  # noqa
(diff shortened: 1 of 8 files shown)


sbraverman:
@jamesls thanks for picking this up and running with it. Overall updates look good. We can close my original PR in favor of this one.
I have a couple questions:
* What is the benefit of passing `os.environ` into `run_local_server` as opposed to just using `os.environ` directly inside of `run_local_server`?
i.e. `os.environ.update(config.environment_variables)` inside of `run_local_server`

* One thing I was trying to do in the original PR was ensure all environment_variables were casted to Strings.
Running the current code with this change currently does...
```
(venv-chalice) Braverman@local:~/Repos/fix-local-env-vars $ chalice local
must be string, not bool
```

Do you think:
* another validation check should be added to `_validate_config_from_disk` to ensure proper strings are used for environment_variables 
* environment variable values should get casted to Strings automatically 

Let me know if that makes sense or more clarification is necessary. Thanks again for looking into this.

jamesls:
Hi, sorry for the delay, I've been on vacation for a few weeks.  To answer your questions:

1. Primarily just to make testing easier.  Not a huge deal but I prefer the existing way so you pass in your own os.environ as well
2. For validation, I think raising an error would make sense here.  Here's what you currently get if you use an invalid type:

```
$ chalice deploy
Updating IAM policy.
Updating lambda function...
Regen deployment package...
Updating IAM policy.
Sending changes to lambda.
Parameter validation failed:
Invalid type for parameter Environment.Variables.foo, value: 5, type: <type 'int'>, valid types: <type 'basestring'>
```

Let me know if you have any more questions.

PR #419, opened 2017-07-19
fix: handle cases where the role policy doesn't exist

josegonzalez:
It shouldn't try to delete non-existent role policies, which can happen if the put_role_policy call fails.


jamesls:
Thanks for the pull request.  I'd prefer to just move this into awsclient with a new "replace_role_policy" or something that contains these four lines.  It's a pretty small change so I'll go ahead an update this.

josegonzalez:
Closing as this was never merged.

PR #421, opened 2017-07-21
Fix edge case when building packages with optional c extensions

stealthycoin:
The edge case occurs when a package can compile optional c speedups
which succeed during the build process but lock the generated wheel file
to the arch the package was built on. This makes the wheel file
incompatible with lambda and causes it to fail to add that dependency
even though it could have by building it without the optional c
extensions.

This is a bit tricky to fix as means we need to selectively break the
ability to compile c extensions during some runs of our wheel building
phase. It is also quite different on POSIX and Windows.

cc @kyleknap @jamesls @dstufft

### DIFF
--- a/chalice/compat.py
+++ b/chalice/compat.py
@@ -1,7 +1,111 @@
 import socket
 import six
+import os
 
 from six import StringIO
+
+if os.name == 'nt':
+    # windows
+    # On windows running python in a subprocess with no environment variables
+    # will cause sevral issues. In order for our subprocess to run normally we
+    # manually copy the relevant environment variables from the parent process.
+    subprocess_python_base_environ = {
+        # http://bugs.python.org/issue8557
+        'PATH': os.environ['PATH']
+    }
+    # http://bugs.python.org/issue20614
+    if 'SYSTEMROOT' in os.environ:
+        subprocess_python_base_environ['SYSTEMROOT'] = os.environ['SYSTEMROOT']
+
+    # This is the acutal patch used on windows to prevent distutils from
+    # compiling C extensions. The msvc compiler base class has its compile
+    # method overridden to raise a CompileError. This can be caught by
+    # setup.py code which can then fallback to making a purepython
+    # package if possible.
+    # We need mypy to ignore these since they are never actually called from
+    # within ourprocess they do not need to be a part of our typechecking
+    # pass.
+    def prevent_msvc_compiling_patch():  # type: ignore
+        import distutils
+        import distutils._msvccompiler
+        import distutils.msvc9compiler
+        import distutils.msvccompiler
+
+        from distutils.errors import CompileError
+
+        def raise_compile_error(*args, **kwargs):  # type: ignore
+            raise CompileError('Chalice blocked C extension compiling.')
+        distutils._msvccompiler.MSVCCompiler.compile = raise_compile_error
+        distutils.msvc9compiler.MSVCCompiler.compile = raise_compile_error
+        distutils.msvccompiler.MSVCCompiler.compile = raise_compile_error
+
+    # This is the setuptools shim used to execute setup.py by pip.
+    # Lines 2 and 3 have been added to call the above function
+    # `prevent_msvc_compiling_patch` and extra escapes have been added on line
+    # 5 because it is passed through another layer of string parsing before it
+    # is executed.
+    _SETUPTOOLS_SHIM = (
+        "import setuptools, tokenize;__file__=%r;"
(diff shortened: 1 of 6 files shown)


chalice/compat.py
```diff
@@ -1,8 +1,112 @@
 import socket
 import six
+import os
 
 from six import StringIO
 
+if os.name == 'nt':
+    # windows
+    # On windows running python in a subprocess with no environment variables
+    # will cause sevral issues. In order for our subprocess to run normally we
```
jamesls: s/sevral/several

chalice/compat.py
```diff
@@ -1,8 +1,112 @@
...
+    # manually copy the relevant environment variables from the parent process.
+    subprocess_python_base_environ = {
+        # http://bugs.python.org/issue8557
+        'PATH': os.environ['PATH']
+    }
+    # http://bugs.python.org/issue20614
+    if 'SYSTEMROOT' in os.environ:
+        subprocess_python_base_environ['SYSTEMROOT'] = os.environ['SYSTEMROOT']
+
+    # This is the acutal patch used on windows to prevent distutils from
```
jamesls: s/acutal/actual

chalice/compat.py
```diff
@@ -1,8 +1,112 @@
...
+    # compiling C extensions. The msvc compiler base class has its compile
+    # method overridden to raise a CompileError. This can be caught by
+    # setup.py code which can then fallback to making a purepython
```
jamesls: s/purepython/pure python/

chalice/compat.py
```diff
@@ -1,8 +1,112 @@
...
+        distutils._msvccompiler.MSVCCompiler.compile = raise_compile_error
+        distutils.msvc9compiler.MSVCCompiler.compile = raise_compile_error
+        distutils.msvccompiler.MSVCCompiler.compile = raise_compile_error
+
+    # This is the setuptools shim used to execute setup.py by pip.
+    # Lines 2 and 3 have been added to call the above function
+    # `prevent_msvc_compiling_patch` and extra escapes have been added on line
+    # 5 because it is passed through another layer of string parsing before it
+    # is executed.
+    _SETUPTOOLS_SHIM = (
```
jamesls: The escaping makes this hard to read.  Could you just use `r"""` to avoid all the double escaping?

chalice/compat.py
```diff
@@ -1,8 +1,112 @@
...
+    # SETUPTOOLS_SHIM that pip normally uses with the one defined above.
+    # When pip goes to run its subprocess for executing setup.py it will
+    # inject _SETUPTOOLS_SHIM rather than the usual SETUPTOOLS_SHIM in pip.
+    # This lets us apply our patches in the same process that will compile
+    # the c extensions before the setup.py file has been executed.
+    # The actual patches used are decribed in the comment above
+    # _SETUPTOOLS_SHIM.
+    pip_no_compile_c_shim = (
+        "import pip;"
+        "pip.wheel.SETUPTOOLS_SHIM = \"\"\"%s\"\"\";"
```
jamesls: Same thing.  Can you use `r"""` to cut down on the amount of escaping needed?

chalice/compat.py
```diff
@@ -1,8 +1,112 @@
...
+    ) % _SETUPTOOLS_SHIM
+    pip_no_compile_c_env_vars = {}  # type: ignore
```
jamesls: I'd still like to get these typed.  These are passed to code that we own in package.py.

chalice/deploy/packager.py
```diff
@@ -374,12 +384,25 @@ def _download_dependencies(self, directory, requirements_filename):
 
         # Re-count the wheel files after the second download pass. Anything
         # that has an sdist but not a valid wheel file is still not going to
-        # work on lambda and our last ditch effort is to try and build the
-        # sdists into wheel files.
+        # work on lambda and our we must now try and build the sdist into a
```
JordonPhillips: s/our we must now/we must now/

docs/source/topics/packaging.rst
```diff
@@ -50,7 +50,9 @@ specific examples).  The ``vendor/`` directory is helpful in these scenarios:
 * You need to include custom packages or binary content that is not accessible
   via ``pip``.  These may be internal packages that aren't public.
 * Wheel files are not available for a package you need from pip.
-
+* A package is installable with ``requirements.txt`` but has optional c
+  extensions. Chalice can build the dependency without the c extensions, but
+  if you want better performance you can vendor a version taht is compiled.
```
JordonPhillips: s/taht/that

PR #437, opened 2017-07-28
Add test for custom CORS local & fix bug

corespace:
Potential fix for https://github.com/awslabs/chalice/issues/436

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -187,24 +187,32 @@ class ChaliceRequestHandler(BaseHTTPRequestHandler):
         else:
             # Otherwise this is a preflight request which we automatically
             # generate.
-            self._send_autogen_options_response()
+            self._send_autogen_options_response(lambda_event)
 
     def _has_user_defined_options_method(self, lambda_event):
         # type: (EventType) -> bool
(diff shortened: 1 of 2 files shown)


stealthycoin:
Hey thanks for the pull request! Looks good to me. I made a few changes and tried to update this PR but it wouldn't let me so I opened another PR at #444.

I updated the code a little bit to explain what was happening. In particular https://github.com/corespace/chalice/blob/f63d8e7f62296c84b37a62125371fde6092a2a67/chalice/local.py#L201 was a little bit to magical for me. I don't know whats happening here from reading it and if I ever need to debug or change this I need some context.

The OPTIONS CORS response also was missing the `Access-Control-Allow-Methods` header. That header is set up during our [deployment to API Gateway](https://github.com/awslabs/chalice/blob/master/chalice/deploy/swagger.py#L180-L186). The test you wrote looks like you used the test that preforms a GET as a template. That test doesn't need that header since it is performing a GET request, which has already passed the OPTIONS stage where `Access-Control-Allow-Methods` would have been returned to either allow or disallow the GET request.

Thanks again for the submission!

corespace:
Thanks! Closing this one.

PR #444, opened 2017-07-29
Add support for local OPTIONS CORS headers

### LINKED ISSUE #436: chalice local does not respect CORS config
Not sure if this is expected or not, but it appears that `chalice local` just ignores my CORS config and returns some defaults: https://github.com/awslabs/chalice/blob/master/chalice/local.py#L197-L208

I would prefer if I could test my CORS config locally. Please let me know if I'm missing something obvious...

stealthycoin:
Looks like I couldn't push directly to the #437 for some reason so I opened a separate PR.

closes #437 
fixes #436 

cc @kyleknap @jamesls

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -198,9 +198,30 @@ class ChaliceRequestHandler(BaseHTTPRequestHandler):
         # type:(EventType) -> None
         route_key = lambda_event['requestContext']['resourcePath']
         route_dict = self.app_object.routes[route_key]
-        first_m = route_dict[next(iter(route_dict))]
+        route_methods = list(route_dict.keys())
+
+        # Chalice ensures that routes with multiple views have the same
+        # CORS configuration, so if any view has a CORS Config we can use
+        # that config since they will all be the same.
+        cors_config = route_dict[route_methods[0]].cors
+        cors_headers = cors_config.get_access_control_headers()
+
+        # We need to add OPTIONS since it is not a part of the CORSConfig
+        # object. APIGateway handles this entirely based on the API definition.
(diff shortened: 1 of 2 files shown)


corespace:
Great commenting. I wasn't sure if you folks would like the change hence my quick-n-dirty fix. Thanks for reviewing!

chalice/local.py
```diff
@@ -187,24 +187,43 @@ def do_OPTIONS(self):
...
+        route_methods = list(route_dict.keys())
+
+        # Chalice ensures that routes with multiple views have the same
+        # CORS configuration, so if any view has a CORS Config we can use
+        # that config since they will all be the same.
+        cors_config = route_dict[route_methods[0]].cors
+        cors_headers = cors_config.get_access_control_headers()
+
+        # The keys will be ordered in python 3 but in python 2 they will not
+        # be. To assert this header is sent correctly we need to introduce
```
kyleknap: It is not necessarily true about the need to sort this for the assertion. You could always just split the header value by commas and assert that the sorted list is the same, but that is a bit more tedious. I only bring it up because it is a little weird to have this sorted in the core logic when it is only being sorted for the test. I'm about +0 in updating this especially since we do the same logic in the non chalice local tests (i.e. assert the header values in the a split list).

kyleknap: Talked offline about this. The reason it is being done is that there is not an easy way to parse the raw body and pull out the specific header/value without using something like the `email` library. Given that this is being ran in `chalice local` it is fine that it is getting sorted as `chalice local` is just a test mode. If this was being done in the actual lambda function, it would be something we would not want to do and should add the extra logic in the tests.

PR #446, opened 2017-07-31
Fix log level out of sync with debug mode

### LINKED ISSUE #386: Setting app.debug does not propagate to app.log logger log level
In https://github.com/awslabs/chalice/blob/master/chalice/app.py#L434-L437, intent is shown to set the logger's log level according to the value of Chalice.debug. However, app.debug is currently hardcoded to false in `Chalice.__init__`, so without calling `Chalice. _configure_logging()` again or setting it manually via `app.log.setLevel(logging.DEBUG)`, it will always remain set to logging.ERROR.

It would probably make more sense to have a `Chalice.debug` setter that also sets the logger's level, and/or have a `Chalice(debug=True)` kwarg to make the code linked above behave as expected.

stealthycoin:
Debug mode can be set with `app.debug = True` which does not update
the logger level from ERROR to DEBUG. This is has been replaced with a
setter that reconfigures the log level after debug is reset. The
initialization method on the Chalice class had some logic that checked
what the debug value was and configured the logger to match; however,
debug was hardcoded to be False so this never did anything useful. A
debug parameter has been added to the the __init__ method to make this
more useful, it defaults to False.

fixes #386 
cc @jamesls @kyleknap

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -418,12 +418,12 @@ class Chalice(object):
 
     FORMAT_STRING = '%(name)s - %(levelname)s - %(message)s'
 
-    def __init__(self, app_name, configure_logs=True, env=None):
+    def __init__(self, app_name, debug=False, configure_logs=True, env=None):
         self.app_name = app_name
         self.api = APIGateway()
         self.routes = defaultdict(dict)
         self.current_request = None
-        self.debug = False
+        self._debug = debug
         self.configure_logs = configure_logs
         self.log = logging.getLogger(self.app_name)
         self.builtin_auth_handlers = []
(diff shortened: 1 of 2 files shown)


chalice/app.py
```diff
@@ -418,12 +418,12 @@ class Chalice(object):
 
     FORMAT_STRING = '%(name)s - %(levelname)s - %(message)s'
 
-    def __init__(self, app_name, configure_logs=True, env=None):
+    def __init__(self, app_name, debug=False, configure_logs=True, env=None):
```
kyleknap: Do we want to put this argument at the end just in case users are providing these options as positional arguments?

jamesls: My preference given we're pre GA is to order them based on frequent usage and I would expect debug to be more common than the other options.  So I vote to keep it as is.

kyleknap: Alright that's fine with me. Just wanted to make sure people were aware of it.

PR #415, opened 2017-07-16
Add lambda context to app properties

jeff-minard-ck:
An application may need access to the variables inside the lambda context such as any of these things:
```
{
    "aws_request_id": "0ff8542d-6a2b-11e7-b421-xxxxxxxxxxxx",
    "log_stream_name": "2017/07/16/[$LATEST]xxxcccxxxxxxxxxxxxxcccxxxxxxxxxx",
    "invoked_function_arn": "arn:aws:lambda:us-west-1:nnnnnnnnn:function:xxxxxx",
    "client_context": null,
    "log_group_name": "/aws/lambda/xxxxxxx",
    "function_name": "xxxxxxx",
    "function_version": "$LATEST",
    "memory_limit_in_mb": "256"
}
```
This provides a barebones minimum way of doing so.

PR #454, opened 2017-08-01
Add lambda context

stealthycoin:
Adds the lambda context argument to the app object.

closes #415 

cc @kyleknap @jamesls

### DIFF
--- a/chalice/app.pyi
+++ b/chalice/app.pyi
@@ -109,6 +109,9 @@ class Chalice(object):
     api = ... # type: APIGateway
     routes = ... # type: Dict[str, Dict[str, RouteEntry]]
     current_request = ... # type: Request
+    # TODO: Change lambda_context to a real type once we have one for local
+    # API Gateway
+    lambda_context = ... # type: Any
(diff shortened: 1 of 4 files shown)


tests/unit/test_app.py
```diff
@@ -241,6 +241,20 @@ def test_no_view_function_found(sample_app):
...
 
 
+def test_can_access_context():
+    demo = app.Chalice('app-name')
+
+    @demo.route('/index')
+    def index_view():
+        return demo.lambda_context
+
+    event = create_event('/index', 'GET', {})
```
jamesls: The context isn't a dictionary here.

chalice/app.pyi
```diff
@@ -109,6 +109,7 @@ class Chalice(object):
     api = ... # type: APIGateway
     routes = ... # type: Dict[str, Dict[str, RouteEntry]]
     current_request = ... # type: Request
+    lambda_context = ... # type: Any
```
jamesls: Can y we at least put a TODO here to type this properly once we add `local` support?

PR #458, opened 2017-08-04
Show user friendly error message if config.json is invalid

achautha:
If my config.json is invalid json, I see this error message in chalice package.
```
$ chalice package .
Expecting property name: line 5 column 5 (char 67) 
```

This is not a very friendly error message and took me a while to figure out my config.json was invalid. It would help if we print a proper error message.


jamesls:
Thanks for the pull request.  I've rebased against master and merged via https://github.com/aws/chalice/commit/bc18c8b9c56f0c4603120215dfb7c339f317be39

PR #463, opened 2017-08-07
Improve output for 'chalice deploy'

jamesls:
This adds more detail to the output of `chalice deploy`, including:

* Add resource names where possible (lambda function names, rest api IDs, etc)
* Print when we're deleting unreferenced lambda functions

I've also introduced a module level logger that includes debug messages for the parts of the deployer.  This is more intended for us to help troubleshoot any deployment issues.

I've also restructured the internals to remove bare print calls.  A `UI` object is now passed through the deployer objects.  This makes testing easier as well as makes it easier to add more UI capabilities in the future.  Right now it just has writing to stdout/stderr, and prompting.  I've also moved the `NoPrompt` interface into the `UI` interface.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -209,7 +209,7 @@ class LambdaDeploymentPackager(object):
         # a way to do this efficiently so we need to create a new
         # zip file that has all the same stuff except for the new
         # app file.
-        self._ui.write("Regen deployment package...\n")
+        self._ui.write("Regen deployment package.\n")
         tmpzip = deployment_package_filename + '.tmp.zip'
 
         with self._osutils.open_zip(deployment_package_filename, 'r') as inzip:
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -1,4 +1,3 @@
-from __future__ import print_function
 import sys
 import hashlib
 import inspect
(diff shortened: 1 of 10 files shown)


jamesls:
If I'm understanding codecov correctly, it's complaining that I changed some of the print statements that were previously untested to `self._ui.write`?

chalice/deploy/packager.py
```diff
@@ -87,7 +84,8 @@ def create_deployment_package(self, project_dir, python_version,
             except MissingDependencyError as e:
                 missing_packages = '\n'.join([p.identifier for p
                                               in e.missing])
-                print(MISSING_DEPENDENCIES_TEMPLATE % missing_packages)
+                self._ui.write(
+                    MISSING_DEPENDENCIES_TEMPLATE % missing_packages)
```
stealthycoin: Need a `\n` here?

stealthycoin:
Yes thats what it looks like its complaining about to me.

chalice/deploy/packager.py
jamesls: I cheated and updated the `MISSING_DEPENDENCIES_TEMPLATE` with a newline.

jamesls:
In which case I'm 0 on adding tests for this, given the upcoming refactor to the deployer...

stealthycoin:
Yea I think its fine to ignore for now and fix when you do something more substantial than change `print` to `ui.write`

chalice/deploy/packager.py
stealthycoin: Ah I missed that change.

PR #467, opened 2017-08-10
Add support for local custom authorizers

### LINKED ISSUE #404: Local mode authorizers
Allow for authorizers to work in local mode for better testing.

stealthycoin:
BuiltinAuthorizer will now work in local mode. Authorizer types that
cannot be run locally will now simply allow the requests to proceed as
if there were no authorizer to allow for local testing.

closes #404

### DIFF
--- a/chalice/app.pyi
+++ b/chalice/app.pyi
@@ -1,4 +1,5 @@
 from typing import Dict, List, Any, Callable, Union, Optional
+from chalice.local import LambdaContext
 
 __version__ = ... # type: str
 
@@ -109,9 +110,7 @@ class Chalice(object):
     api = ... # type: APIGateway
     routes = ... # type: Dict[str, Dict[str, RouteEntry]]
     current_request = ... # type: Request
-    # TODO: Change lambda_context to a real type once we have one for local
-    # API Gateway
-    lambda_context = ... # type: Any
+    lambda_context = ... # type: LambdaContext
     debug = ... # type: bool
     authorizers = ... # type: Dict[str, Dict[str, Any]]
     builtin_auth_handlers = ... # type: List[BuiltinAuthConfig]
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -4,27 +4,99 @@ This is intended only for local development purposes.
 
 """
 from __future__ import print_function
+import re
+import time
+import uuid
 import base64
 import functools
+import warnings
 from collections import namedtuple
 
 from six.moves.BaseHTTPServer import HTTPServer
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler
-from typing import List, Any, Dict, Tuple, Callable  # noqa
+from typing import List, Any, Dict, Tuple, Callable, Optional, Union  # noqa
 
-from chalice.app import Chalice, CORSConfig  # noqa
+from chalice.app import Chalice  # noqa
+from chalice.app import CORSConfig  # noqa
+from chalice.app import ChaliceAuthorizer  # noqa
+from chalice.app import RouteEntry  # noqa
+from chalice.app import Request  # noqa
+from chalice.app import AuthResponse  # noqa
+from chalice.app import BuiltinAuthConfig  # noqa
+from chalice.config import Config  # noqa
+
 from chalice.compat import urlparse, parse_qs
 
 
 MatchResult = namedtuple('MatchResult', ['route', 'captured', 'query_params'])
 EventType = Dict[str, Any]
+ContextType = Dict[str, Any]
+HeaderType = Dict[str, Any]
+ResponseType = Dict[str, Any]
 HandlerCls = Callable[..., 'ChaliceRequestHandler']
 ServerCls = Callable[..., 'HTTPServer']
 
 
-def create_local_server(app_obj, port):
-    # type: (Chalice, int) -> LocalDevServer
-    return LocalDevServer(app_obj, port)
+class Clock(object):
+    def time(self):
+        # type: () -> float
+        return time.time()
+
+
+def create_local_server(app_obj, config, port):
+    # type: (Chalice, Config, int) -> LocalDevServer
+    return LocalDevServer(app_obj, config, port)
+
+
+class LocalARNBuilder(object):
+    ARN_FORMAT = ('arn:aws:execute-api:{region}:{account_id}'
+                  ':{api_id}/{stage}/{method}/{resource_path}')
+    LOCAL_REGION = 'mars-west-1'
+    LOCAL_ACCOUNT_ID = '123456789012'
+    LOCAL_API_ID = 'ymy8tbxw7b'
+    LOCAL_STAGE = 'api'
+
+    def build_arn(self, method, path):
+        # type: (str, str) -> str
+        # In API Gateway the method and URI are separated by a / so typically
+        # the uri portion omits the leading /. In the case where the entire
+        # url is just '/' API Gateway adds a / to the end so that the arn end
+        # with a '//'.
+        if path != '/':
+            path = path[1:]
+        return self.ARN_FORMAT.format(
+            region=self.LOCAL_REGION,
+            account_id=self.LOCAL_ACCOUNT_ID,
+            api_id=self.LOCAL_API_ID,
+            stage=self.LOCAL_STAGE,
+            method=method,
+            resource_path=path
+        )
+
+
+class ARNMatcher(object):
+    def __init__(self, target_arn):
+        # type: (str) -> None
+        self._arn = target_arn
+
+    def _resource_match(self, resource):
+        # type: (str) -> bool
+        # Arn matching supports two special case characetrs that are not
+        # escapable. * represents a glob which translates to a non-greedy
+        # match of any number of characters. ? which is any single character.
+        # These are easy to translate to a regex using .*? and . respectivly.
+        escaped_resource = re.escape(resource)
+        resource_regex = escaped_resource.replace(r'\?', '.').replace(
+            r'\*', '.*?')
+        resource_regex = '^%s$' % resource_regex
+        return re.match(resource_regex, self._arn) is not None
+
+    def does_any_resource_match(self, resources):
+        # type: (List[str]) -> bool
+        for resource in resources:
+            if self._resource_match(resource):
+                return True
+        return False
 
 
 class RouteMatcher(object):
@@ -94,7 +166,7 @@ class LambdaEventConverter(object):
                     'sourceIp': self.LOCAL_SOURCE_IP
                 },
             },
-            'headers': dict(headers),
+            'headers': {k.lower(): v for k, v in headers.items()},
             'queryStringParameters': view_route.query_params,
             'pathParameters': view_route.captured,
             'stageVariables': {},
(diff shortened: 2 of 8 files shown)


chalice/local.py
```diff
@@ -109,98 +128,251 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+        # type: (str, int, int, Optional[Clock]) -> None
+        if time_source is None:
+            time_source = Clock()
+        self._time_source = time_source
+        self._start_time = self._current_time_millis()
+        self._max_runtime = runtime_millis
+        self._initialize_lambda_context_attributes(
+            function_name, memory_size)
+
+    def _initialize_lambda_context_attributes(self, function_name,
```
JordonPhillips: It's a bit bizzare to define instance attributes outside of init, especially since you're not doing any complicated initialization or logic to set their initial values.

chalice/local.py
```diff
@@ -109,98 +128,251 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+        lambda_event['requestContext']['authorizer'] = auth_context
+        return lambda_event
 
-    def __init__(self, request, client_address, server, app_object):
-        # type: (bytes, Tuple[str, int], HTTPServer, Chalice) -> None
-        self.app_object = app_object
+    def _prepare_authorizer_event(self, lambda_event):
+        # type: (EventType) -> EventType
+        """Translate event for an authorizer input."""
+        authorizer_event = lambda_event.copy()
```
JordonPhillips: How nested is this structure? if it's more than one level then maybe you should deepcopy.

chalice/local.py
```diff
@@ -109,98 +128,251 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+        # type: (EventType) -> Tuple[Optional[Any], Optional[RouteEntry]]
+        # Authorizer had to be made into an Any type since mypy couldn't
+        # detect that app.ChaliceAuthorizer was callable.
+        resource_path = lambda_event.get(
+            'requestContext', {}).get('resourcePath')
+        http_method = lambda_event['requestContext']['httpMethod']
+        try:
+            route_entry = self._app_object.routes[resource_path][http_method]
+        except KeyError:
+            # If there a key error is raised when trying to get the route entry
```
JordonPhillips: s/If there/If/

chalice/local.py
```diff
@@ -109,98 +128,251 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+            # then this route does not support this method. A method error
+            # will be raised by the chalice handler method. We can ignore it
+            # here by returning no authorizer to avoid duplicating the logic.
+            return None, None
+        authorizer = route_entry.authorizer
+        return authorizer, route_entry
```
JordonPhillips: If `authorizer` is an attribute of `route_entry` why are you not just returning `route_entry`?

chalice/local.py
```diff
@@ -109,98 +128,251 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+
+    def authorize(self, lambda_event, lambda_context):
+        # type: (EventType, LambdaContext) -> Tuple[EventType, LambdaContext]
+        auth_event = self._prepare_authorizer_event(lambda_event)
+        authorizer, route_entry = self._authorizer_and_route_for_event(
+            lambda_event)
+        # Technically we only need to check for authorizer being `None` here
+        # since the route_entry will always match authorizer but mypy needs us
+        # to explicitly check both before we can use them. In any event, if
+        # either of these are not defined then there was no authroizer or
```
JordonPhillips: You seem to have trailed off here.

chalice/local.py
```diff
@@ -222,21 +394,99 @@ def _send_autogen_options_response(self, lambda_event):
...
+
+    def _handle_binary(self, response):
+        # type: (Dict[str,Any]) -> Dict[str,Any]
+        if response.get('isBase64Encoded'):
+            body = base64.b64decode(response['body'])
+            response['body'] = body
+        return response
+
+
+class ChaliceRequestHandler(BaseHTTPRequestHandler):
```
JordonPhillips: A class doc string for this would be nice

chalice/local.py
stealthycoin: I separated the logic that is specific to the actual lambda context initialization vs the initializations I have to do for my fake object to run.

chalice/local.py
stealthycoin: Yes I think this made more sense before a refactor and I didn't think to update it.

chalice/local.py
stealthycoin: The event is a dict of entirely [str, Union[str, Dict[str, str]] so I don't think it needs to be deepcopied unless I am mistaken about something.

chalice/local.py
```diff
@@ -109,98 +128,254 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+        method_pattern = r'(?:\*|%s)' % route_entry.method
+        route_entry_pattern = r'^.*?/%s%s%s$' % (method_pattern,
+                                                 route_entry.uri_pattern,
+                                                 pattern_suffix)
+        route_entry_re = re.compile(route_entry_pattern)
+        # Look through all statements for one that allows us to invoke this
+        # particular view function.
+        for statement in statements:
+            if statement.get('Effect') == 'Allow' and \
+               statement.get('Action') == 'execute-api:Invoke':
```
jamesls: These technically support wildcards as well.  If you're not going to support them, you should probably make a note about eventually supporting them.

chalice/local.py
```diff
@@ -109,98 +128,254 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+        # user can return any policy at all from the custom authorizer.
+        pattern_suffix = ''
+        if route_entry.uri_pattern == '/':
+            # There is a special case when the / is used since the API Gateway
+            # arn will have the / duplicated. On all other routes it will
+            # omit the leading /.
+            pattern_suffix = '/'
+        method_pattern = r'(?:\*|%s)' % route_entry.method
+        route_entry_pattern = r'^.*?/%s%s%s$' % (method_pattern,
+                                                 route_entry.uri_pattern,
```
jamesls: I don't think this is right.  The uri pattern will include the placeholder values so something like this won't work:

```python
@app.route('/foo/{name}', authorizer=demo_auth)
def index(name):
    return {'context': app.current_request.context}

```

If I return `AuthResponse(routes=['/foo/bar'], principal_id='user')` from my authorizer.

What if we mirrored things more closely to real IAM? 

1. Generate the resource ARN for the specific method/view
2. Check if any of the resources in the returned policy match the ARN (accounting for wildcards).

I also would suggest extracting this policy matching out to a separate class so we can add a lot more tests for the edge cases (using `pytest.paramtrize`).

chalice/app.pyi
```diff
@@ -1,3 +1,4 @@
+from chalice.local import LambdaContext
```
jamesls: Import orders, chalice after 3rd party libs.

chalice/local.py
```diff
@@ -4,27 +4,99 @@
...
+        # type: () -> float
+        return time.time()
+
+
+def create_local_server(app_obj, config, port):
+    # type: (Chalice, Config, int) -> LocalDevServer
+    return LocalDevServer(app_obj, config, port)
+
+
+class LocalArnBuilder(object):
```
jamesls: s/Arn/ARN

chalice/local.py
```diff
@@ -4,27 +4,99 @@
...
+            region=self.LOCAL_REGION,
+            account_id=self.LOCAL_ACCOUNT_ID,
+            api_id=self.LOCAL_API_ID,
+            stage=self.LOCAL_STAGE,
+            method=method,
+            resource_path=path
+        )
+
+
+class ArnMatcher(object):
```
jamesls: s/Arn/ARN

chalice/local.py
```diff
@@ -109,98 +181,246 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+            # BuiltinAuthConfig type. Anything else we will err on the side of
+            # allowing local testing by simply admiting the request. Otherwise
+            # there is no way for users to test their code in local mode.
+            warnings.warn(
+                '%s is not a supported in local mode. All requests made '
+                'against a route will be authorized to allow local testing.'
+                % authorizer.__class__.__name__
+            )
+            return lambda_event, lambda_context
+        auth_result = authorizer(auth_event, lambda_context)
```
jamesls: I don't think this is quite right.  If you don't have the appropriate auth header (`Authorization`) in the request, you get a 401 and the authorizer isn't called at all.

chalice/local.py
```diff
@@ -214,29 +434,106 @@ def _send_autogen_options_response(self, lambda_event):
...
+            )
+            self._send_http_response(response)
+        except NotAuthorizedError as e:
+            self._send_not_authorized_response(e.headers)
+        except NoOptionsRouteDefined as e:
+            self._send_http_response_no_body(e.headers)
+
+    def _send_not_authorized_response(self, headers):
+        # type: (HeaderType) -> None
+        self.send_response(401)
```
jamesls: An authorized error where we generate a policy that forbids access to a resource gives a 403, not a 401.

```
$ http https://rest-api-id.execute-api.us-west-2.amazonaws.com/api/ 'Authorization: deny'
HTTP/1.1 403 Forbidden
Connection: keep-alive
Content-Length: 60
Date: Thu, 24 Aug 2017 00:27:57 GMT
Via: 1.1 7f3f42df8af148df1f9f1ee7175987ad.cloudfront.net (CloudFront)
X-Amz-Cf-Id: 4ZhC3GokHeXr68EKakpF6xQWxVvTDZtQiGK253psOcWbPW3M980rGw==
X-Cache: Error from cloudfront
x-amzn-ErrorType: AccessDeniedException
x-amzn-RequestId: 13ff2f6d-8863-11e7-af78-e1cc9bcdaa69

{
    "Message": "User is not authorized to access this resource"
}
```

tests/unit/test_app.py
```diff
@@ -8,6 +8,9 @@
 from chalice import app
 from chalice import NotFoundError
 from chalice import __version__ as chalice_version
+from tests.unit.conftest import create_event
```
jamesls: I don't think conftest is meant to be imported directly (or at least it's something I see).  I'd suggest converting them to fixtures if you want to move them there.

I think I might have missed this in the wheel based packager PR.

chalice/local.py
```diff
@@ -4,27 +4,99 @@
...
+
+def create_local_server(app_obj, config, port):
+    # type: (Chalice, Config, int) -> LocalDevServer
+    return LocalDevServer(app_obj, config, port)
+
+
+class LocalArnBuilder(object):
+    ARN_FORMAT = ('arn:aws:execute-api:{region}:{account_id}'
+                  ':{api_id}/{stage}/{method}/{resource_path}')
+    LOCAL_REGION = 'mars-west-1'
```
jamesls: Just curious, why aren't we using a real region as a default for the local region?

chalice/local.py
```diff
@@ -109,98 +181,246 @@ def create_lambda_event(self, method, path, headers, body=None):
...
+
+class LambdaContext(object):
+    def __init__(self, function_name, memory_size,
+                 runtime_millis=3000, time_source=None):
+        # type: (str, int, int, Optional[Clock]) -> None
+        if time_source is None:
+            time_source = Clock()
+        self._time_source = time_source
+        self._start_time = self._current_time_millis()
+        self._max_runtime = runtime_millis
```
jamesls: `runtime_millis` is pretty vague, but `max_runtime_ms` is a lot clearer.  Is there a reason to not map to call the arg something different than it's instance attribute name?

tests/unit/test_local.py
```diff
@@ -404,5 +559,344 @@ def test_can_create_lambda_event_for_post_with_formencoded_body():
 
 
 def test_can_provide_port_to_local_server(sample_app):
-    dev_server = local.create_local_server(sample_app, port=23456)
+    dev_server = local.create_local_server(sample_app, None, port=23456)
     assert dev_server.server.server_port == 23456
+
+
+class FakeTimeSource(object):
```
jamesls: We usually put all the non test classes/functions at the top of the file before the actual tests.

tests/unit/test_local.py
```diff
@@ -404,5 +559,344 @@ def test_can_create_lambda_event_for_post_with_formencoded_body():
...
+        # API Gateway arn generation where an extra / is appended to the end
+        # of the arn.
+        authorizer = LocalGatewayAuthorizer(demo_app_auth)
+        event = create_event('/', 'GET', {})
+        event['headers']['authorization'] = 'allow'
+        context = LambdaContext(*lambda_context_args)
+        event, context = authorizer.authorize(event, context)
+        assert event['requestContext']['authorizer']['principalId'] == 'user'
+
+    def test_can_authorize_with_lower_case_headers(self, lambda_context_args,
```
jamesls: Isn't this identical to the previous test `test_can_authorize_empty_path`?

tests/unit/test_local.py
```diff
@@ -404,5 +559,344 @@ def test_can_create_lambda_event_for_post_with_formencoded_body():
...
+        def index_view():
+            return {}
+
+        authorizer = LocalGatewayAuthorizer(demo)
+        event = create_event('/index', 'GET', {})
+        context = LambdaContext(*lambda_context_args)
+        event, context = authorizer.authorize(event, context)
+        # Assert that when the authorizer.authorize is called and there is no
+        # authorizer defined for a particular route that it is a noop.
+        assert event == event
```
jamesls: ?

tests/unit/test_local.py
```diff
@@ -404,5 +559,344 @@ def test_can_create_lambda_event_for_post_with_formencoded_body():
...
+        assert context == context
```
jamesls: ?

chalice/local.py
stealthycoin: Wanted to make it very clear it wasn't a real ARN and to force it never to match with any real ARNs unless they explicitly don't care about the region (with a glob). I can change it if you have a reason you would prefer a real region.

chalice/local.py
stealthycoin: I wanted to give the unit in the name is why its not a direct mirror. I agree with `max_runtime_ms` so I'll do that.

tests/unit/test_local.py
stealthycoin: I was keeping it nearby for easier reference. I'll move it.

tests/unit/test_local.py
stealthycoin: Yea...

tests/unit/test_local.py
stealthycoin: Fixed... still passes.

jamesls:
I'm just going through this now, but it looks like there's still some feedback that hasn't been incorporated: https://github.com/aws/chalice/pull/467#discussion_r132775015

The ARN generated for a route that has a captured param is still: `'arn:aws:execute-api:mars-west-1:123456789012:ymy8tbxw7b/api/GET/foo/{name}'` which isn't right.  I tested the example in the linked comment and it still fails in local mode but passes on actual API gateway/lambda.

PR #473, opened 2017-08-11
Support for both api_key_required AND authorizer on routes

nelsonlj:
Changed to support routes that need both an api_key and an authorizer.


jamesls:
Thanks for the pull request.  Per our [Contributing guidelines](https://github.com/aws/chalice/blob/master/CONTRIBUTING.rst) we just need to get a test added for this before this can get merged.

Would you be able to add a test in tests/unit/deploy/test_swagger.py for this fix?

jamesls:
Rebased against master and added a test for this here: https://github.com/aws/chalice/pull/473

Thanks again for the pull request!

PR #475, opened 2017-08-14
Fix API Gateway stage not being set correctly

### LINKED ISSUE #470: Stage feature not functioning as expected
When attempting to use ```chalice deploy --stage dev``` the dev API Gateway is given the /api context and when using ```chalice deploy --stage prod``` the Prod API Gateway is given the /dev context. It appears order does not matter and I've deleted and started from scratch and still seeing issue. #

stealthycoin:
The API Gateway stage is not set correctly since it was renamed from dev
to api. This change enforces the 'api' default all the way through the
config layer and CLI layer rather than making the end code that consumes
the configuration manually enforce a default everywhere the api gateway
stage name is used.

fixes #470

@achautha Thanks for the contribution. I added a few more tests and more strict defaults in the Config object itself.

closes #474

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Next Release (TBD)
   (`#450 <https://github.com/aws/chalice/issues/450>`__)
 * Print useful error message when config.json is invalid
   (`#458 <https://github.com/aws/chalice/pull/458>`__)
+* Fix api gateway stage being set incorrectly in non-default chalice stage
+ (`#$70 <https://github.com/aws/chalice/issues/470>`__)
 
 
 1.0.0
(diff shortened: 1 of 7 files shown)


jasonrhaas:
Just ran into this problem.  Any idea when the fix will get pushed to a new PyPi release?

jamesls:
[Version 1.0.1](https://pypi.python.org/pypi/chalice/1.0.1), which includes this fix, was just released on pypi.

johntmyers:
I'm using 1.0.1 and within config.json, `api_gateway_stage` is totally ignored and `api` is used. I currently have to use the `--api-gateway-stage` option to force the actual stage I want.

jamesls:
@johntmyers Thanks for reporting, I've created an issue to track this here: https://github.com/aws/chalice/issues/495

PR #476, opened 2017-08-15
Split shell and python code block

0atman:
Primarily to allow highlighting of the first python code that the reader sees

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,6 +30,8 @@ It provides:
     $ chalice new-project helloworld && cd helloworld
     $ cat app.py
 
+.. code-block:: python
+
     from chalice import Chalice
 
     app = Chalice(app_name="helloworld")
@@ -37,6 +39,8 @@ It provides:
     @app.route("/")
     def index():
         return {"hello": "world"}
+
+::
 
     $ chalice deploy
     ...


jamesls:
I like being able to highlight the python code block, but I notice that render as separate sections now: 

image

 Is there a way we can make the all still part of the same code block?

0atman:
Yes, I sympathise. I don't see a way (apart from screenshotting and editing - gross!) to do that, unfortunately. 

The question is: Is it easier to read highlighted, or all in one section? The answer to that question should be to merge or reject this PR. :ok_hand:

jamesls:
Given the two options, I think I'd prefer it as is without the syntax highlighting.  The intent there was just to showcase all the steps needed to get an app up and running in one block.  That code sample is more of an FYI that's explained in detail in the subsequent sections.

If it ever becomes possible to do different syntax highlighting in a single block, I'll all for it.

0atman:
Fair enough, it is more concise. Perhaps make the whole block python?

jamesls:
I tried that out, and the rendering for the shell parts (understandably) looks off.

I'm going to go ahead and close out this pull request, but thanks again for looking at this.

0atman:
Cool cool.

PR #477, opened 2017-08-16
Replace 'The' with 'This' to have sentence read easier.

bou2you:
I found this sentence easier to read if I replace 'The' with 'This'.

Do you agree?


jamesls:
Thanks for the pull request.  Seems reasonable, I might even update it to say "Chalice is a..." but let's start with this.

Issue #391, opened 2017-06-26
Browser-provided Accept header not accepted


twslankard:
Hi all,

I have a Chalice-based app that sends an HTML form that allows the user to download a file upon submission. The response content-type is `application/octet-stream`.

The browser (Firefox) sends the Accept header `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`, which contains the wildcard `*/*`. However, the response from the Chalice app is a JSON response:

    {
        "Message": "Request did not specify an Accept header with application/octet-stream, The response has a Content-Type of application/octet-stream. If a response has a binary Content-Type then the request must specify an Accept header that matches.",
        "Code": "BadRequest"
    }

I've also tried the request in Postman, setting the Accept header to the wildcard `*/*` explicitly, and I receive the same error. Only when I set the Accept header `application/octet-stream` does the request succeed.

I would expect the wildcard accept header (and the browser's default) to allow the request to succeed.

Thanks!

jamesls:
I've run into this as well, but I'm not sure if we have any options in chalice to make this work.  This is the behavior of API gateway.  From: http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html

> When a request contains multiple media types in its Accept header, API Gateway only honors the first Accept media type. In the situation where you cannot control the order of the Accept media types and the media type of your binary content is not the first in the list, you can add the first Accept media type in the binaryMediaTypes list of your API, API Gateway will return your content as binary. For example, to send a JPEG file using an <img> element in a browser, the browser might send Accept:image/webp,image/*,*/*;q=0.8 in a request. By adding image/webp to the binaryMediaTypes list, the endpoint will receive the JPEG file as binary.

Open to suggestions though.

twslankard:
Ah, dang! It's too bad API Gateway has this limitation, as it really reduces the number of possible use cases. Thanks for the info.

melevittfl:
Hi. I can get this to work with a non-chalice lambda function.

```python
header = { 'Content-type': 'image/png' }
    return {
           'statusCode' : 200,
          'headers': header,
          'body' : base64.b64encode(data).decode("utf-8"),
          'isBase64Encoded': True
     }
```

In the API Gateway, I configure the Settings page with a single Binary Types entry of '*/*' and it works. 

However, using Chalice and code like:
```python
return Response(status_code=200,
                    headers=header,
                    body=response_data
                    )
```
[...]

PR #478, opened 2017-08-16
Parse the accept header fully and return success if response matches …

corespace:
…any of accept

Let me explain why I'm making this change. I have an endpoint that I want to return an image like so:
```
@app.route('/sp', methods=['GET'], cors=cors_config)
def pixel():
    return Response(PNG_PIXEL_DATA, headers={
        'Content-Type': 'image/png'
    })
```
This will be embedded in a webpage like ``. However, a browser (like Chrome) won't send the Accept header "image/png", it will send something like `Accept:image/webp,image/apng,image/*,*/*;q=0.8`. Chalice doesn't call this a match and throws out a 400. This is my attempt to fix this so that `Accept: image/*` will match `Content-Type: image/png`. I know this PR has no tests, but I wanted to see if this was an acceptable solution before writing any tests. Please gimme feedback (especially if this behavior exists due to some API Gateway behavior as I've only tested this via chalice local)!

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -8,6 +8,7 @@ import traceback
 import decimal
 import base64
 from collections import defaultdict, Mapping
+from mime import AcceptableType
 
 __version__ = '1.0.0'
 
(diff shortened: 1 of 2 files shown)


corespace:
Dang. Was looking through closed issues and it turns out you have no choice here: https://github.com/aws/chalice/issues/391

SORRY! Closing.

PR #480, opened 2017-08-17
Add scripts for managing releases

jamesls:
The end goal is to generalize these to use for other repos,
but for now we're starting simple and adding it directly to chalice.


jamesls:
@stealthycoin  Consolidated into a single script.

PR #504, opened 2017-08-30
Fix environment variables being passed to subprocess while packaging

### LINKED ISSUE #501: Environment variables do not get passed to subprocess during package building
When Chalice creates subprocesses to build the package it does not pass down the current set of environment variables down. This was introduced in https://github.com/aws/chalice/commit/d6b22b3e8932dd588bf5110d5e5b084d246681a4 so it is a regression over the previous behavior where the subprocess inherited the environment. This matters for any setup.py file that makes use of an environment variable.

Initial issue #497

stealthycoin:
Fixes #501 
cc @kyleknap @jamesls @JordonPhillips @dstufft

### DIFF
--- a/chalice/utils.py
+++ b/chalice/utils.py
@@ -10,6 +10,7 @@ import tarfile
 import click
 from typing import IO, Dict, List, Any, Tuple, Iterator, BinaryIO  # noqa
 from typing import Optional  # noqa
+from typing import MutableMapping  # noqa
 
 from chalice.constants import WELCOME_PROMPT
 
@@ -77,6 +78,10 @@ def create_zip_file(source_dir, outfile):
 
 class OSUtils(object):
     ZIP_DEFLATED = zipfile.ZIP_DEFLATED
+
+    def environ(self):
+        # type: () -> MutableMapping
+        return os.environ
 
     def open(self, filename, mode):
         # type: (str, str) -> IO
--- a/tests/functional/test_package.py
+++ b/tests/functional/test_package.py
@@ -54,6 +54,7 @@ class FakePip(object):
         self._side_effects = defaultdict(lambda: [])
 
     def main(self, args, env_vars=None, shim=None):
+
         cmd, args = args[0], args[1:]
         self._calls[cmd].append((args, env_vars, shim))
         try:
@@ -162,9 +163,17 @@ def osutils():
 
 
 @pytest.fixture
-def pip_runner():
+def empty_env_osutils():
+    class EmptyEnv(object):
+        def environ(self):
+            return {}
+    return EmptyEnv()
+
+
+@pytest.fixture
+def pip_runner(empty_env_osutils):
     pip = FakePip()
-    pip_runner = PipRunner(pip)
+    pip_runner = PipRunner(pip, osutils=empty_env_osutils)
     return pip, pip_runner
 
 
--- a/tests/unit/deploy/test_packager.py
+++ b/tests/unit/deploy/test_packager.py
@@ -39,10 +39,29 @@ class FakePip(object):
 
 
 @pytest.fixture
-def pip_runner():
-    pip = FakePip()
-    pip_runner = PipRunner(pip)
-    return pip, pip_runner
+def pip_factory():
+    def create_pip_runner(osutils=None):
+        pip = FakePip()
+        pip_runner = PipRunner(pip, osutils=osutils)
+        return pip, pip_runner
+    return create_pip_runner
+
+
+class CustomEnv(OSUtils):
+    def __init__(self, env):
+        self._env = env
+
+    def environ(self):
+        return self._env
+
+
+@pytest.fixture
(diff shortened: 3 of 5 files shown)


chalice/utils.py
```diff
@@ -78,6 +78,10 @@ def create_zip_file(source_dir, outfile):
 class OSUtils(object):
     ZIP_DEFLATED = zipfile.ZIP_DEFLATED
 
+    def environ(self):
+        # type: () -> Dict[str,str]
+        return os.environ.copy()
```
kyleknap: Should we be caching the copy of the os.environ on the first call and then return the cached copy on subsequent calls? Mainly just wondering in the case that if we update the environement variables that gets returned, the update will not actually get returned again if `OSUtils.environ` gets called again.

tests/functional/test_package.py
```diff
@@ -162,9 +163,17 @@ def osutils():
 
 
 @pytest.fixture
-def pip_runner():
+def empty_env_osutils():
```
kyleknap: Does it makes sense to subclass from `OSUtils` for the `EmptyEnv` class?

tests/unit/deploy/test_packager.py
```diff
@@ -46,6 +46,26 @@ def pip_runner():
 
 
 @pytest.fixture
+def custom_env_osutils():
+    class CustomEnv(object):
```
kyleknap: Same thing here about subclassing from `OSUtils`.

tests/unit/deploy/test_packager.py
```diff
@@ -46,6 +46,26 @@ def pip_runner():
...
+            self._env = env
+
+        def environ(self):
+            return self._env
+    return CustomEnv
+
+
+@pytest.fixture
+def pip_runner_custom_env(custom_env_osutils):
+    def create_pip_runner(env):
```
kyleknap: So idea. What if instead of making a bunch of different pip runners, we just made a `pip_runner_factory` fixture? That way if we add anything more to the `PipRunner()` class, we do not have to add a new fixture for each of these.

tests/functional/test_package.py
stealthycoin: It won't really help or hurt anything in terms of the actual code, but it may make it clearer to someone reading the tests where this came from and what (partial) interface its trying to replace so I will update that.

chalice/utils.py
stealthycoin: Not needed for this particular use case but that is a reasonable expectation of someone consuming the `OSUtils` class in general so I will add it.

tests/unit/deploy/test_packager.py
stealthycoin: Yes I mainly didn't want to change the interface used in every single test.

tests/unit/deploy/test_packager.py
```diff
@@ -39,10 +39,29 @@ def calls(self):
...
+class CustomEnv(OSUtils):
+    def __init__(self, env):
+        self._env = env
+
+    def environ(self):
+        return self._env
+
+
+@pytest.fixture
+def pip_runner_custom_env(custom_env_osutils):
```
kyleknap: This can be removed now that it is no longer used.

chalice/utils.py
```diff
@@ -78,6 +79,10 @@ def create_zip_file(source_dir, outfile):
 class OSUtils(object):
     ZIP_DEFLATED = zipfile.ZIP_DEFLATED
 
+    def environ(self):
+        # type: () -> MutableMapping
+        return os.environ
```
kyleknap: Yeah that works too as long as we were not just returning back a copy each time.

tests/functional/test_package.py
```diff
@@ -162,9 +163,17 @@ def osutils():
...
+    class EmptyEnv(object):
+        def environ(self):
+            return {}
+    return EmptyEnv()
+
+
+@pytest.fixture
+def pip_runner(empty_env_osutils):
```
kyleknap: Do we need the `pip_runner` fixture anymore after the refactor?

jamesls: What's the alternative?  The `pip_runner` fixture is used in numerous tests still.  I'm not following.

kyleknap: I was referring to the fact that he switched out the use of `pip_runner` with `pip_factory`. I was not sure if the `pip_runner` fixture was being used anymore. When I did a quick check, I did not see it being used anywhere else in the `test_package.py` file

jamesls: Almost all the tests in `TestDependencyBuilder` in the class below use this.  Am I missing something?

kyleknap: Nevermind. I was looking at the `tests/unit/deploy/test_packager.py` file when I was going over the file. The `tests/functional/test_package.py` definitely needs it. Did not help that the same fixture was duplicated in two different test files instead of having it in the `conftest.py`.

PR #506, opened 2017-08-30
Set raw_body to None if no HTTP body was sent

### LINKED ISSUE #503: raw_body property inconsistency
When no body is sent, and `app.current_request.raw_body` is used, there is inconsistent behavior between local Chalice testing and actually how it runs in Lambda.

For local testing, if no body is sent `self._body` is getting set to a string: `{}`

However, when no body is provided in an actual call to API gateway, `self._body` is `None` and there is an unhandled AttributeError exception when this executes: `self._raw_body = self._body.encode('utf-8')` 

This effectively makes it hard for an app developer to gracefully handle a missing body from a call and will always yield a 500 error back to the user.

I personally see no reason to catch if body is None within `local.py` this effectively gives a false impression of how a request will be handled by Lambda itself. Instead the None type should be handled within the `raw_body` property IMO.

Relevant code in app.py:

```python
@property
    def raw_body(self):
        if self._raw_body is None:
            if self._is_base64_encoded:
                self._raw_body = self._base64decode(self._body)
            elif not isinstance(self._body, bytes):
                self._raw_body = self._body.encode('utf-8')
            else:
                self._raw_body = self._body
        return self._raw_body
```

Relevant code in local.py:
```python
def create_lambda_event(self, method, path, headers, body=None):
        # type: (str, str, Dict[str, str], str) -> EventType
        view_route = self._route_matcher.match_route(path)
        event = {
            'requestContext': {
                'httpMethod': method,
                'resourcePath': view_route.route,
                'identity': {
                    'sourceIp': self.LOCAL_SOURCE_IP
                },
            },
            'headers': dict(headers),
            'queryStringParameters': view_route.query_params,
            'pathParameters': view_route.captured,
            'stageVariables': {},
        }
        # if body is None:
        #    event['body'] = '{}'
        if self._is_binary(headers):
            event['body'] = base64.b64encode(body).decode('ascii')
            event['isBase64Encoded'] = True
        else:
            event['body'] = body
        return event
```

jamesls:
This also includes an update the local.py so we get
consistent behavior when no HTTP body is provided.

Fixes #503.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,9 @@ Next Release (TBD)
 * Fix issue where ``api_gateway_stage`` was being
   ignored when set in the ``config.json`` file
   (`#495 <https://github.com/aws/chalice/issues/495>`__)
+* Fix bug where ``raw_body`` would raise an exception if no HTTP
+  body was provided
+  (`#503 <https://github.com/aws/chalice/issues/503>`__)
 
 
 1.0.1
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -265,7 +265,7 @@ class Request(object):
         #: only be set if the Content-Type header is application/json,
         #: which is the default content type value in chalice.
         self._json_body = None
-        self._raw_body = None
+        self._raw_body = b''
         self.context = context
         self.stage_vars = stage_vars
 
@@ -277,7 +277,7 @@ class Request(object):
 
     @property
     def raw_body(self):
-        if self._raw_body is None and self._body is not None:
+        if not self._raw_body and self._body is not None:
             if self._is_base64_encoded:
                 self._raw_body = self._base64decode(self._body)
             elif not isinstance(self._body, bytes):
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -277,7 +277,7 @@ class Request(object):
 
     @property
     def raw_body(self):
-        if self._raw_body is None:
+        if self._raw_body is None and self._body is not None:
             if self._is_base64_encoded:
                 self._raw_body = self._base64decode(self._body)
             elif not isinstance(self._body, bytes):
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -294,7 +294,9 @@ class Request(object):
             return self._json_body
 
     def to_dict(self):
-        copied = self.__dict__.copy()
+        # Don't copy internal attributes.
+        copied = {k: v for k, v in self.__dict__.items()
+                  if not k.startswith('_')}
         # We want the output of `to_dict()` to be
         # JSON serializable, so we need to remove the CaseInsensitive dict.
         copied['headers'] = dict(copied['headers'])
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -99,9 +99,7 @@ class LambdaEventConverter(object):
             'pathParameters': view_route.captured,
             'stageVariables': {},
         }
-        if body is None:
-            event['body'] = '{}'
-        elif self._is_binary(headers):
+        if self._is_binary(headers) and body is not None:
             event['body'] = base64.b64encode(body).decode('ascii')
             event['isBase64Encoded'] = True
         else:
--- a/tests/integration/test_features.py
+++ b/tests/integration/test_features.py
@@ -410,7 +410,7 @@ def test_empty_raw_body(smoke_test_app):
     url = smoke_test_app.url + '/repr-raw-body'
     response = requests.post(url)
     response.raise_for_status()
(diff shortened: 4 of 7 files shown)


jamesls:
I don't have a strong preference either way.  The two things that swayed me were:

1. API Gateway set the `body` key to `null` if you don't provide a body so it's consistent with the event dict from Lambda.
2. The default (for the class `__init__`) has been `None` so this seemed to be a less risky change.  I suppose I could keep the default in the `__init__` as `None`, but then set the default to an empty string if I wanted in the body of the `__init__`, e.g. `if raw_body is None: raw_body = ''`

Do you think it should default to an empty string?

kyleknap:
I was not sure either, which is why I asked. The way I see it is returning a string is nice because you do not have to do any special casing for `None` in the logic.  However, is it possible to send an empty string as the body so that API Gateway set the body as an empty string in the lambda event? Just wondering if it is even possible to directly get an empty string. If so, it may be better to keep it as `None`. Also do you know other frameworks/libraries default to like flask? That may sway my preference.

jamesls:
> However, is it possible to send an empty string as the body so that API Gateway set the body as an empty string in the lambda event? 

I could not get API Gateway to send the empty string.   I think this is the default behavior of the lambda_proxy integration for API gateway

> Also do you know other frameworks/libraries default to like flask?

Flask sets it to the empty string (or more specifically an empty bytestring `b''`).  I'm thinking that might sway my preference as well.

kyleknap:
Given those reasons, I am leaning toward empty string or rather empty bytestring for the raw body. That is also assuming it won't break anyone.

jamesls:
Interesting...

The failures are due to the fact that `to_dict()` isn't able to be JSON serialized in python3 because the `Request` class now has a non json-serializable type (the empty byestring `b''`).

This is actually an independent problem separate from this change.  You can trigger this failure by trying to call `to_dict()` on the current request if you send any request body and access `app.current_request.raw_body` before calling `app.current_request.to_dict()`.

It was never my intent to map internal class attributes as part of the `app.current_request.to_dict()`, and I believe when the method was first written there were no internal attributes.  However, the JSON we generate now is:

```
{
    "_body": null,
    "_is_base64_encoded": false,
    "_json_body": null,
    "_raw_body": null,
    "context": {...},
    "headers": {...},
    "method": "POST",
    "query_params": null,
    "stage_vars": null,
    "uri_params": null
}
```

Which means my initial thoughts of just remove keys that start with `_` could break users.  I still think this is the right thing to do because:

1. The internal attributes will actually vary depending on what you access in the request object, i.e `.json_body` is computed lazily so whether or not this value is `null` depends on if you've accessed the property.
2. You can subtly break your existing app if you're using `to_dict()` and then add an access to `.raw_body` in your view function.

What do you all think?

cc @kyleknap @stealthycoin

kyleknap:
Agreed. I think it is fine to not include the `_` keys in the return value of `to_dict()`. Internal attributes in general should not be included. But I have a couple of follow up questions:

1) For the second breaking change: *You can subtly break your existing app if you're using to_dict() and then add an access to .raw_body in your view function.* Can you describe in code how that breaks? It seems strange that public properties would break based on calling a public get action.

2) Is it a contract that `to_dict()` will always be JSON serializable? If not, is it just recommended and should we make it part of the contract with this API?

jamesls:
1.  Here's an example app:

```python
@app.route('/')
def broken():
    app.current_request.raw_body
    return app.current_request.to_dict()
```

This app currently crashes.  If you omit the line `app.current_request.raw_body` then the app will not crash.  The reason for this is because in the chalice handler code, we construct the dict that lambda needs, so we have to serialize the body to JSON (by default in line 315):

https://github.com/aws/chalice/blob/f787f3e53cc346dbfa42c49331ee147a09451365/chalice/app.py#L312-L323

When you access `app.current_request.raw_body`, this value is also computed lazily in a property, so accessing `.raw_body` will internally cache this computed value as `self._raw_body`, and `self._raw_body` is of type `Optional[bytes]`:

https://github.com/aws/chalice/blob/f787f3e53cc346dbfa42c49331ee147a09451365/chalice/app.py#L278-L287

However if you don't access the `.raw_body` property, then `self._raw_body` is `None`, which serializes to `null` in JSON without issues.


2.  Yes: http://chalice.readthedocs.io/en/latest/api.html#Request.to_dict   .  We have some limited tests but nothing that checks all the different ways a request object can be constructed.  I think I may have found a legitimate use for hypothesis in chalice.

kyleknap:
Cool that makes sense to me. I will take a look at the code.

PR #507, opened 2017-08-31
Allow api_gateway_stage to be set from config.json

### LINKED ISSUE #495: API Gateway Stage ignored in config.json
Originally reported [here](https://github.com/aws/chalice/pull/475#issuecomment-324975657), I've confirmed this is an issue.

jamesls:
Anything that has chained precedence can't be defaulted at
the CLI layer because you can't differentiate between
a default value vs. not set.  In this case, `None` is used
as the sentinel to indicate not set.

The defaulting still happens in the CLI factory layer, it's just
that it needs to do in the "defaults" dict that is the final
fallback if no other value for a config key was found in the
various search dicts.

This is how `autogen_policy` is handled, so this change uses
that same approach but for `api_gateway_stage`.

Fixes #495

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+
+* Fix issue where ``api_gateway_stage`` was being
+  ignored when set in the ``config.json`` file
+  (`#495 <https://github.com/aws/chalice/issues/495>`__)
+
+
 1.0.1
 =====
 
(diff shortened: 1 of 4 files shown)


jamesls:
@kyleknap Yeah that was actually the type of test I was trying to write for this PR but the best compromise I could come up with was copying how the `deploy()` function calls `create_config_obj`.

I only spent a few minutes trying to think through how we could write those kinds of tests.  Let me spend a little more time thinking about this.  I do think this is ripe for future regressions if we don't put better testing in place.

If I can come up with anything I'll send it as a separate PR.

PR #508, opened 2017-08-31
Fix error code propagation from subprocess pip

### LINKED ISSUE #500: Dependency packaging does not propagate errors from subprocess
During the `pip download -r requirements.txt` call that Chalice makes during packaging it does not properly propagate the error code from the subprocess up to Chalice so it doesn't ever error out if there was a failure. 

Initial issue #497

stealthycoin:
Error codes were not returned which would result in pip failures not
cancelling the Chalice deploy process when they should. For example if
you provided a bad package name that pip failed to find the error would
be ignored and all the partial package would be deployed anyway.

Fixes #500

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -584,7 +584,7 @@ class SubprocessPip(object):
             shim = ''
         env_vars.update(subprocess_python_base_environ)
         python_exe = sys.executable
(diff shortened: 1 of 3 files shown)


kyleknap:
Also may want to make sure we put a changelog entry for this as well. Just noticed that when I was looking at one of James's PR's.

PR #527, opened 2017-09-07
Make integ tests more robust

jamesls:
1. Poll for resource id retrieval, it can take a while to propagate
2. Add missing `raise_for_status()` call to get better error messages on failure.

### DIFF
--- a/tests/integration/test_features.py
+++ b/tests/integration/test_features.py
@@ -20,6 +20,20 @@ APP_FILE = os.path.join(PROJECT_DIR, 'app.py')
 RANDOM_APP_NAME = 'smoketest-%s' % str(uuid.uuid4())
 
 
+def retry(max_attempts, delay):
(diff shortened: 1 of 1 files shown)


tests/integration/test_features.py
```diff
@@ -234,6 +227,23 @@ def test_path_params_mapped_in_api(smoke_test_app, apig_client):
     }
 
 
+def _poll_for_resource_id(apig_client, rest_api_id,
+                          max_attempts=10, delay=5):
```
stealthycoin: Sleep is hardcoded on the other one of these you recently submitted. Make them consistent? Make a generic poller for use in the tests? Seems like we may be adding more of these.

jamesls: Pushed an update that refactored the common logic.  For now I'd prefer to keep the constants local to each retry call until we see a more generalized pattern for what these values should be.

PR #532, opened 2017-09-11
Add hypothesis tests

jamesls:
This adds initial support for hypothesis tests.  I've seeded this
with a couple of tests for app.py.  In addition I've added
configuration such that more iterations of these tests are run on
travis.

Originally mentioned in https://github.com/aws/chalice/pull/506#issuecomment-326417807.


jamesls:
Capturing offline discussion.  It seems most projects don't separate property tests from their other tests, and the only reason I've done it this way in other projects is because I didn't also have to handle functional, integration tests etc.

So I'll:

* Add this into `tests/unit/test_app.py`
* Set the default iterations to really low so we still get fast unit tests time.
* Keep the high iteration count on the CI profile.

jamesls:
@kyleknap ready to look at again.

PR #540, opened 2017-09-19
Fix issue with local mode bodiless responses causing clients to hang

### LINKED ISSUE #525: OPTIONS generated by CORS doesn't work correctly in local
First of all, thank you for this awesome framework. Really easy to use, I like it a lot.

I'm developing an API that has to accept CORS and i'm having troubles with the OPTIONS methods.

The problem is easy to reproduce:
In app.py
```
...
@app.route('/test', methods=['POST'], cors=True)
def test_cors():
  return {'hello': 'world'}

```
Then, run it in local: `chalice local --port=8100`

When you try to do the first OPTIONS curl, everything works =)

```
curl -H "Origin: http://example.com" \                                                      
  -H "Access-Control-Request-Method: POST" \
  -H "Access-Control-Request-Headers: X-Requested-With" \
  -X OPTIONS --verbose \
  http://localhost:8100/test
```
Response:
```
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8100 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8100 (#0)
> OPTIONS /test HTTP/1.1
> Host: localhost:8100
> User-Agent: curl/7.51.0
> Accept: */*
> Origin: http://example.com
> Access-Control-Request-Method: POST
> Access-Control-Request-Headers: X-Requested-With
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Server: BaseHTTP/0.6 Python/3.6.0
< Date: Thu, 07 Sep 2017 16:02:02 GMT
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Headers: Authorization,Content-Type,X-Amz-Date,X-Amz-Security-Token,X-Api-Key
< Access-Control-Allow-Methods: POST,OPTIONS
<
* Curl_http_done: called premature == 0
* Closing connection 0
```

But the next OPTIONS curl will not end never =(

```
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8100 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8100 (#0)
> OPTIONS /test HTTP/1.1
> Host: localhost:8100
> User-Agent: curl/7.51.0
> Accept: */*
> Origin: http://example.com
> Access-Control-Request-Method: POST
> Access-Control-Request-Headers: X-Requested-With
>
< HTTP/1.1 200 OK
< Server: BaseHTTP/0.6 Python/3.6.0
< Date: Thu, 07 Sep 2017 16:02:03 GMT
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Headers: Authorization,Content-Type,X-Amz-Date,X-Amz-Security-Token,X-Api-Key
< Access-Control-Allow-Methods: POST,OPTIONS
* no chunk, no close, no size. Assume close to signal end
```

It stops in  `* no chunk, no close, no size. Assume close to signal end`. It seems like the server only send the close signal only in the first call. If you stop the second curl and try it again, it wont stop neither.
 
Do you have the same problem? Can you reproduce it?

P.S.: I'm using macOS, but I hope that is not a problem to do the local development =P

stealthycoin:
fixes #525

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+
+* Fix issue with chalice local mode causing http clients to hang on responses
+  with no body
+  (`#525 <https://github.com/aws/chalice/issues/525>`__)
+
+
 1.0.2
 =====
 
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -574,6 +574,7 @@ class ChaliceRequestHandler(BaseHTTPRequestHandler):
 
     def _send_http_response_no_body(self, code, headers):
         # type: (int, HeaderType) -> None
+        headers['Content-Length'] = '0'
         self.send_response(code)
         for k, v in headers.items():
             self.send_header(k, v)
--- a/tests/integration/test_local.py
+++ b/tests/integration/test_local.py
@@ -1,0 +1,112 @@
+import socket
+import contextlib
+from threading import Thread
+from threading import Event
+
+import pytest
+import requests
+
+from chalice import app
+from chalice.local import LocalDevServer
+from chalice.config import Config
+
+
+class ThreadedLocalServer(Thread):
+    def __init__(self, port):
+        super(ThreadedLocalServer, self).__init__()
+        self._app_object = None
+        self._config = None
+        self._port = port
+        self._server = None
+        self._server_ready = Event()
+
+    def configure(self, app_object, config):
+        self._app_object = app_object
+        self._config = config
+
+    def run(self):
+        self._server = LocalDevServer(
+            self._app_object, self._config, self._port)
+        self._server_ready.set()
+        self._server.serve_forever()
+
+    def make_call(self, method, path, port, timeout=0.5):
+        self._server_ready.wait()
+        return method('http://localhost:{port}{path}'.format(
+            path=path, port=port), timeout=timeout)
+
+    def shutdown(self):
+        if self._server is not None:
+            self._server.server.shutdown()
+
+
+@pytest.fixture
+def config():
+    return Config()
+
+
+@pytest.fixture()
+def unused_tcp_port():
+    with contextlib.closing(socket.socket()) as sock:
+        sock.bind(('127.0.0.1', 0))
+        return sock.getsockname()[1]
+
+
+@pytest.fixture()
+def local_server_factory(unused_tcp_port):
+    threaded_server = ThreadedLocalServer(unused_tcp_port)
+
+    def create_server(app_object, config):
+        threaded_server.configure(app_object, config)
+        threaded_server.start()
+        return threaded_server, unused_tcp_port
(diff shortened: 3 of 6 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,13 @@
 CHANGELOG
 =========
 
+Next Release (TBD)
+==================
+* Fix issue with chalice local mode causing http clients to hang on responses
```
jamesls: Line between header and changelog entry.

tests/integration/test_local.py
```diff
@@ -0,0 +1,93 @@
+import socket
```
jamesls: Hmm, I'm surprised flake8 didn't complain about this.  Please use pep8 guidelines for grouping imports.

chalice/local.py
```diff
@@ -201,7 +201,7 @@ class NotAuthorizedError(LocalGatewayException):
 
 
 class NoOptionsRouteDefined(LocalGatewayException):
-    CODE = 403
+    CODE = 200
```
jamesls: Why was this changed?

stealthycoin: Because thats what API gateway does. I missed it before but I noticed it when going over these the tests for cors cases. I can do this as a separate PR if you would prefer.

jamesls: I don't see a test for it.  I don't mind it in this PR, but we just have a test for this.  Might also be good to have a separate commit for this with a description of why it's being changed.

stealthycoin: The existing tests should have covered it and I corrected the tests while I was there anyway and thats when I noticed it was incorrect.

I added these assertions:
* https://github.com/stealthycoin/chalice/blob/52f48152c7d2206c6fe9fe62cc39cf243f52e9fa/tests/unit/test_local.py#L348

* https://github.com/stealthycoin/chalice/blob/52f48152c7d2206c6fe9fe62cc39cf243f52e9fa/tests/unit/test_local.py#L364

jamesls: I don't follow.  This is updating the code for an exception right?  Those tests are for successful requests aren't they?  Are you saying we were previously returning a 403 response when we should have been returning a 200?

stealthycoin: The exception is raised when there is no OPTIONS method defined on a route which LocalGateway exceptions are raised when the routes cannot process the request, so that is reasonable so far. The calling code can still generate the OPTIONS request that API Gateway would return and then send a successful response if there is a CORSConfig object on it. I missed that it was a success case somehow before. But looking at it again it seems like it should be decoupled entirely from that type of error since all the others exceptions like this are actual error cases. Whereas this one should be caught and turned into a successful response by auto-generating the CORS headers.

I will fix this in a separate PR.

Issue #545, opened 2017-09-20
Add --stage Parameter to Local Mode


jpeters71:
It would be really handy to be able to specify the stage parameter when running 'chalice local'.  This would allow developers the opportunity to have different local settings than development settings.  A classic use case would be to run against a local DB rather than the shared development DB out in AWS.

stealthycoin:
Seems like a reasonable feature request to me.

PR #546, opened 2017-09-20
Added stage parameter to the local server.

jpeters71:
This is a simple change to resolve https://github.com/aws/chalice/issues/545

Having the ability to have local specific settings would be very useful.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -77,7 +77,7 @@ def cli(ctx, project_dir, debug=False):
 @cli.command()
 @click.option('--port', default=8000, type=click.INT)
 @click.option('--stage', default=DEFAULT_STAGE_NAME,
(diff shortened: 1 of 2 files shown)


chalice/cli/__init__.py
```diff
@@ -76,18 +76,22 @@ def cli(ctx, project_dir, debug=False):
 
 @cli.command()
 @click.option('--port', default=8000, type=click.INT)
+@click.option('--stage', default=DEFAULT_STAGE_NAME,
+              help=('Name of the Chalice stage to deploy to. '
```
jamesls: Could you update the docstring here to just say that this is the stage configuration we'll use for local mode?

jpeters71: Sure, just finished changing it.  Good catch.

Issue #516, opened 2017-09-05
[proposal] AWS Managed Resouces for Chalice


kyleknap:
# Abstract

This issue proposes a mechanism for managing additional AWS resources
that may be relied on in a serverless Chalice application. In terms of
management, it will handle creates, updates, and deletion of the resources
upon deployment of the application and allow users to easily interact
with these resources within their application. The only AWS resouce that
will be added in this proposal is a DynamoDB table, but the mechanism should
be able to support any future AWS resource.

# Motivation

Currently, Chalice does no management of any AWS resources that are part of
the core buisness logic of a Chalice application but are not part of
the Chalice decorator interface (i.e. `app.lambda_function()`, `app.route()`,
etc.). There are a lot of AWS resources an application may rely on such as
an S3 bucket or a DynamoDB Table. However, these resources must be created
out of band of the actual Chalice deployment process which is inconvenient
because:

* There is no way to hook into the `chalice deploy` command. So users have
  to deploy the resources themself outside of `chalice deploy` by manually
  deploying the resources, relying on out of band deployment scripts, etc.
* If users are using CloudFormation and `chalice package`, they will need to
  modify the CloudFormation template to add the resources you need.
* Users will need to manage the deployment of resources per stage if they are
  looking for resources in each of these stages to be purely isolated from
  the other stages.
* Within the core logic of the Chalice application, there is nothing
  guaranteeing that the AWS resource you are trying to access exists. The
  user needs to be extra careful to make sure the out of band resources they
  deploy match up with what is used in the Chalice application.

Therefore, it is a much more friendly user story if Chalice handles the
deployment of these resources for the user. Also since Chalice did the
deployment, it can easily provide references to those deployed resources
from within the Chalice application.

# Specification

This section will go into detail about the interfaces for adding these
managed resources, code samples of how users will interact with the interface,
and the deployment logic in deploying these managed resources. Since
DynamoDB tables is the only resource this proposal is suggesting to add,
this section will be specific to DynamoDB tables.

To have Chalice manage AWS resources, users will first have to declare
their resources in code via `resources.py` file and then may configure these
resources using the Chalice config file.

## Code Interface

The top level interface into these managed resources is a `resources.py`
file. This is used for declaring all additional AWS resources to be managed
by managed by Chalice in an application. The `resources.py` file will live
alongside the `app.py` file in a Chalice application:

```
myapp$ tree .
.
|-- app.py
|-- requirements.txt
|-- resources.py
```

[...]

PR #549, opened 2017-09-20
Initial POC of new deployer

jamesls:
The existing deployer module is starting to get out of hand.
Initially it only had to handle API gateway and a single Lambda
function, but over time we've added support for AWS resources,
and continue to add additional resources (#516).

I don't think the existing deployer design is going to hold up
as all these new resources are being added.  The code is hard to
understand, hard to use, and hard to modify.  There's also a number
of customer reported bugs from the existing deployer.

Digging further into the deployer code, there are a number of
inconsistencies as well as unexpected and unintuitive behaviors.

I've been experimenting with a new redesign of the deployer to
not only fix the existing issues, but also make future changes
easier.  This initial POC only supports `@app.lambda_function()`
and only sketches out how to do the initial creation of resources,
but I think this should be able to handle the remaining gaps.

To help with initial testing, I've added a temporary `deploy-new`
command so both deployers can live side by side during development.
Before this gets merged to master, `deploy-new` will just become
`deploy` and the old code will be removed.

The `newdeployer` module docstring has a description of how
the new system works, and is a good starting point for a technical
overview.

### DIFF
--- a/chalice/deploy/models.pyi
+++ b/chalice/deploy/models.pyi
@@ -1,0 +1,124 @@
+from typing import List, Dict, Any, TypeVar, Union
+import enum
+
+class DeployPhase(enum.Enum):
+    BUILD = 'build'
+    DEPLOY = 'deploy'
+
+
+T = TypeVar('T')
+DV = Union[DeployPhase, T]
+STR_MAP = Dict[str, str]
+
+
+class Model:
+    def dependencies(self) -> List[Model]: ...
+
+
+class ManagedModel(Model):
+    resource_name = ...  # type: str
+    resource_type = ...  # type: str
+
+
+class Application(Model):
+    stage = ...  # type: str
+    resources = ... # type: List[Model]
+
+    def __init__(self,
+                 stage,       # type: str
+                 resources,   # type: List[Model]
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class DeploymentPackage(Model):
+    filename = ...  # type: str
+
+    def __init__(self,
+                 filename,   # type: DV[str]
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class IAMPolicy(Model):
+    pass
+
+
+class FileBasedIAMPolicy(IAMPolicy):
+    filename = ...  # type: str
+
+    def __init__(self,
+                 filename,   # type: str
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class AutoGenIAMPolicy(IAMPolicy):
+    document = ...  # type: DV[Dict[str, Any]]
+
+    def __init__(self,
+                 document,   # type: DV[Dict[str, Any]]
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class IAMRole(Model):
+    pass
+
+
+class PreCreatedIAMRole(IAMRole):
+    role_arn = ... # type: str
+
+    def __init__(self,
+                 role_arn,   # type: str
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class ManagedIAMRole(IAMRole, ManagedModel):
+    role_arn = ... # type: DV[str]
+    role_name = ... # type: str
+    trust_policy = ... # type: Dict[str, Any]
+    policy = ... # type: IAMPolicy
+
+    def __init__(self,
+                 resource_name,  # type: str
+                 role_arn,       # type: DV[str]
+                 role_name,      # type: str
+                 trust_policy,   # type: Dict[str, Any]
+                 policy,         # type: IAMPolicy
+                 ):
+        # type: (...) -> None
+        ...
+
+
+class LambdaFunction(ManagedModel):
+    function_name = ... # type: str
+    deployment_package = ... # type: DeploymentPackage
+    environment_variables = ... # type: STR_MAP
+    runtime = ... # type: str
+    handler = ... # type: str
+    tags = ... # type: STR_MAP
+    timeout = ... # type: int
+    memory_size = ... # type: int
+    role = ... # type: IAMRole
+
+    def __init__(self,
+                 resource_name,           # type: str
+                 function_name,           # type: str
+                 deployment_package,      # type: DeploymentPackage
+                 environment_variables,   # type: STR_MAP
+                 runtime,                 # type: str
+                 handler,                 # type: str
+                 tags,                    # type: STR_MAP
+                 timeout,                 # type: int
+                 memory_size,             # type: int
+                 role,                    # type: IAMRole
+                 ):
+        # type: (...) -> None
+        ...
--- a/chalice/deploy/models.pyi
+++ b/chalice/deploy/models.pyi
@@ -1,13 +1,13 @@
 from typing import List, Dict, Any, TypeVar, Union
 import enum
 
-class DeployPhase(enum.Enum):
-    BUILD = 'build'
-    DEPLOY = 'deploy'
+class Placeholder(enum.Enum):
+    BUILD_STAGE = 'build_stage'
+    DEPLOY_STAGE = 'deploy_stage'
 
 
 T = TypeVar('T')
-DV = Union[DeployPhase, T]
+DV = Union[Placeholder, T]
 STR_MAP = Dict[str, str]
 
 
--- a/chalice/deploy/newdeployer.py
+++ b/chalice/deploy/newdeployer.py
@@ -36,18 +36,19 @@ This takes the ordered list of resources and allows any local build processes
 to occur.  The rule of thumb here is no remote AWS calls.  This stage includes
 auto policy generation, pip packaging, injecting default values, etc.  To
 clarify which attributes are affected by the build stage, they'll usually have
-a value of ``models.DeployPhase.BUILD``.  Processors in the build stage will
-replaced those ``models.DeployPhase.BUILD`` values with whatever the "built"
-value is (e.g the filename of the zipped deployment package).
+a value of ``models.Placeholder.BUILD_STAGE``.  Processors in the build stage
+will replaced those ``models.Placeholder.BUILD_STAGE`` values with whatever the
+"built" value is (e.g the filename of the zipped deployment package).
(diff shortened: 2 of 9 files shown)


tests/unit/deploy/test_newdeployer.py
```diff
@@ -0,0 +1,692 @@
...
+        generator.generate_policy.return_value = {'policy': 'doc'}
+        policy = models.AutoGenIAMPolicy(models.DeployPhase.BUILD)
+        config = Config.create()
+
+        p = PolicyGenerator(generator)
+        p.handle(config, policy)
+
+        assert policy.document == {'policy': 'doc'}
+
+    def test_no_policy_generated_if_exists(self):
```
stealthycoin: Maybe something more descriptive than `exists`, like `_if_xxx_populated` which was used in a test below.

chalice/deploy/models.pyi
```diff
@@ -0,0 +1,124 @@
+from typing import List, Dict, Any, TypeVar, Union
```
stealthycoin: This file looks really similar to the `models.py`, it would be nice if this was easier to maintain.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+from chalice.deploy.packager import PipRunner, SubprocessPip
+from chalice.deploy.packager import DependencyBuilder as PipDependencyBuilder
+from chalice.policy import AppPolicyGenerator
+from chalice.constants import LAMBDA_TRUST_POLICY
+from chalice.constants import DEFAULT_LAMBDA_TIMEOUT
+from chalice.constants import DEFAULT_LAMBDA_MEMORY_SIZE
+from chalice.awsclient import TypedAWSClient, ResourceDoesNotExistError
+
+
+def create_default_deployer(session):
```
stealthycoin: Capturing offline discussion about changing this to use a builder pattern instead.

jamesls:
Just a heads up, I've pushed an upstream branch `new-deployer` that's branched off of master.  I've also updated the base of this PR to be `upstream/new-deployer`.  The idea being as features are added we can use `upstream/new-deployer` to aggregate all the changes and once we have parity we can merge this over to master.

chalice/deploy/newdeployer.py
jamesls: Yep, though I'm going to hold off on that until we need it for deletes.  For now a static factory function is sufficient and simpler to follow.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+        # type: (Config, str) -> Dict[str, Any]
+        application = self._resource_builder.build(config, chalice_stage_name)
+        resources = self._deps_builder.build_dependencies(application)
+        self._build_stage.execute(config, resources)
+        plan = self._plan_stage.execute(config, resources)
+        self._executor.execute(plan)
+        return {'resources': self._executor.resources}
+
+
+class ResourceGraphBuilder(object):
```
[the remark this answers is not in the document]
jamesls: I like `ApplicationGraphBuilder`, I can change it to that.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+             'role_arn': role_arn,
+             'zip_contents': self._osutils.get_file_contents(
+                 resource.deployment_package.filename, binary=True),
+             'runtime': resource.runtime,
+             'handler': resource.handler,
+             'environment_variables': resource.environment_variables,
+             'tags': resource.tags,
+             'timeout': resource.timeout,
+             'memory_size': resource.memory_size},
+            '%s_arn' % resource.resource_name,
```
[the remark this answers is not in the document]
jamesls: I agree I'll update that.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+                resource.role_arn = role_arn
+            except ResourceDoesNotExistError:
+                document = None
+                if isinstance(resource.policy, models.AutoGenIAMPolicy):
+                    document = resource.policy.document
+                return APICall('create_role',
+                               {'name': resource.role_name,
+                                'trust_policy': resource.trust_policy,
+                                'policy': document},
+                               '%s_role_arn' % resource.role_name,
```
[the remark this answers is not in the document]
jamesls: Do you mean have the `APICall` accept `**kwargs`?  We' have to be careful about shadowing the existing `APICall` args as well which is dependent on the method signature of `TypedAWSClient` so I'd prefer to keep it as is.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+        # are made.  These can be used in subsequent API calls.
+        self.variables = {}  # type: Dict[str, Any]
+        self.resources = {}  # type: Dict[str, Dict[str, Any]]
+
+    def execute(self, api_calls):
+        # type: (List[APICall]) -> None
+        for api_call in api_calls:
+            final_kwargs = self._resolve_variables(api_call.params)
+            method = getattr(self._client, api_call.method_name)
+            result = method(**final_kwargs)
```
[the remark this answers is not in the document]
jamesls: Yeah I'll make a TODO about that.  I imagine the logic will need to change a bit.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+                    )
+                    mapping[varname] = result
+
+    def _resolve_variables(self, params):
+        # type: (Dict[str, Any]) -> Dict[str, Any]
+        final = {}
+        for key, value in params.items():
+            if isinstance(value, Variable):
+                final[key] = self.variables[value.name]
+            else:
```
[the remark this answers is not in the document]
jamesls: Good point I'll get that updated.

chalice/deploy/newdeployer.py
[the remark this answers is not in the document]
stealthycoin: I just meant explicitly write out `output='%s_role_arn % resource.role_name` as an argument to `APICall` so I can see what it is without referencing `APICall`. For ease of reading.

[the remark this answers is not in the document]
jamesls: oh ok yeah I can do that.

chalice/deploy/models.pyi
```diff
@@ -0,0 +1,124 @@
...
+import enum
+
+class DeployPhase(enum.Enum):
```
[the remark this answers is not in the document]
jamesls: What just called it a placeholder? so:

```python
class Placeholder(enum.ENUM):
    BUILD_STAGE = 'build_stage'
    DEPLOY_STAGE = 'deploy_stage'
```

We could also add a `PLAN_STAGE` if we ever need place holder values that filled in during the plan stage.

jamesls:
Ok I believe I've incorporated all the feedback so far.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,472 @@
...
+        msg = self.MSG % (key, value, method_name)
+        super(UnresolvedValueError, self).__init__(msg)
+        self.key = key
+        self.value = value
+        self.method_name = method_name
+
+
+class Deployer(object):
+    def __init__(self,
+                 resource_builder,  # type: ApplicationGraphBuilder
```
stealthycoin: Change this name to match its new type more closely? resource_builder seems pretty vague to me.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,472 @@
...
+        # type: (...) -> None
+        self._resource_builder = resource_builder
+        self._deps_builder = deps_builder
+        self._build_stage = build_stage
+        self._plan_stage = plan_stage
+        self._executor = executor
+
+    def deploy(self, config, chalice_stage_name):
+        # type: (Config, str) -> Dict[str, Any]
+        application = self._resource_builder.build(config, chalice_stage_name)
```
stealthycoin: application_graph? I still think of a chalice app object when I see application.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,472 @@
...
+            else:
+                filename = os.path.join(config.project_dir,
+                                        '.chalice',
+                                        'policy-%s.json' % stage_name)
+            policy = models.FileBasedIAMPolicy(filename=filename)
+        else:
+            resource_name = 'default-role'
+            role_name = '%s-%s' % (config.app_name, stage_name)
+            policy = models.AutoGenIAMPolicy(
+                document=models.Placeholder.BUILD_STAGE)
```
stealthycoin: Yea I think its a lot clearer this way than it was before.

chalice/deploy/newdeployer.py
```diff
@@ -0,0 +1,472 @@
...
+            ordered.append(resource)
+
+
+class BaseDeployStep(object):
+
+    def handle(self, config, resource):
+        # type: (Config, models.Model) -> None
+        name = 'handle_%s' % resource.__class__.__name__.lower()
+        handler = getattr(self, name, None)
+        if handler is not None:
```
stealthycoin: If handler is `None` do we want it to silently do nothing? Is there any case where it should be `None`?

jamesls: Being `None` is ok, it just means that there's no action associated with that resource for whatever step we're processing.  For example, not every resource is going to have default values injected.

chalice/deploy/newdeployer.py
jamesls: Updated.

chalice/deploy/newdeployer.py
jamesls: Eh, I mean the type of the object *is* a `models.Application`, and the methods do make it clear the type it returns (and accepts).  Also the chalice app is a type `Chalice` not `application`.  I don't see this being confusing given the required type hints and the fact that this is all in the deployer module.  Technically all of these model objects are graphs because they all have a `dependencies()` method.

PR #550, opened 2017-09-20
Stop infinite recursion in analyzer

### LINKED ISSUE #531: Maximum recursion depth exceeded when running chalice deploy
Hey all,

I'm working on deployment for a my new Chalice-based project (using version 1.0.2.)

When I run `chalice deploy` I get the error `maximum recursion depth exceeded` with no stack trace. After running the debugger, I was able to narrow down the problem to Chalice's AST traversal code. From there, I created a reduction using my `app.py`. Given this, it appears as though the AST traversal code does not tolerate recursion.

Here's the reduced `app.py`. If I remove the recursive function, then `chalice deploy` appears to work as expected.

```
from chalice import Chalice

app = Chalice(app_name='recursion')
app.debug = True


def recursive_function(obj):
    if False:
        recursive_function(obj)


@app.route('/')
def index():
    recursive_function({})
    return {}
```

joguSD:
Fixes #531 

The analyzer would follow any recursive function infinitely. This adds a visited set of AST nodes to the AST walker so that we don't attempt to jump into a function definition multiple times.

### DIFF
--- a/tests/unit/test_analyzer.py
+++ b/tests/unit/test_analyzer.py
@@ -61,6 +61,50 @@ def test_inferred_module_type():
     """) == {'boto3': Boto3ModuleType()}
 
 
+def test_recursive_function_none():
+    assert aws_calls("""\
+        def recursive_function():
+            recursive_function()
+        recursive_function()
+    """) == {}
+
+
+def test_recursive_comprehension_none():
+    assert aws_calls("""\
+        xs = []
+        def recursive_function():
+            [recursive_function() for x in xs]
+        recursive_function()
+    """) == {}
+
+
+def test_recursive_function_client_calls():
+    assert aws_calls("""\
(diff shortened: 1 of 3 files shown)


tests/unit/test_analyzer.py
```diff
@@ -61,6 +61,50 @@ def test_inferred_module_type():
     """) == {'boto3': Boto3ModuleType()}
 
 
+def test_recursive_function_none():
```
stealthycoin: what does the test name ending in `_none` mean?

joguSD: That there are no botocore client calls.

jamesls:
Looks travis is failing because the type signatures didn't get updated.

joguSD:
Yeah, I just saw that working on updating them now.

PR #554, opened 2017-09-25
Chalice local 200s CORS OPTIONS rather than 403

corespace:
Hi folks,

My local preflights got broken, so this should fix them. Ran coverage reports before/after and no changes there. You could probably delete `class NoOptionsRouteDefined`, but I didn't want to make that call myself.

Thanks!
-Kyle


stealthycoin:
@corespace Thanks for the submission! Could you also throw out `NoOptionsRouteDefined` since it is no longer being used?

corespace:
Waaaa? Why you close this?

PR #555, opened 2017-09-26
Fix wheel packages with _ in their name not being converted to -

stealthycoin:
Since wheel files use - as a delimeter between the different
components of the filename such as distribution-version-... if the
distribution has - in it they are converted to _. So when we read a
wheel file that has _ in the distribution, they need to be remaped to
-.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -469,6 +469,11 @@ class Package(object):
         # The directory format is {distribution}-{version}.data
         return '%s-%s.data' % (self._name, self._version)
 
+    def _normalize_name(self, name):
+        # type: (str) -> str
+        # Taken directly from PEP 503
+        return re.sub(r"[-_.]+", "-", name).lower()
+
     @property
     def identifier(self):
         # type: () -> str
@@ -499,7 +504,7 @@ class Package(object):
             # {distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-
             # {platform tag}.whl
             name, version = self.filename.split('-')[:2]
-            name = name.replace('_', '-')
+            name = self._normalize_name(name)
         else:
             info_fetcher = SDistMetadataFetcher(osutils=self._osutils)
             sdist_path = self._osutils.joinpath(self._directory, self.filename)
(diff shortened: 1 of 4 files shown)


jamesls:
Can you update your commit message with specific packages that have this issue?  It'd be great just to have some real world data here.

chalice/deploy/packager.py
```diff
@@ -499,6 +499,7 @@ def _calculate_name_and_version(self):
             # {distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-
             # {platform tag}.whl
             name, version = self.filename.split('-')[:2]
+            name = name.replace('_', '-')
```
dstufft: It might be smart to use the PEP 503 normalization rules here, or at least in the ``__eq__()`` implementation.

stealthycoin: Ill update the commit and look at PEP 503

tests/integration/test_package.py
```diff
@@ -39,6 +39,22 @@ def _get_random_package_name():
...
 
 class TestPackage(object):
+    def test_can_package_with_dashes_in_name(self, runner, app_skeleton):
+        req = os.path.join(app_skeleton, 'requirements.txt')
+        package = 'googleapis-common-protos==1.5.2'
+        with open(req, 'w') as f:
+            f.write('%s\n' % package)
+        cli_factory = factory.CLIFactory(app_skeleton)
+        package_output_location = os.path.join(app_skeleton, 'pkg')
+        print(package_output_location)
```
jamesls: Remove print statement

tests/integration/test_package.py
```diff
@@ -39,6 +39,22 @@ def _get_random_package_name():
...
+        result = runner.invoke(
+            cli.package, [package_output_location],
+            obj={'project_dir': app_skeleton,
+                 'debug': False,
+                 'factory': cli_factory})
+        assert result.exit_code == 0
+        assert result.output.strip() == 'Creating deployment package.'
```
jamesls: Can we actually assert that the package dir looks correct?  It doesn't have to be detailed, but this test would pass with code that literally did `print "Creating deployment pakage."`

PR #556, opened 2017-09-26
Adding new line for saved JSON files

artkay:
### The problem

Every time when I deploy with chalice it rewrites the JSON documents (the policy and the `deployed.json`) and there is no new line at the end of these files. I am committing these files to the repos and GitHub is complaining about this:

screen shot 2017-09-26 at 4 14 24 pm

So for now I just add the new lines manually because a PR with files with incorrect last byte will fail PR review at the company I am working for.

### Proposed solution

Simple add new line to the policy writer and to the `deployed.json` writer. 


jamesls:
Thanks for the pull request.  I pulled this in and added a little more refactoring + tests in #557.

PR #560, opened 2017-10-02
Add file encoding argument to the get_file_contents

stealthycoin:
Specifically using the default causes an issue on windows if any file
has contents that are not readable by cp1252. This comes up during the
packaging of a deployment. Some egg-info files contain unicode
characters. Windows fails to build the deployment bundle since it
crashes trying to read the egg-info file.

@jamesls @kyleknap

### DIFF
--- a/tests/functional/test_utils.py
+++ b/tests/functional/test_utils.py
@@ -2,7 +2,14 @@ import zipfile
 import json
 import os
 
+import pytest
+
 from chalice import utils
+
+
+@pytest.fixture
+def osutils():
+    return utils.OSUtils()
 
 
 def test_can_zip_single_file(tmpdir):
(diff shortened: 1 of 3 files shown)


tests/functional/test_utils.py
```diff
@@ -98,3 +105,14 @@ def test_remove_stage_from_deployed_values_no_file(tmpdir):
 
     # Make sure it doesn't create the file if it didn't already exist
     assert not os.path.isfile(filename)
+
+
+class TestOSUtils(object):
+    def test_can_read_unicode(self, tmpdir, osutils):
+        filename = tmpdir.join('file.txt')
+        checkmark = u'\2713'
+        with open(filename, 'w', encoding='utf-8') as f:
```
jamesls: Can you change this test to not use the default value as the encoding to test?

jamesls:
Also looks like the travis builds have failed.  Marked as changes-requested,

jamesls:
Hey looks like the build is still failing.  Would you mind running `make prcheck` before asking for another review?

PR #562, opened 2017-10-02
Check comprehension generators in the analyzer

joguSD:
This implements more closely the handle_comprehension logic present in the CPython [source code](https://github.com/python/cpython/blob/20cbc1d29facead32c2da06c81a09af0e057015c/Python/symtable.c#L1496-L1531)

This allows us to handle dictionary comprehensions as well as explore the generators.

Also present is a change to how we handle ambiguous child scope. Previously on Python 2, if two comprehensions of the same type (excluding List) were present in the same scope we would fail to disambiguate them by line number as Python 2 always passes 0 for these as can be seen [here](https://github.com/python/cpython/blob/20cbc1d29facead32c2da06c81a09af0e057015c/Python/symtable.c#L1509). From my research it isn't possible to perform this disambiguation perfectly in Python 2 as the appropriate C bindings are not exposed. Specifically, the AST used to create the symbol table is kept in an arena within the C code and those objects are never exposed to Python. Because of this we cannot use the `get_id` method on the child symbol table to match the `id(ast_node)` as the AST nodes no longer exist. [This](https://github.com/python/cpython/blob/dae0276bb6bc7281d59fb0b8f1aab31634ee80dc/Python/pythonrun.c#L1074-L1095) function in the CPython source is what takes our source code string and produces the symbol table.

This means we will fail to understand references to variables that are introduced in the comprehension's child scope (via a KeyError exception), however this is not really a big issue as the analyzer currently cannot infer the type of variables introduced in the comprehension scope as that would require understanding what type is coming out of the iterator present in the generator.

### DIFF
--- a/chalice/analyzer.py
+++ b/chalice/analyzer.py
@@ -567,21 +567,26 @@ class SymbolTableTypeInfer(ast.NodeVisitor):
         # with the name "genexpr".
         self._handle_comprehension(node, 'genexpr')
 
-    def _handle_comprehension(self, node, comprehension_type):
-        # type: (ComprehensionNode, str) -> None
+    def _visit_first_comprehension_generator(self, node):
+        # type: (ComprehensionNode) -> None
+        if node.generators:
+            # first generator's iterator is visited in the current scope
+            first_generator = node.generators[0]
+            self.visit(first_generator.iter)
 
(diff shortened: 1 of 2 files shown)


joguSD:
Working on reducing complexity to pass the travis build.

PR #583, opened 2017-10-26
Added symlinks support for vendor directory walkthrough.

jcbertin:


jamesls:
Sorry for the delay.  I think this would be a good change, I'd just like to get a functional test added to make sure we don't regress on this behavior in the future.

jcbertin:
Sorry for delay too…

hannes-ucsc:
You also need to follow symlinks when hashing the vendor dir:

https://github.com/hannes-ucsc/chalice/commit/cebd09f632b65fd08d97fd0efa4aa64327db6f87

The tests should have caught that ;-)

jcbertin:
Should be ok now… 😉

jcbertin:
Please, can someone review this definitive version?

jcbertin:
ping…

hannes-ucsc:
@jamesls, is this something you are still considering for inclusion?

jcbertin:
🙋‍♂️

haidaraM:
Any news ? 

To complete this issue, we may also follow links when adding the `chalicelib` in the ZIP ? Currently a symlinked folder is not added in the ZIP. I can fill a new PR to address this specific case :-)

jcbertin:
@jamesls any comment? Did I miss something?

jcbertin:
@jamesls Come on, guys. Three years for that PR, really?

jamesls:
I'll take a look.  To be honest I don't quite remember the context here, and there's not a tracking issue nor a description in this PR to help remind me.  I need to figure out:

1. What's the current behavior with symlinks?
2. Would anyone ever want the existing behavior (should it be configurable or is this a bug fix)?
3. If so, should the default be changed to follow symlinks or should you have to opt-in via a config option?

The change itself looks fine.

jamesls:
After looking into it:

>  What's the current behavior with symlinks?

If it's a symlink to a file (e.g `newname.py -> ../realpackage2/c.py`) the file is added to the zipfile the the filename of the symlink and the contents of the original file (this is what we want).

If the symlink is to a directory (e.g. `linkpackage -> ../realpackage`) the directory is skipped.  This seems to be the issue here.  We ignore symlinks that are a directory.

> Would anyone ever want the existing behavior (should it be configurable or is this a bug fix)?

I can't see a case.  Right now, directory symlinks are ignored so there wouldn't be a reason to have a symlink in the vendor directory in the first place.

> If so, should the default be changed to follow symlinks or should you have to opt-in via a config option?

I would consider this a bugfix, we're ignoring directories that I think have a reasonable expectation to be created, especially because you have to create the `vendor/` directory yourself.


I also checked what the `zip` command line does and it has the behavior that this PR, so I think it's good that they're consistent.

tl;dr, ship-it, thanks for your patience.

jamesls:
Merged in 96c670d

PR #586, opened 2017-10-29
Add a link to Domovoi in the Related Projects section

kislyuk:


jamesls:
Sorry for the delay.  Looks good to me.

Also, what are you thoughts on eventually bringing this functionality into chalice?

kislyuk:
I would be quite happy if this was brought into Chalice. I'm not sure what it would take to accomplish that though, in terms of test strategy, code style, achieving a good fit with the Chalice architecture, etc. It would be great if you took a look and outlined a possible plan.

PR #588, opened 2017-10-31
Fix handling of trailing slash to match API Gateway

### LINKED ISSUE #582: No route found - exposing too much information?
By mistake, I tried to authenticate against a route in our application that did not exist, and was a bit "overwhelmed" by the response. 

To me it seems that the information returned by the Chalice app contains too much information.
As I tried to authenticate using an Authorization Bearer token, the complete token was "echoed" back along  with several indicators of what I might have been doing wrong.

Example (wrapped for readability-reasons):
```
{
    "message": "Authorization header requires 'Credential' parameter. 
      Authorization header requires 'Signature' parameter. 
      Authorization header requires 'SignedHeaders' parameter. 
      Authorization header requires existence of either a 'X-Amz-Date' or a 'Date' header. 
      Authorization=Bearer ey....nfg"
}
```

Wouldn't it have been easier just to tell the user that "..the requested route does not exist"?

Traced this to the `LocalGateway.handle_request()` method. I see that this is supposed to mirror the behavior of the API Gateway, which returns an even worse response. Would it be possible to override this, only return a simple message? ..and the more detailed if the Chalice app is running in debug mode?

Steps to reproduce:
1. Configure a route, e.g. ..../users/{id}
2. Configure the route such as it accepts an empty id value.
3. Request the route with id - works as expected
4. Request the route without id - works as expected
5. Request the route .../users (without the trailing slash) - returns the message mentioned above.

stealthycoin:
API Gateway will not match a URI against a route that has a capture
group as the last path component if that capture group would be filled
with an empty string. Example:

With the route:
/resource/{name}

/resource/bob matches
/resource/    does not match

Previously local mode would match both URIs to that route, setting the
name parameter to an empty string.

closes #582

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,11 +2,19 @@
 CHANGELOG
 =========
 
+
+Next Release (TBD)
+==================
+
+* Fix local mode handling of routes with trailing slashes
+  (`#582 <https://github.com/aws/chalice/issues/582>`__)
+
+
 1.0.4
 =====
 
 * Fix issue deploying some packages in Windows with utf-8 characters
-   (`#560 <https://github.com/aws/chalice/pull/560>`__)
+  (`#560 <https://github.com/aws/chalice/pull/560>`__)
 * Add support for custom authorizers with ``chalice package``
   (`#580 <https://github.com/aws/chalice/pull/580>`__)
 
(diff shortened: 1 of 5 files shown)


jamesls:
Looks good, I'm just going to add another test.  I want a test that verifies if you have `/a` and `/a/{capture}` that a request to `/a/` will route to `/a`.

I'll also go ahead and rebase on master.

jamesls:
Merged in bd05f56

Issue #593, opened 2017-11-03
query params default to different values when running locally V.S. running in lambda


nplutt:
Hey,

When running an application locally the `current_request.query_params` defaults to `{}` if there are no query params. This can be very misleading when running locally because when the lambda is deployed `current_request.query_params` defaults to `None`.

In order to fix this I would propose changing the local development server so that the  `current_request.query_params` also defaults to `None`.

Thanks,
Nick

dstufft:
Marking this as an enhancement, thanks!

PR #594, opened 2017-11-03
Fix Local Query Params Issue

nplutt:
When running an application locally the `current_request.query_params` defaults to `{}` if there are no query params. This can be very misleading when running locally because when the lambda is deployed `current_request.query_params` defaults to `None`.

Closes [#593](https://github.com/aws/chalice/issues/593)

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -167,7 +167,8 @@ class LambdaEventConverter(object):
                 },
             },
             'headers': {k.lower(): v for k, v in headers.items()},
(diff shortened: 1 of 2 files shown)


nplutt:
Hey, just wanted to follow up on this. Let me know if you see any issues or want anything else added/changed.

Thanks!

jamesls:
Hi sorry for the delay, taking a look now.  Thanks for the pull request.

PR #595, opened 2017-11-04
Add support for deleting unreferenced resources

jamesls:
As part of this work, general improvements to the executor
and deployer have been added that fixed several issues with
the update process.  You can now add new lambda functions
to your app.py and the deployer will properly record these
values in the deployed.json.

There are a few high level changes included.

* The deployed.json was updated to have a consistent format
  for all resources.  All resources now go under a ``resources``
  key that's specific to a stage.  Each resource must have a
  ``resource_type``.  This allows us to add support for new resource
  types without having to worry about updating the deployed.json
  file.  To handle the old format I've added a ``schema_version`` key.
  I'll still need to add some backwards compatibility code that can
  load the v1 format.  For now you can't mix the v1 and v2 formats.
  This will need to be addressed before merging to master.
* The APICall has been split out into 6 new instructions and the planner
  and executor now take a list of Instructions.  An APICall is now
  a type of Instruction.  This level of granularity was needed to
  properly track which resources we've accounted for locally so we
  can compare them to the set of remote resources.
* One of the new instructions, ``JPSearch``, requires a dependency
  on JMESPath.  This was required in order to properly generate
  the deployed.json file.  The ``create_function`` call returns
  a function arn, but the ``update_function`` returns the entire
  response of ``update_function_code``.  We need to extract out
  the ``FunctionArn`` key from the response.  We'll also need this
  capability in the future.  In practice this isn't a big deal,
  botocore already requires JMESPath, we're just also calling out
  an explicit dependency in our setup.py.

### DIFF
--- a/chalice/config.py
+++ b/chalice/config.py
@@ -2,7 +2,7 @@ import os
 import sys
 import json
 
-from typing import Dict, Any, Optional  # noqa
+from typing import Dict, Any, Optional, List  # noqa
 from chalice import __version__ as current_chalice_version
 from chalice.app import Chalice  # noqa
 from chalice.constants import DEFAULT_STAGE_NAME
@@ -275,6 +275,29 @@ class Config(object):
         return clone
 
     def deployed_resources(self, chalice_stage_name):
+        # type: (str) -> Optional[DeployedResources2]
+        """Return resources associated with a given stage.
+
+        If a deployment to a given stage has never happened,
+        this method will return a value of None.
+
+        """
+        # This is arguably the wrong level of abstraction.
+        # We might be able to move this elsewhere.
+        deployed_file = os.path.join(self.project_dir, '.chalice',
+                                     'deployed.json')
+        if not os.path.isfile(deployed_file):
+            return None
+        with open(deployed_file, 'r') as f:
+            data = json.load(f)
+        schema_version = data.get('schema_version', '1.0')
+        if schema_version == '2.0':
+            if chalice_stage_name not in data['stages']:
+                return None
+            return DeployedResources2(data['stages'][chalice_stage_name])
+        raise ValueError("")
+
+    def old_deployed_resources(self, chalice_stage_name):
         # type: (str) -> Optional[DeployedResources]
         """Return resources associated with a given stage.
 
@@ -293,6 +316,20 @@ class Config(object):
         if chalice_stage_name not in data:
             return None
         return DeployedResources.from_dict(data[chalice_stage_name])
+
+
+class DeployedResources2(object):
+    def __init__(self, deployed_values):
+        # type: (Dict[str, Any]) -> None
+        self._deployed_values = deployed_values
+
+    def resource_values(self, name):
+        # type: (str) -> Dict[str, str]
+        return self._deployed_values['resources'][name]
+
+    def resource_names(self):
+        # type: () -> List[str]
+        return list(self._deployed_values['resources'])
 
 
 class DeployedResources(object):
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -60,7 +60,7 @@ class RemoteState(object):
                            role_arn=role['Arn'])
 
 
-class Sweeper(object):
+class UnreferencedResourcePlanner(object):
 
     def execute(self, plan, config):
         # type: (List[models.Instruction], Config) -> None
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -1,8 +1,9 @@
 import json
 
 import attr
-from typing import List, Dict, Any, Optional, Union, Tuple, cast  # noqa
+from typing import List, Dict, Any, Optional, Union, Tuple, Set, cast  # noqa
 
+from chalice.config import Config, DeployedResources2  # noqa
 from chalice.utils import OSUtils  # noqa
 from chalice.deploy import models
 from chalice.awsclient import TypedAWSClient, ResourceDoesNotExistError  # noqa
@@ -59,6 +60,42 @@ class RemoteState(object):
                            role_arn=role['Arn'])
 
 
+class Sweeper(object):
+
+    def execute(self, plan, config):
+        # type: (List[models.Instruction], Config) -> None
+        marked = self._mark_resources(plan)
+        deployed = config.deployed_resources(config.chalice_stage)
+        if deployed is not None:
+            deployed_resource_names = deployed.resource_names()
+            remaining = list(set(deployed_resource_names) - set(marked))
+            self._plan_deletion(plan, remaining, deployed)
+
+    def _mark_resources(self, plan):
+        # type: (List[models.Instruction]) -> List[str]
+        marked = []  # type: List[str]
+        for instruction in plan:
+            if not isinstance(instruction, models.RecordResourceValue):
+                continue
+            marked.append(instruction.resource_name)
+        return marked
+
+    def _plan_deletion(self,
+                       plan,       # type: List[models.Instruction]
+                       remaining,  # type: List[str]
+                       deployed,   # type: DeployedResources2
+                       ):
+        # type: (...) -> None
+        for name in remaining:
+            resource_values = deployed.resource_values(name)
+            if resource_values['resource_type'] == 'lambda_function':
+                apicall = models.APICall(
+                    method_name='delete_function',
+                    params={'function_name': resource_values['lambda_arn']},
+                )
+                plan.append(apicall)
+
+
 class PlanStage(object):
     def __init__(self, remote_state, osutils):
         # type: (RemoteState, OSUtils) -> None
@@ -66,8 +103,8 @@ class PlanStage(object):
         self._osutils = osutils
 
     def execute(self, resources):
-        # type: (List[models.Model]) -> List[models.APICall]
-        plan = []  # type: List[models.APICall]
+        # type: (List[models.Model]) -> List[models.Instruction]
+        plan = []  # type: List[models.Instruction]
         for resource in resources:
             name = 'plan_%s' % resource.__class__.__name__.lower()
             handler = getattr(self, name, None)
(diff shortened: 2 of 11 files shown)


chalice/deploy/planner.py
```diff
@@ -59,15 +60,51 @@ def get_remote_model(self, resource):
                            role_arn=role['Arn'])
 
 
+class Sweeper(object):
```
kyleknap: `Sweeper` is still a little vague to me. Mainly I first interpreted is as an abstraction that sweeps through all of the resources, deleting them or sweeping through the two deployments (the previous one and the current planned one) and deletes the ones accounted for. So maybe make it something more specific like `UnreferencedResourceSweeper`?

chalice/deploy/planner.py
```diff
@@ -155,6 +216,13 @@ def plan_managediamrole(self, resource):
                         'policy_name': resource.role_name,
                         'policy_document': document},
                 resource=resource
+            ),
+            models.Pop(),
+            models.Push(resource.role_arn),
```
kyleknap: So when is it recommended to the `Pop` and `Push` as opposed to a `StoreValue`?  The way I could have seen this code go is not do the whole Pop/Push workflow and just have the `APICall` and `RecordResourceValue` that provided the `resource.role_arn` directly. Seems like we need to do an unnecessary amount of hops to provide the `resource.role_arn` to the `RecordResourceValue` object.

chalice/deploy/planner.py
```diff
@@ -59,15 +60,51 @@ def get_remote_model(self, resource):
...
+            deployed_resource_names = deployed.resource_names()
+            remaining = list(set(deployed_resource_names) - set(marked))
+            self._plan_deletion(plan, remaining, deployed)
+
+    def _mark_resources(self, plan):
+        # type: (List[models.Instruction]) -> List[str]
+        marked = []  # type: List[str]
+        for instruction in plan:
+            if not isinstance(instruction, models.RecordResourceValue):
+                continue
```
kyleknap: Is there a reason this is just not:
```python
if isinstance(instruction, models.RecordRecourceValue):
     marked.append(instruction.resource_name)
```

jamesls:
I don't see how chaining is more explicit that a list of instructions.  I find your examples harder to read.  Consider your first example:

```
                models.APICall(
                    method_name='create_role',
                    params={'name': resource.role_name,
                            'trust_policy': resource.trust_policy,
                            'policy': document},
                    resource=resource
                ).store_value(variable_name=varname),
                models.RecordResourceValue(
                    resource_type='iam_role',
                    resource_name=resource.resource_name,
                    name='role_arn',
                    variable_name=varname,
                )
          ]
```

vs.

```
            varname = '%s_role_arn' % resource.role_name
            return [
                models.APICall(
                    method_name='create_role',
                    params={'name': resource.role_name,
                            'trust_policy': resource.trust_policy,
                            'policy': document},
                    resource=resource
                ),
                models.StoreValue(varname),
                models.RecordResourceValue(
                    resource_type='iam_role',
                    resource_name=resource.resource_name,
                    name='role_arn',
                    variable_name=varname,
                )
            ]
```

In your first example, you have to carefully note that there's a `.store_value()` after the `APICall` which still is a separate instruction but is specified differently than the `APICall` and `RecordResourceValue` instruction.  I find it much more clear to see that there's 3 instructions we're adding in the latter example.

> ...get better validation value handling on specific API calls so say if we were trying to store a value for an APICall that did not return anything or using a JMESPath expression that does not resolve to anything.

The existing approach does not preclude that.  It's just as straightforward to have a validator that iterates through the instructions and verifies that a JMESPath expression resolves to something if an APICall instruction precedes it.

I think we can possibly update the `RecordResourceValue` to also accept an optional "value" argument which will use instead of pop()'ing from the stack.  But in general I'm not really a fan of the chaining API because there's no longer a single way to build up instructions.

chalice/config.py
```diff
@@ -275,6 +275,29 @@ def scope(self, chalice_stage, function_name):
...
+        with open(deployed_file, 'r') as f:
+            data = json.load(f)
+        schema_version = data.get('schema_version', '1.0')
+        if schema_version == '2.0':
+            if chalice_stage_name not in data['stages']:
+                return None
+            return DeployedResources2(data['stages'][chalice_stage_name])
+        raise ValueError("")
+
+    def old_deployed_resources(self, chalice_stage_name):
```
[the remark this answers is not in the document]
jamesls: I will need something to load the "old" version of the deployed.json file (to migrate to the latest version) so this method will need to exist in some form, but I'm not tied to the name.

chalice/deploy/planner.py
jamesls: If the `Sweeper` name is confusing, it might be better to just call it `UnreferencedResourcePlanner`?

chalice/deploy/planner.py
jamesls: No, I'll update.

chalice/deploy/planner.py
jamesls: * `StoreValue` - If you're going to reference it later either in later instructions or as a param to an `APICall` instruction
* `Push` - Used in the next instruction.

It's a tradeoff between how simple your instructions are (easier executor to implement but more instructions to accomplish something) vs. more complex instructions with slightly more complicated executor.

The `RecordResourceValue` currently has two modes:

* It can record a value from a variable name (something that you've previously `StoreValue`'d)
* It can record a value that's at the top of the stack

What you'd be proposing here is a third mode:

* Record a value that's specified directly as an argument to this instruction.

It's doable, it's just a tradeoff of how complex we want to make the instruction.  The argument could go the other way as well.  "Why have a third mode when you can already accomplish this with the existing instructions."

chalice/deploy/planner.py
kyleknap: Yeah that's better.

chalice/config.py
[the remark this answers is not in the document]
kyleknap: Interesting. Yeah it just seemed inconsistent with the class names which through me off. Furthermore if there is a third version of the `deployed.json` to ever come out what would the `old_deployed_resources()` method be renamed to?

chalice/deploy/planner.py
kyleknap: So looking at it again, I think my biggest issue is with direct exposure of the stack. Thinking through the interface from the perspective of someone new adding deployment support for a new resource, the biggest concern was the exposed stack interface and most of my suggestions (this third mode and the chaining suggestion) were related to how I would want to avoid it.

Specifically, my concerns with interacting with the stack were:
* If I pushed something onto the stack in an `execute()` call, what is the guarantee that the value pushed on will not affect a subsequent `execute()` call? And does that mean that I have to handle cleanup of what was pushed onto the stack? Right now this is protected by all of the `execute()` calls that we are doing starts with an `APICall()`, but there is no guarantee in the future. This concern was also furthered by the existence of a `Pop()` instruction that seemed to exist for the sole purpose of cleaning up values as no value is returned from the `Pop()` instruction.
* The stack interface opens the door for some poorly designed `execute()` workflows. Ideally, users of the interface should process the value for variable storage or resource recording immediately after the `Push()` or `APICall()`. But nothing is stopping them from making multiple  back to back `APICall()` or `Push()` and not immediately process the value for variable storage. Then at that point, it gets complicated because they need to keep track of the current state of all values in the stack. So like if they popped off a value, later down the road I would need to be aware that if they are storing is a variable or recording a resource value it will be of value that was underneath of what I had popped off.


I think a lot of these concerns would be satisfied if the stack was switched to more of a register, where:
* The register stores the value from the previous `APICall`. So if you want to store it or record it do so then as after the next `APICall` it will be a new value.
* On the call of every `execute()`, the register is cleared to start so there is no previous state.

That way it forces me to interact with it more in the way that is recommended and I do not have to worry about state carrying over.


Then as to the comment about the third mode of `RecordResourceValue`, could we just have different classes for these instructions? So to make sure the instruction class is not too complicated so we could break this out to a `RecordResourceFromValue`, `RecordResourceFromVariable`, and `RecordResourceFromStack` (or register or whatever comes out of this).

jamesls: > If I pushed something onto the stack in an execute() call, what is the guarantee that the value pushed on will not affect a subsequent execute() call? And does that mean that I have to handle cleanup of what was pushed onto the stack?

This isn't how the executor is intended to be used.  You build up an entire list of instructions and you ask the executor to execute it.  If we wanted to support multiple calls to `Executor.execute()` we could just add a `clear()` or a `reset()` state method.

Additionally, I don't see how your suggestion of register resolves this issue.  This concern is about leftover state not being cleaned up in subsequent `execute()` calls.  You have that whether you use a stack or registers.

> The stack interface opens the door for some poorly designed execute() workflows.

So does registers.  If you screw up the set of instructions to execute, you're going to have a problem.  Nothing is stopping multiple `APICall()s` from writing to the same register and thereby wiping out a variable that was previously written.

> So like if they popped off a value, later down the road I would need to be aware that if they are storing is a variable or recording a resource value it will be of value that was underneath of what I had popped off.

I don't understand what this means.  If you want to store a value, then don't pop something off the stack.  Who's "they" and "I" in this case?  No one's going to inject arbitrary instructions in the middle of the list of instructions needed to modify a resource.

> That way it forces me to interact with it more in the way that is recommended and I do not have to worry about state carrying over.

Not sure what this means either.  What is the "way that is recommended"?  State will carry over in either approach.  The way to solve that is with some method to clear out state.


I'm not opposed to removing the stack, but none of these arguments make a convincing case that using registers alleviate any of these problems.  In fact the typical tradeoff with a stack vs. register-based approach is that a stack is much easier to implement (and emit) at the cost of slightly more instructions.  I would like to know why *that* isn't the right tradeoff here.

kyleknap: Talked about this offline. Given this rewrite is a work in-progress, we will go ahead with this approach because it is simpler to implement and there is future code that will utilize the interface more. We will look into refactoring this once more of the deployment logic for the other resources are fleshed out using the interface

PR #611, opened 2017-11-16
Allow user to override default codebuild image

### LINKED ISSUE #609: Allow generate-pipeline to not fail even when using python3
If you use `generate-pipeline` when running on python3 you get:

```
$ chalice generate-pipeline foo
CodeBuild does not yet support python version 'python3.6'.
```

I'd like an option that still generates the pipeline provided you specify some codebuild image.  For example, if you provide a `--codebuild-image python:3.6.1` then we'd generate the template for you.

jamesls:
If you're not using python2.7, you can still generate a
pipeline as long as you provide a codebuild image to use.
This allows pipeline generation in python3.

Fixes #609.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -312,8 +312,10 @@ def generate_pipeline(ctx, codebuild_image, filename):
     from chalice import pipeline
     factory = ctx.obj['factory']  # type: CLIFactory
     config = factory.create_config_obj()
-    p = pipeline.CreatePipelineTemplate(codebuild_image)
-    output = p.create_template(config.app_name, config.lambda_python_version)
+    p = pipeline.CreatePipelineTemplate()
+    output = p.create_template(config.app_name,
+                               config.lambda_python_version,
+                               codebuild_image)
     with open(filename, 'w') as f:
         f.write(serialize_to_json(output))
 
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -286,10 +286,14 @@ def package(ctx, single_file, stage, out):
 
 
 @cli.command('generate-pipeline')
+@click.option('-i', '--codebuild-image',
+              help=("Specify default codebuild image to use.  "
+                    "This option must be provided when using a python "
+                    "version besides 2.7."))
 @click.argument('filename')
 @click.pass_context
-def generate_pipeline(ctx, filename):
-    # type: (click.Context, str) -> None
+def generate_pipeline(ctx, codebuild_image, filename):
+    # type: (click.Context, str, str) -> None
     """Generate a cloudformation template for a starter CD pipeline.
 
     This command will write a starter cloudformation template to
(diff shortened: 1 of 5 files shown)


chalice/pipeline.py
```diff
@@ -45,12 +38,14 @@ class CreatePipelineTemplate(object):
...
 
-    def __init__(self):
-        # type: () -> None
-        pass
+    def __init__(self, codebuild_image=None):
+        # type: (Optional[str]) -> None
+        self._codebuild_image = codebuild_image
 
-    def _codebuild_image(self, lambda_python_version):
+    def _get_codebuild_image(self, lambda_python_version):
```
kyleknap: I would expect the  `lambda_python_version` here be optional now since you could have provided an explicit codebuild image. I think this applies to the `create_template` method as well.

chalice/pipeline.py
```diff
@@ -45,12 +38,14 @@ class CreatePipelineTemplate(object):
         "Outputs": {},
     }
 
-    def __init__(self):
-        # type: () -> None
-        pass
+    def __init__(self, codebuild_image=None):
```
kyleknap: What was the particular reason for adding the `codebuild_image` to the initializer instead of the `create_template()` method?  At first glance, I would assume the `codebuild_image` would be provided to the `create_template()` method and from there you would validate that both `codebuild_image` and `lambda_python_version` were not provided. By putting it in the initializer, an instance of the `CreatePipelineTemplate` could not be reused to build different pipelines for different images.

tests/functional/cli/test_cli.py
```diff
@@ -311,6 +311,23 @@ def test_can_generate_pipeline_for_all(runner):
             assert "Outputs" in template
 
 
+def test_no_errors_if_override_codebuil_image(runner):
```
kyleknap: `codebuil` -> `codebuild`

chalice/cli/__init__.py
```diff
@@ -286,10 +286,14 @@ def package(ctx, single_file, stage, out):
 
 
 @cli.command('generate-pipeline')
+@click.option('-i', '--codebuild-image',
+              help=("Specify default codebuild image to use.  "
+                    "This option must be provided when using a python "
+                    "version besides 2.7."))
```
kyleknap: Do you think we should print out that specific error message now when they try to use an unknown python with no `--codebuild-image`?

chalice/pipeline.py
jamesls: Given I'm expecting to add more configuration options for generating your pipeline (e.g separate buildspec.yml, github source) I'm thinking about creating a `PipelineConfig` object you can pass to `create_template` to customize this process.  That way that object can own the validation.  It would also make it a little easier to add new options in the future.

jamesls:
@kyleknap Should be ready to look at again.  I ended up keeping the lambda_python_version required for now.  I think we can always make it optional in the future, but we couldn't go back from optional to required.  I'm not positive at this point that lambda python version might become required again in the future so I'd prefer to keep it as is for now.

PR #612, opened 2017-11-16
Add option to map env vars to template params in package command

### LINKED ISSUE #608: Add option to extract lambda env vars to cloudformation parameters
When you run `chalice package`, it would be really helpful if we could take the lambda environment variables you have in your config file and map them to template parameters.  That way you can override these values when you deploy but still reuse the same SAM template.  They'd still default to the existing values in the `.chalice/configjson`.  I'd prefer this was an additional flag you can specify.

The only thing I've ran into is that we'd need to transform the env vars to something supported by cloudformation.  The typical format of `UPPER_SNAKE_CASE` for env vars won't work for parameter key names.  We could just camel case them so you'd get `UPPER_SNAKE_CASE` -> `UpperSnakeCase`.

jamesls:
Fixes #608.

For example, given this `.chalice/config.json`:

```json
$ cat .chalice/config.json
{
  "version": "2.0",
  "app_name": "foo",
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "environment_variables": {
        "FOO_BAR": "yeah",
        "BAR_BAZ": "ok"
      }
    }
  }
}
```

You can get these cfn template params:

```
$ chalice package --map-env-to-params outdir
$ cat outdir/sam.json
...
  "Parameters": {
    "ApiHandlerFooBar": {
      "Default": "yeah",
      "Type": "String"
    },
    "ApiHandlerBarBaz": {
      "Default": "ok",
      "Type": "String"
    }
  }
```

Notes:

* Need to translate `UPPER_SNAKE` to `UpperSnake` because of the alphanumeric restriction on cfn params
* Rather than using the `to_cfn_resource_name` which appends a hash based on the name, I went for a more direct translation.  This is because these values are user facing and I wanted something with a more readable name (vs. resource names which a user doesn't directly need to worry about).
* In the case of duplicate names, an exception is raised.

### DIFF
--- a/tests/unit/test_package.py
+++ b/tests/unit/test_package.py
@@ -10,12 +10,23 @@ from chalice.deploy.swagger import SwaggerGenerator
 
 @pytest.fixture
 def mock_swagger_generator():
-    return mock.Mock(spec=SwaggerGenerator)
+    swagger = mock.Mock(spec=SwaggerGenerator)
+    swagger.generate_swagger.return_value = {
+        'swagger': 'document'
+    }
+    return swagger
 
 
 @pytest.fixture
 def mock_policy_generator():
     return mock.Mock(spec=ApplicationPolicyHandler)
+
+
+@pytest.fixture
+def cfn_gen(mock_swagger_generator, mock_policy_generator):
+    p = package.SAMTemplateGenerator(mock_swagger_generator,
+                                     mock_policy_generator)
+    return p
 
 
 def test_can_create_app_packager():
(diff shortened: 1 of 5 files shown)


tests/unit/test_package.py
```diff
@@ -76,212 +83,196 @@ def test_sam_injects_policy(sample_app,
...
-    p = package.SAMTemplateGenerator(
-        mock_swagger_generator, mock_policy_generator)
-    mock_swagger_generator.generate_swagger.return_value = {
-        'swagger': 'document'
+def test_can_extract_env_vars_as_cfn_params(sample_app, cfn_gen):
+    config = Config.create(
+        chalice_app=sample_app,
+        api_gateway_stage='dev',
+        environment_variables={
+            'MY_ENV_VAR': 'BAR'
```
JordonPhillips: Could we also get a test to make sure it capitalizes properly from lowercase? So `my_env_var` becomes `MyEnvVar`

PR #614, opened 2017-11-17
Add docs example of specific lambda functions config

### LINKED ISSUE #607: Allow lambda settings outside of config.json
From my understanding, the `lambda_timeout` and `lambda_memory_size` values in the config file will always apply to any lambdas generated by chalice. Obviously this includes the main lambda for the application, but also applies to lambdas created using scheduled events.

It would be nice to be able to overwrite lambda settings defined in the config file for such events. A use case could be a daily scheduled event that is expected to run all functions in your app and requires a longer timeout and/or more memory than an individual route in your app.

By the way, chalice is a pleasure to use. Thanks!

BigChief45:
Closes #607


BigChief45:
Would appreciate some feedback

CC @jamesls

Issue #584, opened 2017-10-26
Setup environment variables when loading app.py


JordonPhillips:
Chalice imports `app.py` during execution [here](https://github.com/aws/chalice/blob/master/chalice/cli/factory.py#L166). Naturally this results in all of the code at module level being executed. The problem is that we don't do anything to setup the environment beforehand. This results in an issue where local execution is different than remote execution.

By setting up the environment, we would be able to allow features such as #576 without any additional work. Stage-specific setup would be generally possible.

As an example, here's an app that I may want to deploy:

```python
from chalice import Chalice
from chalice import CognitoUserPoolAuthorizer

app = Chalice(app_name='resources')
authorizer = CognitoUserPoolAuthorizer(
    'MyPool', provider_arns=[os.environ['USER_POOL_ARN']])


@app.route('/', methods=['GET'], authorizer=authorizer)
def index():
    return {'success': True}
```

With this config:
```json
{
  "version": "2.0",
  "app_name": "resources",
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "environment_variables": {
        "USER_POOL_ARN": "arn:aws:cognito-idp:us-west-2:...:userpool/name"
      }
    },
    "prod": {
      "api_gateway_stage": "prod-api",
      "environment_variables": {
        "USER_POOL_ARN": "arn:aws:cognito-idp:us-west-2:...:userpool/othername"
      }
    }

  },
  "profile": "tmp-chalice"
}
```

As a user, I would expect `USER_POOL_ARN` to be set based on which stage I'm deploying.

Additionally, you have issues like #573 that suffer from being run locally in a different environment. I'm not sure how we want to handle cases like that though, they may need to be separate cases.

kadrach:
I have the same problem with `chalice package`. For example, take a Lambda function with an IAM role that grants read access to an s3 bucket:

```python
import pandas as pd
from chalice import Chalice

app = Chalice(app_name='hello')
data = pd.read_csv('s3://mybucket/some_file.txt')  # no permissions here during `package`

@app.route('/')
def hello():
    # do something with data
    pass
```

The CI toolchain should not have these permissions. I can catch and ignore the exception, but I'd not want to do this in the actual environment. Is there a clever way to work around this?

dmulter:
+1

dmulter:
Until this is resolved, is there any way to determine the stage an app is in during initialization? I could create my own parameters per stage mechanism, but I need some form of api.xxx() to figure out the stage.

[...]

PR #618, opened 2017-11-22
Setup Environment Variables Locally

bubba-h57:
The current implementation [cli/__init__.py#L97](https://github.com/aws/chalice/blob/master/chalice/cli/__init__.py#L97) sets environment variables after [factory.create_config_obj()](https://github.com/aws/chalice/blob/master/chalice/cli/factory.py#L90) is called, which happs to call [self.load_chalice_app()](https://github.com/aws/chalice/blob/master/chalice/cli/factory.py#L108) before `run_local_server()` [calls it as well](https://github.com/aws/chalice/blob/master/chalice/cli/__init__.py#L98).

The pull request resolves issue #584 by configuring the environment in the factory, prior to calling `self.load_chalice_app()` in such a way that it is only done when running *local* but ensures that the environment is properly configured as expected.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -89,12 +89,10 @@ def local(ctx, host='127.0.0.1', port=8000, stage=DEFAULT_STAGE_NAME):
 def run_local_server(factory, host, port, stage, env):
     # type: (CLIFactory, str, int, str, MutableMapping) -> None
     config = factory.create_config_obj(
-        chalice_stage_name=stage
+        chalice_stage_name=stage,
+        env=env
     )
-    # We only load the chalice app after loading the config
-    # so we can set any env vars needed before importing the
-    # app.
-    env.update(config.environment_variables)
+
     app_obj = factory.load_chalice_app()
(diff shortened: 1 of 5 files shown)


owenbrown:
It's annoying having to update my environmental variables in two places when I run chalice locally. It would be great if Chalice Local sourced all environmental variables. Or, maybe come up with a method different than os.enviro["VARIABLENAME"] for accessing environmental variables, if that's simpler.

ben-adazza:
I made a non-invasive work around for this with https://stedolan.github.io/jq/. This script assumes it is run in the same directory as the `.chalice` directory. 
```
#!/usr/bin/env bash

tier=$1
eval $(jq -r ".environment_variables | to_entries | map(\"export \(.key)=\(.value|tostring)\")|.[]" .chalice/config.json)
eval $(jq -r ".stages.$tier.environment_variables | to_entries | map(\"export \(.key)=\(.value|tostring)\")|.[]" .chalice/config.json)
exec chalice "${@:2}" --stage ${tier}
```

stealthycoin:
Thanks for the PR, it looks like this issue was fixed in #833

PR #619, opened 2017-11-25
Add github source and buildspec extraction to generate-pipeline

jamesls:
This PR adds two options to the `generate-pipeline` command:

* The ability to specify a github source via `--source github`.  CFN template params are injected that require you to specify the owner and repo name when this option is provided.
* The ability to write the CodeBuild buildspec file to a separate file.  The default behavior is to include the contents as part of the cloudformation template.


jamesls:
@kyleknap Yep exactly.  It means that the people that have access to the source repo can now control how the codebuild process happens (vs. whoever manages the cfn template, which could be either separate teams or at least separate roles).  I'll update the help message with this.

jamesls:
@kyleknap ready to look at again.

Issue #344, opened 2017-05-20
Increase lambda deployment timeout?


jonapich:
I had to raise botocore/endpoint.py/DEFAULT_TIMEOUT manually in order to succeed in pushing my lambda function.

jamesls:
Interesting.  Out of curiosity what value did you have to change it to?

jonapich:
120 (out of frustration) ;)

rerngvit:
I have precisely the same problem. I got "Connection aborted. Timed out sending your app to Lambda.", repeatedly, when trying to do "chalice deploy".  It would be great, if a command line option to specify timeout is available.

zjost:
Same problem here.  This started happening when I added numpy/pandas to my requirements.txt, so presumably it has something to do with bundling that up and sending it.

nplutt:
Hey @jamesls, I started working on this and was wondering if you like the idea of allowing this to be configured through an option in the `deploy` command.  I was imagining something like:
```bash
$ chalice deploy --botocore-timeout 120
```
After the timeout value is consumed by the cli I was thinking about changing the `CLIFactory` class so that a `botocore_timeout` value can be set and modified.

lenards:
When this is available, I believe they settled on `--connection-timeout` in #629, based on the additions in commit [eafa12975](https://github.com/aws/chalice/commit/eafa1297506e204ff0d97df90afc2ae73339ddae#diff-1eb429fc268c241f1f6f9e8f44508a33R124)

However, at the time of this post, I think this in `master` and not up on pypi yet. 


```
✘-1 ~/gh/tableflipless [master L|✚ 2…7] 
08:54 $ pipenv run chalice deploy --botocore-timeout 145
Error: no such option: --botocore-timeout
✘-2 ~/gh/tableflipless [master L|✚ 2…7] 
08:56 $ pipenv run chalice deploy --connection-timeout 145
Error: no such option: --connection-timeout
✘-2 ~/gh/tableflipless [master L|✚ 2…7] 
08:56 $ pipenv run chalice deploy --connection-timeout=140
Error: no such option: --connection-timeout
```

With `chalice==1.1.0`

nplutt:
You are correct, if you look at the change log this feature will be released with the next release which is currently TBD.

lenards:
Cool, thanks!

I just wanted to tag this issue with the command flag that was used so if someone lands here from a search, they got it.

moneroexamples:
Not the change timeout command is:

[...]

PR #629, opened 2017-12-05
Botocore timeout

nplutt:
Enhancement for issue #344. 

Let me know if you would like me to change anything. I decided to go with the option of consuming the timeout value through the cli but I could also see if you want this to be consumed through the config file.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -121,13 +121,13 @@ def run_local_server(factory, host, port, stage, env):
               help=('Name of the Chalice stage to deploy to. '
                     'Specifying a new chalice stage will create '
                     'an entirely new set of AWS resources.'))
-@click.option('--botocore-timeout',
+@click.option('--connection-timeout',
               type=int,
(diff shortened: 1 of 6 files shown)


chalice/cli/__init__.py
```diff
@@ -121,17 +121,27 @@ def run_local_server(factory, host, port, stage, env):
               help=('Name of the Chalice stage to deploy to. '
                     'Specifying a new chalice stage will create '
                     'an entirely new set of AWS resources.'))
+@click.option('--botocore-timeout',
```
jamesls: I'd prefer we just call this `--connection-timeout`.

nplutt:
Hey @jamesls, I made all of the requested changes. Thanks for the review, also good catch on the `set_default_client_config()` method it really simplified the code!

jamesls:
Rebased to a single commit (eafa1297506e204ff0d97df90afc2ae73339ddae) and merged via f091f7d

Issue #565, opened 2017-10-07
After fail to create IAM role, cannot deploy anymore.


chroju:
Hello.

When I execute `chalice deploy`, faced error of IAM role.

```bash
$ chalice deploy --no-autogen-policy
Creating role: sample-chalice
ERROR - While deploying your chalice application, received the following error:

 An error occurred (MalformedPolicyDocument) when calling the PutRolePolicy
 operation: Syntax errors in policy.
```

and I have fixed syntax error, deploy again.
However, the IAM role had been made incompletely by the last deploy, so the deploy failed with another error.

```
$ chalice deploy --no-autogen-policy
Updating IAM policy for role: sample-chalice
ERROR - While deploying your chalice application, received the following error:

 An error occurred (NoSuchEntity) when calling the DeleteRolePolicy operation:
 The role policy with name sample-chalice cannot be found.
```

If I wanted deploy to succeed, I had to delete incomplete IAM role once with AWS console or AWS CLI. `chalice delete` command was useless in this case.

Is this as you expected ?

## environment

* chalice 1.0.3
* Python 3.6.1

jamesls:
Thanks for reporting, this is a bug.  We'll get this fixed.

jamesls:
Fixed via https://github.com/aws/chalice/pull/636

mabdullah353:
I have started the app using chalice and stucked, the steps that i did on EC2 ubuntu instance were following:

```
ubuntu@ip-172-31-44-209:~$ chalice new-project traindexServiceAPI
ubuntu@ip-172-31-44-209:~$ cd traindexServiceAPI/
ubuntu@ip-172-31-44-209:~/traindexServiceAPI$ chalice deploy
Creating role: traindexServiceAPI-dev
ERROR - While deploying your chalice application, received the following error:

 An error occurred (AccessDenied) when calling the CreateRole operation: User: 
 arn:aws:iam::xxxxx:user/devop is not authorized to perform: 
 iam:CreateRole on resource: arn:aws:iam::xxxxx:role/traindexServiceAPI-
 dev

# After allowing iam:CreateRole from AWS Console. 
ubuntu@ip-172-31-44-209:~/traindexServiceAPI$ chalice deploy
Creating role: traindexServiceAPI-dev
ERROR - While deploying your chalice application, received the following error:

 An error occurred (AccessDenied) when calling the PutRolePolicy operation: 
 User: arn:aws:iam::xxxxx:user/devop is not authorized to perform: 
[...]

PR #632, opened 2017-12-06
Failed iam role creation

nplutt:
Fixes bug #565.

So there ended up being a couple issue that I fixed here, first being bug #565 which was fixed by making a change to the `awsclient.py` file.

The second issue that I discovered while working on this bug was the following. If you do:
```bash
$ chalice new-project muck
$ cd muck
$ chalice deploy --no-autogen-policy
```
The same bug happens because the `policy-env.json` file does not exist. This results in a policy document of:
```json
"Version": "2012-10-17",		
"Statement": []
```
being uploaded which throws a `MalformedPolicyDocument` error. In order to fix this issue I made changes to the `deployer.py` file. I removed the default policy and now raise a `RuntimeError` and alert the user to the fact that they do not have a `policy-env.json` file. 

The last change that I made was to handle malformed json by adding nicer logging modeled after the config file loader.

Thanks!

### DIFF
--- a/tests/functional/test_package.py
+++ b/tests/functional/test_package.py
@@ -732,8 +732,10 @@ def test_can_create_app_packager_with_no_autogen(tmpdir):
     appdir = _create_app_structure(tmpdir)
 
     outdir = tmpdir.mkdir('outdir')
+    default_params = {'autogen_policy': True}
     config = Config.create(project_dir=str(appdir),
-                           chalice_app=sample_app())
+                           chalice_app=sample_app(),
+                           **default_params)
     p = package.create_app_packager(config)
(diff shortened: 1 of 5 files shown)


tests/functional/test_package.py
```diff
@@ -732,8 +732,10 @@ def test_can_create_app_packager_with_no_autogen(tmpdir):
     appdir = _create_app_structure(tmpdir)
```
nplutt: From what I could tell it doesn't look like this test `test_can_create_app_packager_with_no_autogen` has anything to do with the `deploy --no-autogen-policy` command so I changed it to create the policy automatically so as not to trigger the `RuntimeError`.

PR #645, opened 2017-12-20
Implement Resource Deletion

jamesls:
### Implement resource deletion

This PR adds a `delete-new` command, similar to the `deploy-new` command.

The basic idea is to rely on the sweeper code to handle deletions.
Conceptually this is similar to creating an empty chalice app
and running `chalice deploy`.

The other alternative was create a new planner type that specifically
handled deletions.  The reason for leveraging the sweeper for deploys
is two fold:

1. We have to write sweeper logic anyways.  This was originally to
   handle the case where you remove resources from your app.py and
   rerun `chalice deploy`.  This logic has to stay regardless of what's
   decided for deletion.
2. A deletion planner would unnecessarily have to take into account
   what's currently specified in your app.py along with the remote state
   (stuff that's already been deployed).  This is unnecessary
   complication.  For deletion we only care about what exists
   remotely.

Another benefit of (2) is that we handle an edge case where you
remove resources from your app.py but don't run `chalice deploy`.
In the old deployer code, we base deletion based on your app.py
so we'd essentially leak a resource.  This is now handled by this
new deletion code because it's always based on the deployed.json
file.

As part of this change, the first commit is a change to the deployed.json
structure.  Instead of a dictionary of deployed values, the `resources`
key is now a list.  This list is the order in which the resources were
deployed by the executor.  The sweeper is updated to always iterate
through the resources in reverse order so we always delete resources
in the reversed order they were created (even in the case of deletions
for a redeploy).

### Example Test

From a new project, we have two lambda functions, with
explicit policies defined:

```
 $ cat app.py
from chalice import Chalice

app = Chalice(app_name='test-deletes')


@app.lambda_function()
def foo(event, context):
    return {'hello': 'foo'}


@app.lambda_function()
def bar(event, context):
    return {'hello': 'bar'}


$ cat .chalice/config.json
{
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "lambda_functions": {
        "foo": {"iam_policy_file": "policy-one.json", "autogen_policy": false},
        "bar": {"iam_policy_file": "policy-two.json", "autogen_policy": false}
      }
    }
  },
  "version": "2.0",
  "app_name": "test-deletes"
}

$ ls -la .chalice/
total 24
Dec 20 15:34 .
Dec 20 15:32 ..
Dec 20 15:33 config.json
Dec 20 15:33 policy-one.json
Dec 20 15:33 policy-two.json
```

Now the initial deploy will create 4 resources:

```
$ chalice deploy-new
Creating deployment package.
(chalice) /tmp/test-deletes $ cat .chalice/deployed.json
{
  "stages": {
    "dev": {
      "resources": [
        {
          "role_arn": "arn:aws:iam::1:role/test-deletes-dev-foo",
          "name": "role-foo",
          "resource_type": "iam_role"
        },
        {
          "lambda_arn": "arn:aws:lambda:us-west-2:1:function:test-deletes-dev-foo",
          "name": "foo",
          "resource_type": "lambda_function"
        },
        {
          "role_arn": "arn:aws:iam::1:role/test-deletes-dev-bar",
          "name": "role-bar",
          "resource_type": "iam_role"
        },
        {
          "lambda_arn": "arn:aws:lambda:us-west-2:1:function:test-deletes-dev-bar",
          "name": "bar",
          "resource_type": "lambda_function"
        }
      ]
    }
  },
  "schema_version": "2.0"
}
```

We can verify the functions are created remotely:

```
$ aws lambda list-functions --query 'Functions[?starts_with(@.FunctionName, `test-deletes`)].FunctionName'
[
    "test-deletes-dev-foo",
    "test-deletes-dev-bar"
]
```

Same for roles:

```
$ aws iam list-roles --query 'Roles[?starts_with(@.RoleName, `test-deletes`)].RoleName'
[
    "test-deletes-dev-bar",
    "test-deletes-dev-foo"
]
```

And finally, we can delete the app:

```
$ chalice delete-new
(chalice) /tmp/test-deletes $ cat .chalice/deployed.json
{
  "stages": {
    "dev": {
      "resources": []
    }
  },
  "schema_version": "2.0"
}
```

And the resources don't exist remotely:

```
(chalice) /tmp/test-deletes $ aws lambda list-functions --query 'Functions[?starts_with(@.FunctionName, `test-deletes`)].FunctionName'
[]
(chalice) /tmp/test-deletes $ aws iam list-roles --query 'Roles[?starts_with(@.RoleName, `test-deletes`)].RoleName'
[]
```

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -182,6 +182,9 @@ def deploy_new(ctx, autogen_policy, profile, api_gateway_stage, stage):
 @click.pass_context
 def delete_new(ctx, profile, stage):
     # type: (click.Context, str, str) -> None
+    # TODO: We should consolidate the logic here with
+    # deploy_new, there's similar logic in both functions
+    # that can be shared.
     factory = ctx.obj['factory']  # type: CLIFactory
     factory.profile = profile
     config = factory.create_config_obj(chalice_stage_name=stage)
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -175,6 +175,23 @@ def deploy_new(ctx, autogen_policy, profile, api_gateway_stage, stage):
             config.project_dir, '.chalice', 'deployed.json'))
 
 
+@cli.command('delete-new')
+@click.option('--profile', help='Override profile at deploy time.')
+@click.option('--stage', default=DEFAULT_STAGE_NAME,
+              help='Name of the Chalice stage to delete.')
+@click.pass_context
+def delete_new(ctx, profile, stage):
+    # type: (click.Context, str, str) -> None
+    factory = ctx.obj['factory']  # type: CLIFactory
+    factory.profile = profile
+    config = factory.create_config_obj(chalice_stage_name=stage)
+    session = factory.create_botocore_session()
+    d = factory.create_deletion_deployer(session=session)
+    deployed_values = d.deploy(config, chalice_stage_name=stage)
+    record_deployed_values(deployed_values, os.path.join(
+        config.project_dir, '.chalice', 'deployed.json'))
+
+
 @cli.command('delete')
 @click.option('--profile', help='Override profile at deploy time.')
 @click.option('--stage', default=DEFAULT_STAGE_NAME,
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -62,8 +62,6 @@ class RemoteState(object):
 
 class UnreferencedResourcePlanner(object):
 
-    # Add support for IAM roles
-    # get deps order correct.
     def execute(self, plan, config):
         # type: (List[models.Instruction], Config) -> None
         marked = set(self._mark_resources(plan))
@@ -258,6 +256,16 @@ class PlanStage(object):
             document = json.loads(
                 self._osutils.get_file_contents(resource.filename))
         return document
+
+
+class NoopPlanner(PlanStage):
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def execute(self, resources):
+        # type: (List[models.Model]) -> List[models.Instruction]
+        return []
 
 
 class Variable(object):
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -62,13 +62,17 @@ class RemoteState(object):
 
 class UnreferencedResourcePlanner(object):
 
+    # Add support for IAM roles
+    # get deps order correct.
     def execute(self, plan, config):
         # type: (List[models.Instruction], Config) -> None
-        marked = self._mark_resources(plan)
+        marked = set(self._mark_resources(plan))
         deployed = config.deployed_resources(config.chalice_stage)
         if deployed is not None:
-            deployed_resource_names = deployed.resource_names()
-            remaining = list(set(deployed_resource_names) - set(marked))
+            deployed_resource_names = reversed(deployed.resource_names())
+            remaining = [
+                name for name in deployed_resource_names if name not in marked
+            ]
             self._plan_deletion(plan, remaining, deployed)
 
     def _mark_resources(self, plan):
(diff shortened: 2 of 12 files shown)


chalice/deploy/planner.py
```diff
@@ -93,6 +95,15 @@ def _plan_deletion(self,
                     params={'function_name': resource_values['lambda_arn']},
                 )
                 plan.append(apicall)
+            elif resource_values['resource_type'] == 'iam_role':
+                # TODO: add the role_name to the deployed.json?
```
joguSD: Can you elaborate on this todo?

tests/unit/deploy/test_planner.py
```diff
@@ -365,6 +365,15 @@ def sweeper(self):
     def function_resource(self):
         return create_function_resource('myfunction')
 
+    def one_deployed_lambda_function(name='myfunction', arn='arn'):
+        return {
+            'resources': [{
+                'name': 'myfunction',
```
joguSD: Did you mean to use the `name` and `arn` as values here?

chalice/deploy/planner.py
kyleknap: +1. I think it is because we are parsing the name from the role arn, which is not a great practice as there may be edge cases. If it was included in the `deployed.json`, we would not have to worry about including this logic.

tests/unit/deploy/test_planner.py
kyleknap: Yeah the `name` variable needs to be used here.

chalice/cli/__init__.py
```diff
@@ -175,6 +175,23 @@ def deploy_new(ctx, autogen_policy, profile, api_gateway_stage, stage):
...
+@cli.command('delete-new')
+@click.option('--profile', help='Override profile at deploy time.')
+@click.option('--stage', default=DEFAULT_STAGE_NAME,
+              help='Name of the Chalice stage to delete.')
+@click.pass_context
+def delete_new(ctx, profile, stage):
+    # type: (click.Context, str, str) -> None
+    factory = ctx.obj['factory']  # type: CLIFactory
+    factory.profile = profile
+    config = factory.create_config_obj(chalice_stage_name=stage)
```
kyleknap: It may be worth consolidating this logic in the future with `deploy_new()`. Maybe just add a `TODO` for now. Seems to be a decent chunk of duplicated code in this function.

chalice/cli/__init__.py
```diff
@@ -175,6 +175,23 @@ def deploy_new(ctx, autogen_policy, profile, api_gateway_stage, stage):
...
+    session = factory.create_botocore_session()
+    d = factory.create_deletion_deployer(session=session)
+    deployed_values = d.deploy(config, chalice_stage_name=stage)
+    record_deployed_values(deployed_values, os.path.join(
```
kyleknap: Don't think that it needs to be updated in this PR, but ran into it when I was looking through the code, but the annotation for the `record_deployed_values` function looks off: https://github.com/jamesls/chalice/blob/79bd3b83624c80e2e29f51ccf3513c5ea5ca73d7/chalice/utils.py#L60. It is has the deployed values annotated as `Dict[str, str]` when it is really `Dict[str, Any]` now especially since the `deployed_values` contains each stage and a list of resources now.

chalice/deploy/planner.py
jamesls: Yep, I've updated the todo.

tests/unit/deploy/test_planner.py
jamesls: Yeah, good catch.

chalice/cli/__init__.py
jamesls: Interesting, I wonder why mypy didn't fail on this.  I'll update this.

jamesls:
@kyleknap @joguSD Feedback incorporated.

PR #653, opened 2018-01-01
attempt to build a wheel for packages installed from local directories

ewdurbin:
# Use case

I've been working on a project that uses both [apex](https://apex.run) and chalice in a single repository. Many of the packaged lambda functions will require access to a shared library that we don't intend to publish to an index... since the whole sum of the "raw" Lambda functions we're deploying via apex and the Lambda/API Gateway endpoints we're deploying via chalice will be versioned and deployed simultaneously.

# Status of this PR

This is mostly intended as a provisional or "Request for Comment" PR. If the maintainers of Chalice like this approach, I'd be happy to continue work on adding tests and documentation for this :)

Basically the run down is that if the `pip download` step fails, we just blindly try to build a wheel and it will either be picked up if it passes the `_is_compatible_wheel_filename` test, or we'll get an error I think (needs tests).

Again I'm mostly looking for feedback here on if an approach like this is acceptable before working on tests and docs :)

# Example!

```sh
$ tree
├── my-app-http
│   ├── app.py
│   └── requirements.txt
├── my-app-lambda
│   ├── functions
│   │   └── hello
│   │       ├── event.json
│   │       ├── function.json
│   │       ├── lambda.pth
│   │       ├── main.py
│   │       └── requirements.txt
│   └── project.json
└── lib
    ├── my-app
    │   ├── __init__.py
    │   ├── encryption
    │   │   └── __init__.py
    │   └── lambda_helpers
    │       ├── __init__.py
    │       └── kms.py
    ├── setup.py
    └── tests
        ├── __init__.py
        ├── context.py
        ├── encryption
        │   ├── __init__.py
        │   └── test_encryption.py
        └── lambda_helpers
            ├── __init__.py
            └── test_kms.py
```

Here, we have a directory structure containing a chalice application `my-app-http` as well as an apex function named `hello` that lives in an apex project named `my-app-lambda`.

### `my-app-http/requirements.txt`

```
../lib
```

### `my-app-lambda/functions/hello/requirements.txt`

```
../../../lib
```

This PR allows for `chalice deploy` to work in this scenario! Normally it silently passes over the error from `pip download` like

```
$ pip download -r requirements.txt 
Processing /Users/ewdurbin/codez/my-app/lib
  Link is a directory, ignoring download_dir
```

which does _not_ raise a non-zero status code.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -588,8 +588,12 @@ class SubprocessPip(object):
             osutils = OSUtils()
         self._osutils = osutils
 
-    def main(self, args, env_vars=None, shim=None):
-        # type: (List[str], EnvVars, OptStr) -> Tuple[int, Optional[bytes]]
+    def main(self,
+             args,           # List[str]
+             env_vars=None,  # EnvVars
+             shim=None       # OptStr
+             ):
+        # type: (...) -> Tuple[int, OptBytes, OptBytes]
         if env_vars is None:
             env_vars = self._osutils.environ()
         if shim is None:
@@ -601,9 +605,9 @@ class SubprocessPip(object):
         p = subprocess.Popen(invoke_pip,
                              stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                              env=env_vars)
-        _, err = p.communicate()
+        out, err = p.communicate()
         rc = p.returncode
-        return rc, err
+        return rc, err, out
 
 
 class PipRunner(object):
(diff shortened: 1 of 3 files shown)


stealthycoin:
I think this is a cool feature. I wanted to add more ways of plumbing in dependencies since that is one of the hardest things to manage with Python and Lambda. My personal vote is yes to this feature though it should probably have more discussion since it makes a fairly permanent decision.

Thoughts @kyleknap @jamesls?

kyleknap:
Yeah this seems reasonable to me. I also agree we should have more of a discussion about it to make sure we are comfortable with the approach.

PR #655, opened 2018-01-02
Move end of quickstart down to include all the tutorials

stealthycoin:


README.rst
```diff
@@ -1083,6 +1082,7 @@ role associated with your lambda function, so you'll
 need to use an ``AWS_PROFILE`` that has sufficient permissions
 to your AWS resources used in your ``app.py``.
 
+.. quick-start-end
```
jamesls: Can we move this to after the Delete section below?  We're already including more than the quickstart section so we might as well include the "Deleting your App" section.

Issue #316, opened 2017-04-28
auto reload LocalDevServer on module change


aldegoeij:
Hello guys! Love the work on Chalice, makes life with API Gateway much easier.

What framework do you use for the LocalDevServer? Currently it does not auto reload on a change (save) of a module under development, which slows down development quite a bit.

Are you using bare BaseHTTPServer only? Is it easy to implement a reloader?

jamesls:
We're using BaseHTTPServer.  It should be possible to implement a reloader, not sure on how easy it would be, but I think it would be useful to have.  Marking as feature request.

Marcelo-Theodoro:
+1.

It's easy to look for changed files using `os.static` and checking the `st_mtime` attribute.  Like werkzeug[1]. The checking could be done in the `LocalDevServer.serve_forever`  but instead of using the `HTTPServer.serve_forever`, the `LocalDevServer.serve_forever` would use the `HTTPServer.handle_request` inside a loop. Would this change have bad implications?

However I'm not sure of how to reload the application itself when we detect that a file changed. Just re-import everything?


[1] https://github.com/pallets/werkzeug/blob/368ccb07d097660156eb19a0ec2042bc53e68238/werkzeug/_reloader.py#L137

tzookb:
wish I wasnt a python newb and could have helped with this :/

rapidoo:
+1

mlecoq:
+ 1

marcmillien:
While waiting for the feature, you can use `watchdog` and the command `watchmedo`:
```bash
watchmedo \
  auto-restart \
  --patterns="*.py" \
  --ignore-directories \
  --recursive \
  -- chalice local
```

trickidicki:
@marcmillien Alas watchmedo auto-restart doesn't work on Windows... 
https://github.com/gorakhargosh/watchdog/issues/387
:-((

juliensimon:
+1, would love to have this.

cdalar:
+1

jamesls:
Implemented in https://github.com/aws/chalice/pull/846.  This will go out in the next release.

PR #658, opened 2018-01-03
Automatic reload on code change (#316)

piotrowski:
Added werkzeug reloader for automatic restart of local development server on code change.


jamesls:
Thanks for the pull request.  This is a feature I'd like to support but I'd prefer not to take a dependency on werkzeug.  I'd prefer to just take a dependency on a file watcher package with our own code to handle reloading.

I haven't looked at how werkzeug handles this but I imagine there's some tricky edge cases in robustly reloading the entire app.

Issue #378, opened 2017-06-13
Connecting to an RDS from Chalice?


watdeo:
I'm new to chalice and aws in general. What's the best practice when it comes to connecting to Amazon's RDS? I have already set up a postgres RDS through the web panel, so I want my chalice app to be able to retrieve and store data after a GET or POST request from the user.

Here is what I have:

```
from chalice import Chalice
import psycopg2

app = Chalice(app_name='test')
db_user = 'test'
db_pass = 'password'
db_host = 'https://.....'
db_port = 5432

@app.route('/save_data', methods=["POST"])
def save_data():
    with psycopg2.connect(user=db_user, password=db_pass, ...) as conn:
        with conn.cursor() as cur:
            cur.execute("INSERT INTO test (num, data) VALUES (%s, %s)", (1,2))

@app.route('/get_data')
def get_data():
    with psycopg2.connect(user=db_user, password=db_pass, ...) as conn:
        with conn.cursor() as cur:    
            cur.execute("SELECT * FROM test")


@app.route('/something')
def something():
    with psycopg2.connect(user=db_user, password=db_pass, ...) as conn:
        with conn.cursor() as cur:    
            cur.execute(....)  # some other query
```

Is this a valid way of connecting to my RDS instance? Is there an easier way to build/teardown the db connection after a request? Is this safe from concurrency issues across many concurrent requests?

Or, if I'm completely off, what's the proper way to connect to the RDS from Chalice routes?

stealthycoin:
It should be safe to make the connection at the module level and reuse it in all of your routes because any one instance of your lambda function will only every be handling one request at a time, so there will be no contention for your connection object.

 This will speed things up since it won't be creating a new connection for each request, only on fresh cold-starts of the lambda container, which will only happen when concurrent requests come in and need additional containers spun up to handle them.

jamesls:
Let us know if you have any more questions.

aneesha:
Is this still the recommended way to connect to a database?

ilhamfadllah:
@aneesha Hi, newbie here to use Chalice. Not sure this recommended way. Just share how i connect to Aurora serverless w/ Data API enabled

```
import boto3

rdsClient = boto3.client('rds-data')

dbName = ''
dbArn = ''
dbCredentialsSecretStoreArn = ''

def execute_sql_statement(sql):
    response = rdsClient.execute_statement(
        secretArn=dbCredentialsSecretStoreArn,
        database=dbName,
[...]

PR #659, opened 2018-01-04
Add better support for browsers in local mode

### LINKED ISSUE #628: Chalice local not working across browsers/tools
I have the following routes:

```
@app.route("/item/{item_id}", methods=['GET'])
def view_item(item_id):
    return

```
When I run `chalice local`
If I view the page on browser(GET), works fine, but then using POSTMAN to get to the page, I endup with endless wait, but GET on browser still works

restarts `chalice local`

If I use POSTMAN to GET to the page first, it works, then GET page on browser doesn't work, but POSTMAN GET still works.

stealthycoin:
Browsers tend to hold open sockets for future use. In the default single
threaded model that Python SimpleHTTPServer uses this causes the server
to become blocked permanently. This change adds a threading mixin to
the local server to prevent it from becoming blocked.

fixes #628

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Next Release (TBD)
   (`#344 <https://github.com/aws/chalice/issues/344>`__)
 * Fix IAM role creation issue
   (`#565 <https://github.com/aws/chalice/issues/565>`__)
+* Fix `chalice local` handling of browser requests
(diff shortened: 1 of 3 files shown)


dflanzer-s1:
This fix much improved local browser support but it did cause some side-effects with using an RDS connection at the module level (as recommended in #378 ) , due to cursors not being thread-safe.  Workaround might be to detect local mode and implement single connections per route request. Other thoughts?

Issue #672, opened 2018-01-10
API Gateway Compression


nplutt:
Now that [compression](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gzip-compression-decompression.html) has been released for api gateway, it would be great if there was some way configure it from Chalice. My proposal is to add an optional configuration parameter of `minimum_compression_size` to each stage in the `config.json` file. This would look something like:
```javascript
{
  "version": "2.0",
  "app_name": "compression-poc"
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "minimum_compression_size": 0,
      "lambda_memory_size": 1536
    }
  }
}
```

kyleknap:
Marking as a feature request. I think the configuration would make sense.

stealthycoin:
Closing since #676 was merged.

PR #676, opened 2018-01-11
API Gateway Compression

nplutt:
Completes feature for issue [#672](https://github.com/aws/chalice/issues/672).

I was disappointed that there is no way to do this using the [API Gateway Extensions to Swagger](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions.html), so I had to use the [update_rest_api](http://boto3.readthedocs.io/en/latest/reference/services/apigateway.html#APIGateway.Client.update_rest_api) method and set the compression level after the API Gateway is deployed. 

The new functionality can is enabled by using  an optional configuration parameter of `minimum_compression_size` in the `config.json` file. The parameter can be set at either the global level or the stage level and if both exist the stage level overrides the global level. Below is an example:
```javascript
{
  "minimum_compression_size": 0,
  "stages": {
    "dev": {
      "api_gateway_stage": "api"
    },
    "test": {
      "api_gateway_stage": "test",
      "minimum_compression_size": 1000
    }
  },
  "version": "2.0",
  "app_name": "compression"
}
```
In this example `dev` would have a `minimum_compression_size` of 0 and `test` would have a `minimum_compression_size` of 1000.

Let me know if you would like anything changed, thanks!


nplutt:
Hey, just wanted to follow up on this. Let me know if you see any issues or want anything else added/changed.

Thanks!

nplutt:
Please wait on this PR, I just saw that the API gateway swagger now supports compression with a value of `x-amazon-apigateway-minimum-compression-size` and will be updating the PR accordingly.

nplutt:
Hey, just wanted to follow up on this. Let me know if you see any issues or want anything else added/changed.

Thanks!

pickfire:
Any news on this?

gallmerci:
@nplutt I tried to reproduce your problem but for me it is working everything as expected. Maybe they fixed it?
Another comment: Can you also add the compression configuration to the swagger definition in the SAM file generated by `chalice deploy`? Or is this directly updated with your current implementation?

nplutt:
Thanks for the feedback @gallmerci. When I created this it was less than a week after they released the compression feature to API Gateway. I'm happy to hear that the issue that I was running into was fixed.

gallmerci:
@nplutt I would love to be able to do this, but unfortunately, I am not :( 
I can just loudly vote for this idea!

wutachih:
Is how the compression handled documented? I need to post a large json from an ETL process. The compression should help.

stealthycoin:
@nplutt Thanks for the submission. I rebased this off of the current state of the master branch and updated it to use the new deployer. Added support for cloudformation as well.

docs/source/topics/configfile.rst
```diff
@@ -45,6 +45,12 @@ be checked.
   will also be the URL prefix for your API
   (``https://endpoint/prefix/your-api``).
 
+* ``minimum_compression_size`` - An integer value that indicates
+  the minimum compression size to apply to the API gateway. If
+  this key is specified in both a stage specific config option
+  as well as a top level key, the stage specific key will
+  override the top level key for the given stage.
```
kyleknap: It might be worth linking the service docs for this as well (e.g. https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gzip-compression-decompression.html)

chalice/deploy/validate.py
```diff
@@ -146,6 +149,18 @@ def _validate_cors_for_route(route_url, route_methods):
             )
 
 
+def validate_minimum_compression_size(config):
+    # type: (Config) -> None
+    if config.minimum_compression_size is not None:
```
JordonPhillips: Not super important, but it would be nice to reduce the indentation here

Issue #481, opened 2017-08-17
Support encrypting environment variables with KMS


boosh:
I need to store sensitive data (e.g. passwords) for use by lambda. I'm currently planning on doing this by using a custom config file, but I see that Lambda supports KMS. Ideally I'm looking for a way that allows me to locally encrypt this data as well so I can safely store everything in my git repo (for example how ansible has a command to encrypt secrets using `ansible vault`) without adding extra tooling.

I plan on using separate accounts per stage to fully segregate my resources, so it'd useful if there was support for `config-<stage>.json` files as well. Doing that could allow these stage-specific config files to be encrypted separately per stage. It could require a password to decrypt the config for whichever stage while running/deploying.

Have you given any thought to how to handle encrypted config, and then push it up to lambda so it's encrypted there as well?

jamesls:
The built in support for KMS with Lambda is to encrypt your env vars at rest.  Lambda would automatically handle the decryption before invoking your function for you as described [here](http://docs.aws.amazon.com/lambda/latest/dg/env_variables.html#env_encrypt).

We've discussed adding support for KMS in chalice so the process is more seamless, but it is possible to do this today, it's just more manual.


If you wanted the values encrypted client side you can follow a similar process to the "encryption helpers" that are in the AWS Lambda console.  The process would be:

1. Create a KMS key if you don't already have one
2. Encrypt your variables via KMS:

```
$ aws kms encrypt --key-id $KEY_ID --plaintext secretpassword --output text --query CiphertextBlob
AQICAHjZ+JlI8KKmiVc++NhnBcO0xX3LFAaCfsjH8Yjig3Yr2AFPIyKCp3XVOEDlbiTMWiejAAAAbDBqBgkqhkiG9w0BBwagXTBbAgEAMFYGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMhai9vkA2KdU5gd+qAgEQgCnWW4F3fb7pTwmA2ppskJhUl0dJGEXIE5oDCr3ZsH7TlN5X381juPg0LA==
```

3. That encrypted value you'd store in your config file.  You can have environment variables specified per stage:

```
{
  "version": "2.0",
  "app_name": "app",
  "stages": {
    "dev": {
      "environment_variables": {
[...]

PR #677, opened 2018-01-11
[WIP] add option to encrypt env variables in config

pkit:
adds new command `encrypt-config` that encrypts all the env vars using thep provided kms key

#481


pkit:
Work in progress, needs tests.
Comments?

pkit:
"Load it from there" means infrastructure code in lambda.
I'm not sure I want that.
After all I can always run full web server in lambda and then it will be no different than EC2.

skwashd:
I'd prefer to have 2 extra lines of code than to have to mess around with encrypting stuff on the command line and then copying and pasting that into my config.json

```python
ssm = boto3.client('ssm', region_name='us-east-1')
params = ssm.get_parameter(Name='my-key',  WithDecryption=True)['Parameters']['Value']
```

pkit:
You will need another script to store these parameters, CI/CD to set it up, etc., etc.

mvoitko:
Please merge.

pkit:
Rebased, in case somebody still want to merge it

Issue #601, opened 2017-11-08
CTRL+C takes too long to terminate "chalice local" process


byrro:
On Windows 8.1 command line:

When I run:
`chalice local --port=8080`

If I immediately press `CTRL+C`, it interrupts the script right after.

When I run:
`chalice local --port=8080`
Then open http://localhost:8080/ on my browser and after that press `CTRL+C` on the command prompt, the script only gets interrupted after 1-2 minutes. The only way to interrupt it quickly is killing the Python process.

Expected behavior:
Have the script run by `chalice local` terminated immediately after pressing `CTRL+C`, regardless of having made an HTTP request or not before that.

I'm experiencing the issue with the windows 8.1 command prompt. Same issue happens when using Console2 and ConEmu.

joguSD:
I've reproduced this on windows 10. Closing the browser (tried both firefox and IE) seems to allow `chalice local` to close. Not sure if this is an issue with `HTTPBaseServer` on windows or an issue with our `ChaliceRequestHandler`.

wollerman:
Potentially related is #628 where the browser seems to interfere with other requests.

noelmccrory:
This happens to me also when running chalice local from a Powershell terminal inside VS Code (Win 10). As a workaround, I typically add a new terminal, run chalice local, test, then kill the terminal when I'm finished. CTRL-C seems to have no effect at all.

byrro:
@noelmccrory I'm using [ConEmu](https://conemu.github.io/). Instead of killing the entire terminal, you can configure a shortcut in ConEmu to kill just the Python process within the terminal. Instead of using CTRL+C, I configured CTRL+Q to kill Python and quit the Chalice local process without having to close the Terminal window every time.

Open ConEmu settings, go to "Keys & Macro" and search for this shortcut: "Terminate (kill) active process in the current console". Then set the key combination of your preference.

variable:
Just had the same problem

leason:
On Windows 10 you can `ctrl+c`, then just send a request to chalice.  The interrupt isn't picked up until another request is sent.  That causes the local process to quit, and then you can restart it.  I don't know how to implement it, but having an automatic reload when files change similar to how React does this when you run `npm start` would be awesome.

byrro:
[...]

PR #675, opened 2018-01-11
Shutdown 'chalice local' with ctrl+c

atagar:
When running a HTTPServer keyboard interrupts aren't properly shutting it
down: https://github.com/aws/chalice/issues/601

Users reported this for Windows, and I'm seeing something similar on Ubuntu. In
particular...

  * Run 'chalice local' with the hello world demo app.

  * In a browser issue a request to exercise it. This is important - the
    HTTPServer must have handled a request for the issue to manifest.

  * Hit 'ctrl+c'. You'll see a stacktrace but the server will continue.

  * Hit 'ctrl+c' a second time. This will get it to terminate for realz.

Here's what it looks like...

```
% chalice local
Serving on 127.0.0.1:8000
127.0.0.1 - - [10/Jan/2018 17:14:18] "GET / HTTP/1.1" 200 -
^C----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 46352)
Traceback (most recent call last):
  File "/usr/lib/python2.7/SocketServer.py", line 290, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 318, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/local/ANT/damiajoh/demo/venv/local/lib/python2.7/site-packages/chalice/local.py", line 529, in __init__
    self, request, client_address, server)  # type: ignore
  File "/usr/lib/python2.7/SocketServer.py", line 652, in __init__
    self.handle()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 342, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 310, in handle_one_request
    self.raw_requestline = self.rfile.readline(65537)
  File "/usr/lib/python2.7/socket.py", line 480, in readline
    data = self._sock.recv(self._rbufsize)
KeyboardInterrupt
----------------------------------------
^C
Aborted!
```

Last line shows that the ChaliceRequestHandler constructor blocks on reading
the socket when idle. Tried quite a few things (timeouts, manually closing the
socket, handle_error override, etc) but in the end this was the nicest kind of
shutdown I could come up with. From a user perspective keyboard interrupts now
end the server right away without a stacktrace.

Tips welcome. I fiddled with this for quite a few hours attempting to include
[...]

PR #680, opened 2018-01-11
Shut down 'chalice local' with ctrl+c

atagar:
Interesting. Our ThreadingMixIn actually makes shutdown a bit easier than our
previous HTTPServer.

If our local server has handled a request then ThreadingMixIn will have a
spawned thread waiting on socket.recv. To avoid hanging flagging these as
daemon threads so they don't block interpreter shutdown when serve_forever()
is done.

For the previous patch for shutting down a single threaded HTTPServer see
https://github.com/aws/chalice/pull/675.

This should fix https://github.com/aws/chalice/issues/601

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -606,6 +606,8 @@ class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
     the browswer will simply open another one and sit on it.
     """
 
+    daemon_threads = True
+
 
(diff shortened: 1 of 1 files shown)


stealthycoin:
Thanks for the PR! Didn't seem worthwhile to bother you to add a test so I threw one in and opened a new PR with your commit: #687.

Issue #574, opened 2017-10-20
More complex Swagger ability?


noahgift:
Hi,

Do you have any plans to support more complex Swagger configurations, something like this:  https://github.com/rochacbruno/flasgger, etc?  If not, how are people currently supplementing their swagger api docs.

Thanks!

JordonPhillips:
I think it would be a really good idea to extract python doc strings as apigateway docs. That would make them immediately available in the swagger exports, in addition to any other apigateway features. So the following code:

```python
from chalice import Chalice

app = Chalice(app_name="documentation-as-a-service")


@app.route("/")
def index():
    """This doc string would be added to apigateway."""
    return {'hello': 'world'}
```

Would automatically have the documentation for the base route set as "This doc string would be added to apigateway."

Is that the sort of thing you're looking for?

noahgift:
Exactly!  Almost considering writing it myself, but didn't want to have a fork that conflicted.

RobertAtomic:
@JordonPhillips Do you guys accept PRs for stuff like this?  We're just getting started with a chalice based project and we're trying to decide if we implement a sidecar process to generate more robust swagger, or if we fork chalice and implement there.  We don't like the fork option unless there's some expectation that it might get merged.

JordonPhillips:
@RobertAtomic we sure do!

dmulter:
I've added the following code that I believe should handle adding Summary and Description in the swagger output (deploy/swagger.py). Unfortunately I can't seem to figure out how to get these lines to actually execute, so I can't verify the code. It doesn't seem to walk all paths every time I do a deploy. Hopefully someone can take the next step based on the code below. The 5 new lines are right after initialization of the base current dict.

```python

    def _generate_route_method(self, view):
        # type: (RouteEntry) -> Dict[str, Any]
        current = {
            'consumes': view.content_types,
            'produces': ['application/json'],
            'responses': self._generate_precanned_responses(),
            'x-amazon-apigateway-integration': self._generate_apig_integ(
                view),
        }  # type: Dict[str, Any]
        if view.view_function.__doc__:
            doc_lines = view.view_function.__doc__.splitlines()
            current['summary'] = doc_lines[0]
            if doc_lines[1]:
                current['description'] = '\n'.join(doc_lines[1:])
        if view.api_key_required:
            # When this happens we also have to add the relevant portions
            # to the security definitions.  We have to someone indicate
            # this because this neeeds to be added to the global config
            # file.
[...]

PR #681, opened 2018-01-11
Add Swagger output for method summary and description fields from docstrings

dmulter:
PR for #574

### DIFF
--- a/chalice/deploy/swagger.py
+++ b/chalice/deploy/swagger.py
@@ -1,4 +1,5 @@
 import copy
+import inspect
 
 from typing import Any, List, Dict, Optional  # noqa
 
@@ -128,6 +129,11 @@ class SwaggerGenerator(object):
             'x-amazon-apigateway-integration': self._generate_apig_integ(
                 view),
         }  # type: Dict[str, Any]
+        if view.view_function.__doc__:
+            doc_lines = inspect.getdoc(view.view_function).splitlines()
+            current['summary'] = doc_lines[0]
+            if len(doc_lines) > 1:
+                current['description'] = '\n'.join(doc_lines[1:]).strip('\n')
         if view.api_key_required:
             # When this happens we also have to add the relevant portions
             # to the security definitions.  We have to someone indicate
--- a/chalice/deploy/swagger.py
+++ b/chalice/deploy/swagger.py
@@ -129,8 +129,9 @@ class SwaggerGenerator(object):
             'x-amazon-apigateway-integration': self._generate_apig_integ(
                 view),
         }  # type: Dict[str, Any]
-        if view.view_function.__doc__:
-            doc_lines = inspect.getdoc(view.view_function).splitlines()
+        docstring = inspect.getdoc(view.view_function)
+        if docstring:
+            doc_lines = docstring.splitlines()
             current['summary'] = doc_lines[0]
             if len(doc_lines) > 1:
                 current['description'] = '\n'.join(doc_lines[1:]).strip('\n')
--- a/tests/unit/deploy/test_swagger.py
+++ b/tests/unit/deploy/test_swagger.py
@@ -41,6 +41,96 @@ def test_can_produce_doc_for_method(sample_app, swagger_gen):
     # Same for x-amazon-apigateway-integration.
 
 
+def test_can_produce_doc_for_no_docstring(sample_app, swagger_gen):
+    @sample_app.route('/method')
+    def method():
+        pass
+
+    doc = swagger_gen.generate_swagger(sample_app)
+    view_config = doc['paths']['/method']['get']
+    assert 'summary' not in view_config
+    assert 'description' not in view_config
+
+
+def test_can_produce_doc_for_single_docstring(sample_app, swagger_gen):
+    @sample_app.route('/method1')
+    def method1():
+        """Single line method summary"""
+        pass
+
+    @sample_app.route('/method2')
+    def method2():
+        '''Single line method summary'''
+        pass
+
+    @sample_app.route('/method3')
+    def method3():
+        """
+        Single line method summary
+        """
+        pass
+
+    doc = swagger_gen.generate_swagger(sample_app)
+    for method in [1, 2, 3]:
+        view_config = doc['paths']['/method' + str(method)]['get']
+        assert 'summary' in view_config
+        assert 'description' not in view_config
+        assert view_config['summary'] == 'Single line method summary'
+
+
+def test_can_produce_doc_for_multi_docstring(sample_app, swagger_gen):
+    @sample_app.route('/method1')
+    def method1():
+        """Multiline method summary
+
+        And here is a more detailed description that can span multiple
+        lines. It can also handle indenting for things like method
+        arguments like so:
+            param1 - description
+            param2 - description
+        """
+        pass
+
+    @sample_app.route('/method2')
+    def method2():
+        """Multiline method summary
+
+            And here is a more detailed description that can span multiple
+            lines. It can also handle indenting for things like method
+            arguments like so:
+                param1 - description
+                param2 - description
+        """
+        pass
+
+    @sample_app.route('/method3')
+    def method3():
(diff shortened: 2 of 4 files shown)


dmulter:
any chance this can make into a release?

chalice/deploy/swagger.py
```diff
@@ -128,6 +129,11 @@ def _generate_route_method(self, view):
             'x-amazon-apigateway-integration': self._generate_apig_integ(
                 view),
         }  # type: Dict[str, Any]
+        if view.view_function.__doc__:
+            doc_lines = inspect.getdoc(view.view_function).splitlines()
+            current['summary'] = doc_lines[0]
+            if len(doc_lines) > 1:
```
stealthycoin: `if doc_lines:` is more succinct.

tests/unit/deploy/test_swagger.py
```diff
@@ -41,6 +41,96 @@ def test_can_produce_doc_for_method(sample_app, swagger_gen):
...
+        lines. It can also handle indenting for things like method
+        arguments like so:
+            param1 - description
+            param2 - description
+
+
+        """
+        pass
+
+    doc = swagger_gen.generate_swagger(sample_app)
```
stealthycoin: Could we add a test that actually deploys some documentation and then uses the API to confirm that it is set correctly?

tests/unit/deploy/test_swagger.py
```diff
@@ -41,6 +41,96 @@ def test_can_produce_doc_for_method(sample_app, swagger_gen):
...
+
+    @sample_app.route('/method3')
+    def method3():
+        """
+        Single line method summary
+        """
+        pass
+
+    doc = swagger_gen.generate_swagger(sample_app)
+    for method in [1, 2, 3]:
```
stealthycoin: How about parameterizing these? https://docs.pytest.org/en/latest/example/parametrize.html

chalice/deploy/swagger.py
```diff
@@ -128,6 +129,11 @@ def _generate_route_method(self, view):
             'x-amazon-apigateway-integration': self._generate_apig_integ(
                 view),
         }  # type: Dict[str, Any]
+        if view.view_function.__doc__:
+            doc_lines = inspect.getdoc(view.view_function).splitlines()
```
stealthycoin: How come we use `__doc__` in the if statement and `inspect.getdoc(...)`to get the actual docs? No need to import inspect if we just use `__doc__` in both places. It looks like you manually clean up the string anyway so you aren't really getting any benefit from the fact that `getdoc` calls `cleandoc` as far as I can tell.

chalice/deploy/swagger.py
dmulter: but I'm looking for > 1, not > 0 right?

tests/unit/deploy/test_swagger.py
dmulter: Great idea, but I'm not sure how to do that in the framework of your tests. I tried to follow the model of tests that were already there. If you could point me to an example in the code that would help a lot.

tests/unit/deploy/test_swagger.py
dmulter: yes, working on it...

chalice/deploy/swagger.py
dmulter: I'm distinguishing single line docstrings from multiline.

chalice/deploy/swagger.py
dmulter: I use ```__doc__``` because it's a quick check if a docstring exists at all. If it does exist, I need to get the docstring with the indention at the function level. I saw that the ```getdoc()``` version returns corrected indentation. I'm open to other ways to get it, but I found that ```getdoc()``` seemed to be the best approach.

tests/unit/deploy/test_swagger.py
dmulter: So I've used parameterized tests before and love them for static data sets, and tried refactoring the code for this case. I think it reads much cleaner as it is though. If I move to parameterized, then each of the methods get defined outside the test, then I have the block of function data mapped to a ```method``` variable. In the end I thought it was less readable, though you do get cleaner reporting if one of the sub-tests fail.

chalice/deploy/swagger.py
stealthycoin: My mistake.

tests/unit/deploy/test_swagger.py
stealthycoin: The integration tests `test_features.py` should be what you want. It deploys `test/integration/testapp/` and then uses various API methods to determine if it was deployed correctly.

 So you can add a new method to that app that has a documentation string and then write a test similar to 

```python
def test_path_params_mapped_in_api(smoke_test_app, apig_client):
```

Which gives you the deployed app and a client to interact with the APIG API.

chalice/deploy/swagger.py
stealthycoin: How about just:

```python
doc_lines = inspect.getdoc(....)
if doc_lines:
    ...
```
Reads cleaner to me and I'm not worried about the small performance difference during a deployment.

tests/unit/deploy/test_swagger.py
stealthycoin: It was mainly the cleaner reporting that I was going for. I was thinking these would be clearer as separate tests but didn't want the repeated logic since they were all so separate.

chalice/deploy/swagger.py
dmulter: sounds good

tests/unit/deploy/test_swagger.py
dmulter: I could still break into separate tests if you prefer, but I think the parameterization didn't look right to me.

tests/unit/deploy/test_swagger.py
dmulter: I'll check that out.

chalice/deploy/swagger.py
dmulter: np ;-)

chalice/deploy/swagger.py
dmulter: Just want to confirm again that this looks better to you:

```python
        docstring = inspect.getdoc(view.view_function)
        if docstring:
            doc_lines = docstring.splitlines()
            current['summary'] = doc_lines[0]
            if len(doc_lines) > 1:
                current['description'] = '\n'.join(doc_lines[1:]).strip('\n')
```

tests/unit/deploy/test_swagger.py
dmulter: Got it all in, and happy with the results. Sorry for the lint fails that I should have checked first.

dmulter:
I hope this is good to go now...

dmulter:
Please let me know if you require further changes before approval. I'm not sure if I should just be patient, or if you are expecting something more. Thanks.

dmulter:
I'm still waiting to hear on my PR. I think it's good to go, but somehow has fallen into a black hole I think.

sbassi:
I am also interested in this.

tests/unit/deploy/test_swagger.py
dmulter: Is this what is holding up the PR approval? I will complete this work if you want, just want to be sure that's all.

dmulter:
Still happy to do whatever it takes to get this merged...

jamesls:
Looks good to me.  Sorry about the delay on this.

jamesls:
Rebased off of master and merged via 482bd7cace01fdc487ab333fafe814c117ba7da9

dperconti:
Thanks!!

careem-eng-mobile:
Nice feature! Do we have this feature documented somewhere with the examples?

qalandarov:
Nice feature! Do we have this feature documented somewhere with the examples?

saivarunk:
Do we have any documentation for writing docstrings?

PR #683, opened 2018-01-12
Add request context to 'chalice local' from 'local-request-context' header

bohui:
In local test I didn't find a proper way to mocker context variables. With a new header of 'local-request-context', we can mock any requestContext. This is an example of how to mock identity/cognitoIdentityId by header.
```
curl -X GET \
  http://localhost:8080/path \
  -H 'content-type: application/json' \
  -H 'host: localhost:8080' \
  -H 'local-request-context: {"identity":{"cognitoIdentityId":"test"}}' 
```

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -165,7 +165,7 @@ class LambdaEventConverter(object):
         return headers.get('content-type', '') in self._binary_types
 
     def _get_local_context(self, headers):
-        # type: (Dict[str,str]) -> Any
+        # type: (Dict[str,Any]) -> Any
         local_context = headers.pop('local-request-context', "")
         if not local_context:
             return None


chalice/local.py
```diff
@@ -161,6 +162,22 @@ def _is_binary(self, headers):
         # type: (Dict[str,Any]) -> bool
         return headers.get('content-type', '') in self._binary_types
 
+    def _get_local_context(self, headers):
+        # type: (Dict[str,Any]) -> bool
```
skwashd: This is causing the tests to fail. Looking at your changes in here and in `create_lambda_event()` it seems you should be using `# type: (Dict[str, str])) -> Dict[str, str]`.

The `headers` passed into `create_lambda_event()` are `Dict[str, str])`. You are returning a collection that can be used to replace the existing `headers`.

stealthycoin:
Seems like #695 would do what you need as well. Still an interesting idea to have generic way of injecting any extra info into the lambda context. Especially since keeping local mode in sync with API Gateway and Lambda is a moving target. 

Are you still interested in trying to drive this feature through @bohui?

PR #698, opened 2018-01-31
Added error checking for invalid JSON bodies

johntmyers:
Hi,

I've found that using `app.current_request.json_body` is obviously supposed to be a nice handy way to just access JSON requests, but it does not provide a way to validate that the body was actually proper JSON.

In order to do this, you'd have to do your own check on `app.current_request.raw_body` and then handle however you'd like. This seems to defeat the purpose of using the `app.current_request.json_body` property. I've added a simple check that returns `None` if the JSON is not able to be loaded.


stealthycoin:
Can you tell me about the use case for this? Silencing an error and replacing it with `None` is not something we would want to do, especially in a framework. I would expect that if the `content-type` says the payload is JSON and it doesn't parse as JSON that I would get an error back which we is currently what would happen. And if you want to handle it in a different way you can use `raw_data` as you indicated.

johntmyers:
> it doesn't parse as JSON that I would get an error back which we is currently what would happen.

You get an unhandled `ValueError` and the app crashes yielding a 500, I think this should be handled more gracefully. So the use case is maybe more return a BadRequestError b/c the JSON is invalid.

stealthycoin:
I agree a `400` and maybe something like `Error parsing JSON` would make more sense. I don't have a problem with changing the error message and  the code to make more sense.

johntmyers:
Cool -- I updated the PR if you want to take a look.

johntmyers:
Any issues with this?

jamesls:
Merged in 05b9ef2

PR #706, opened 2018-02-08
Auto reload for local mode

vtitor:
I have implemented auto reload feature by #316 request. It doesn't have any dependencies (according to #658 remarks). 
@jamesls, @stealthycoin Could you help me merge this PR as soon as possible, please?


vtitor:
@jamesls, @stealthycoin Hi guys! If I have some mistakes in this pr could you please tell me about them

.pylintrc
```diff
@@ -304,7 +304,7 @@ valid-metaclass-classmethod-first-arg=mcs
 
 # List of member names, which should be excluded from the protected access
 # warning.
-exclude-protected=_asdict,_fields,_replace,_source,_make
```
stealthycoin: I would prefer not to reach into the `_thread` if at all possible. Are there other options?

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+                continue
+            old_time = self.mtimes.setdefault(path, mtime)
+            if mtime > old_time:
+                return True
+        return False
+
+    def reload(self):
+        # type: () -> None
+        self.triggered = True
+        if sys.platform == 'win32':
```
stealthycoin: We prefer to have all platform specific logic in the `compat.py` file in one place so we don't need to think about it here.

chalice/cli/__init__.py
```diff
@@ -79,11 +80,14 @@ def cli(ctx, project_dir, debug=False):
 @click.option('--port', default=8000, type=click.INT)
 @click.option('--stage', default=DEFAULT_STAGE_NAME,
               help='Name of the Chalice stage for the local server to use.')
+@click.option('--autoreload', is_flag=True)
```
stealthycoin: Seems like it should be turned on by default so people don't need to discover the functionality, it just magically works. We could flip it and let them disable it if they want to for some reason. `--no-autoreload`. Thoughts?

chalice/cli/__init__.py
```diff
@@ -79,11 +80,14 @@ def cli(ctx, project_dir, debug=False):
...
 @click.pass_context
-def local(ctx, host='127.0.0.1', port=8000, stage=DEFAULT_STAGE_NAME):
-    # type: (click.Context, str, int, str) -> None
+def local(ctx, host='127.0.0.1', port=8000, stage=DEFAULT_STAGE_NAME,
+          autoreload=False):
+    # type: (click.Context, str, int, str, bool) -> None
     factory = ctx.obj['factory']  # type: CLIFactory
-    run_local_server(factory, host, port, stage, os.environ)
+    with Reloader(autoreload):
```
stealthycoin: Seems a little bit strange to me to pass `False` to class called Reloader. Could we refactor this to just not use the Reloader context manager if its disabled?

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+                 traceback  # type: Optional[types.TracebackType]
+                 ):
+        # type: (...) -> None
+        if exc_type is KeyboardInterrupt and self.triggered:
+            sys.exit(1)
+
+    def run(self):
+        # type: () -> None
+        while True:
+            time.sleep(AUTORELOAD_INTERVAL)
```
stealthycoin: This would be nice as a parameter that can be supplied to the reloader.

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+            if self.find_changes():
+                self.reload()
+
+    def find_changes(self):
+        # type: () -> bool
+        for module in list(sys.modules.values()):
```
stealthycoin: * No need to cast to `list` here. 
* Can we scope this to just the files that are in their Chalice project?

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+        return False
+
+    def reload(self):
+        # type: () -> None
+        self.triggered = True
+        if sys.platform == 'win32':
+            subprocess.Popen(sys.argv, close_fds=True)
+            six.moves._thread.interrupt_main()
+        else:
+            os.execv(sys.executable, [sys.executable] + sys.argv)
```
stealthycoin: I haven't had the chance to try out this branch yet but I'm curious about this approach and how it compares to managing a subprocess yourself, or hot reloading the python code. For example:
https://docs.python.org/3.4/library/importlib.html#importlib.reload
https://docs.python.org/3.2/library/imp.html#imp.load_module
etc...

I'm mainly trying to think of a more testable approach that would not need any platform specific logic. Would like to hear your thoughts on why you chose this approach.

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+
+from chalice.constants import AUTORELOAD_INTERVAL
+
+
+class Reloader(threading.Thread):
+    def __init__(self, autoreload=True):
+        # type: (bool) -> None
+        super(Reloader, self).__init__()
+        self.autoreload = autoreload
+        self.triggered = False
```
stealthycoin: I recognize that this is whether or not the process has ever reloaded. But can you give some clarification on why this is important to keep track of?

chalice/cli/reload.py
```diff
@@ -0,0 +1,71 @@
...
+            if mtime > old_time:
+                return True
+        return False
+
+    def reload(self):
+        # type: () -> None
+        self.triggered = True
+        if sys.platform == 'win32':
+            subprocess.Popen(sys.argv, close_fds=True)
+            six.moves._thread.interrupt_main()
```
stealthycoin: Can you explain why this is needed on windows?

.pylintrc
vtitor: No, there are no other options.

chalice/cli/reload.py
vtitor: Sure, I'll fix this.

chalice/cli/__init__.py
vtitor: It makes sense.

chalice/cli/__init__.py
vtitor: You are right, I'll fix it.

chalice/cli/reload.py
vtitor: Yep.

chalice/cli/reload.py
vtitor: > No need to cast to ``list`` here.

Why?

>  Can we scope this to just the files that are in their Chalice project?

We can.. but I think it's better to check all modules.

chalice/cli/reload.py
vtitor: Reload python module at runtime looks like a bad idea to me (because it causes inconsistently application state)

chalice/cli/reload.py
vtitor: We need to have additional behavior when interrupting for the reloaded process in windows (raise SystemExit).

chalice/cli/reload.py
vtitor: ``execv`` in windows is completely different than in linux. Briefly, it's necessary to close the opened file descriptors.

vtitor:
@jamesls, @stealthycoin Could you review my changes, please?

jamesls:
Taking a look now.  I haven't had a chance to look at this PR in general (thanks @stealthycoin for the feedback so far), but I think the feature itself would be good to have in chalice.

Was there a discussion about pulling in some existing library that can watch for file changes?  Seems like it would be preferable to use something based on inotify/fsevents instead of polling for changes.

jamesls:
Hi, I know there's been a lot of requests for this feature so I'll go ahead and pull in these changes and make a few adjustments before merging.

jamesls:
Merged via https://github.com/aws/chalice/pull/846.  This will go out in the next release.  Thanks again for the original PR.

Issue #630, opened 2017-12-05
Request: Route Checking and Response


wollerman:
It's extremely easy setup routes. However, this means it's also easy to make a mistake. The current error message on a bad route is ambiguous and doesn't seem to relate to the missing route. For example, given the below route without the leading `/`:

```
@app.route('status', cors=True)
def index():
    return {'status': 'available'}
```
A request returns:
``` 
> http localhost:8000/status
HTTP/1.1 403 Forbidden
Content-Length: 43
Content-Type: application/json
Date: Tue, 05 Dec 2017 18:06:30 GMT
Server: BaseHTTP/0.3 Python/2.7.12
x-amzn-ErrorType: UnauthorizedException
x-amzn-RequestId: 52e78aea-aaf5-455e-8f6f-fece9ceffc80

{
    "message": "Missing Authentication Token"
}
```

It would make more sense to have information about a bad route. Or perhaps providing a console warning about a request to a non-existent route. Currently the console only shows a `403`:
```
> chalice local
Serving on localhost:8000
127.0.0.1 - - [05/Dec/2017 10:06:30] "GET /status HTTP/1.1" 403 -  # response when above request is made
```

jamesls:
Thanks for the feedback, marking as an enhancement.

leason:
Is anyone working on this?  If not I'm happy to open a PR and try to share a patch for it.  No idea how complex the fix is yet, but I get bit by this a fair amount since my team is divided into frontend and backend developers and sometimes they fat finger a route.  This error causes confusion because it looks like a valid route was requested.

wollerman:
@leason: I would wager no one is currently working on it. 

Looking into the `local.py` [it seems like match_route()](https://github.com/aws/chalice/blob/master/chalice/local.py#L112-L145) is throwing a nice message about no matching route. However, that boils up into a [_generate_lambda_event()](https://github.com/aws/chalice/blob/master/chalice/local.py#L433) call which [stomps on the lower `ValueError`](https://github.com/aws/chalice/blob/master/chalice/local.py#L460-L462). So it seems like it would be really easy to name the `ValueError` and include that message in the `ForbiddenError`.

leason:
Interesting - but that's just for local development server, right?  This problem actually affects the deployed lambdas in the same way.  For example: https://api.tekata.io/v1/warmup VS https://api.tekata.io/v1/wakeup

`warmup` is not a valid route, but `wakeup` is.

wollerman:
[...]

PR #713, opened 2018-02-13
Update message returned from local server for bad routes

wollerman:
In response to #630.

This is my first pull request, so please let me know if there's anything else I should do!

Thanks!

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -417,7 +417,7 @@ class LocalGateway(object):
         try:
             lambda_event = self._generate_lambda_event(
                 method, path, headers, body)
-        except ValueError:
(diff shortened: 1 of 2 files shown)


jamesls:
Hi, thanks for the pull request and sorry for the delay in looking at this.  I [commented on the original issue](https://github.com/aws/chalice/issues/630#issuecomment-388926383).  I'm thinking we should just make these a validation error if you define a route with a leading `/` char.  What do you think?

wollerman:
Replied there as well. Although I think updating the error message may still be helpful. Instead of stomping on the message, it may be useful to see other underlying errors. So maybe this `local.py` change is still desirable?

PR #715, opened 2018-02-15
Adding CORS headers to error_response()

artkay:
## The Problem

We are using Chalice to build backend for Single Page Apps written in JavaScript. Since the apps are working in customer's browsers, it is important for the front end developers to be able to parse erroneous response from an API. Currently if Chalice needs to respond with 415 it returns basic `Response` without CORS headers even if they were defined in the route.

## Proposed solution

To add CORS headers parsing function and optional `cors_headers` arg to `error_response()`. Now if we decide to add route headers to the erroneous response we can just pass dict with the headers.


chalice/app.py
```diff
@@ -38,9 +38,14 @@ def handle_decimals(obj):
     return obj
 
 
-def error_response(message, error_code, http_status_code):
+def error_response(message, error_code, http_status_code, cors_headers=None):
     body = {'Code': error_code, 'Message': message}
     response = Response(body=body, status_code=http_status_code)
+    if cors_headers:
```
stealthycoin: No need to add the headers one at a time given that there aren't any other headers on the response at the moment. And in the future if they are we can rely on some up stream function to do the merging and just pass this one the final set of headers.

So how about changing `cors_headers` to just `headers` and passing it directly into the `Response` object.

chalice/app.py
```diff
@@ -603,6 +608,11 @@ def __call__(self, event, context):
                                        event['requestContext'],
                                        event['stageVariables'],
                                        event.get('isBase64Encoded', False))
+        # We're getting the CORS headers before validation to be able to
+        # output desired headers with
+        cors_headers = dict()
```
stealthycoin: Can we change this to `None` instead?

If we do this the `None` default of the headers argument in the `error_response` function will never be useful since its always a `dict`, empty or not. The expected input to the `Response` object if there are no headers is `None` so I would like to keep that consistent.

chalice/app.py
```diff
@@ -615,11 +625,12 @@ def __call__(self, event, context):
                     error_code='UnsupportedMediaType',
                     message='Unsupported media type: %s' % content_type,
                     http_status_code=415,
+                    cors_headers=cors_headers
                 )
         response = self._get_view_function_response(view_function,
                                                     function_args)
-        if self._cors_enabled_for_route(route_entry):
-            self._add_cors_headers(response, route_entry.cors)
+        if cors_headers:
```
stealthycoin:
```python
if cors_headers is not None:
```

If we use `None` instead of empty `dict`.

chalice/app.py
```diff
@@ -692,8 +704,14 @@ def _validate_response(self, response):
     def _cors_enabled_for_route(self, route_entry):
         return route_entry.cors is not None
 
-    def _add_cors_headers(self, response, cors):
+    def _get_cors_headers(self, cors):
```
stealthycoin: Is the intent here just to copy the dictionary? 

You can just use `dict(cors.get_access_control_headers())` or `cors.get_access_control_headers().copy()`. No need to manually iterate.

artkay:
@stealthycoin thanks for the review. I will update the PR in a few days.

artkay:
@stealthycoin PR updated

artkay:
@stealthycoin PR updated and ready for re-review

chalice/app.py
```diff
@@ -693,8 +702,11 @@ def _validate_response(self, response):
     def _cors_enabled_for_route(self, route_entry):
         return route_entry.cors is not None
 
-    def _add_cors_headers(self, response, cors):
-        for name, value in cors.get_access_control_headers().items():
+    def _get_cors_headers(self, cors):
+        return dict(cors.get_access_control_headers())
```
stealthycoin: `cors.get_access_control_headers()` returns a `dict` already.

jamesls: I don't think we need the `dict()` call because `get_access_control_headers()` return a dict.

jamesls:
Rebased and merged via a7e12d3

PR #719, opened 2018-02-18
Custom JSONEncoder per application

Segflow:
We are using a homemade ORM to access DynamoDB, and we needed a way to simply return a Model in a view to get it's JSON representation. 

Working with the existent `JSONEncoder` (That only adds Decimal encoding) wasn't very helpful for us. As a workaround we made it possible to plug a custom `JSONEncoder` when creating the application.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -36,7 +36,7 @@ class DefaultJSONEncoder(json.JSONEncoder):
         # to support that as well.
         if isinstance(o, decimal.Decimal):
             return float(o)
-        return super().default(o)
+        return super(DefaultJSONEncoder, self).default(o)
 
 
 def error_response(message, error_code, http_status_code):
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -315,7 +315,8 @@ class Request(object):
 
 
 class Response(object):
-    def __init__(self, body, headers=None, status_code=200, json_encoder=DefaultJSONEncoder):
+    def __init__(self, body, headers=None, status_code=200,
+                 json_encoder=DefaultJSONEncoder):
         self.body = body
         if headers is None:
             headers = {}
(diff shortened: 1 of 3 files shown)


docs/source/api.rst
```diff
@@ -42,6 +42,23 @@ Chalice
          app = Chalice(app_name="appname")
          app.debug = True
 
+    .. attribute:: json_encoder
```
stealthycoin: Could you move this down to the `Response` section of the API docs. This is in the `Chalice` section which will not have a `json_encoder` property.

tests/unit/test_app.py
```diff
@@ -1316,7 +1316,7 @@ def test_http_request_to_dict_is_json_serializable(http_request_kwargs):
     request_dict = request.to_dict()
     # We should always be able to dump the request dict
     # to JSON.
-    assert json.dumps(request_dict, default=handle_decimals)
+    assert json.dumps(request_dict, cls=DefaultJSONEncoder)
```
stealthycoin: Could we get another test added that injects a custom encoder and does something with it that the default encoder cannot do?

chalice/app.py
```diff
@@ -430,7 +433,8 @@ class Chalice(object):
 
     FORMAT_STRING = '%(name)s - %(levelname)s - %(message)s'
 
-    def __init__(self, app_name, debug=False, configure_logs=True, env=None):
```
stealthycoin: I would also rather not expose this through the initializer. Could we instead do something like we do for default binary content types where we expose some property off of the `Chalice` object. For example we could do something like:

```python
app = Chalice(...)
app.default_response_serializer = SuperGreatSerializerClass
```

stealthycoin:
Another thought about this is instead of having a separate JSON serializer. How about using a `post_request` hook similar to how flask works? You could return a model object from the view and register a post request hook to convert that to a response object, or something similar to that workflow.

jamesls:
I think that the after_request hook from https://github.com/aws/chalice/pull/181 would be best given it has more general applicability.  I'll take a look at that PR.

vyscond:
So when does it going to arrive on a release?

jamesls:
I believe this can now be implemented via middleware (https://github.com/aws/chalice/pull/1514)

PR #722, opened 2018-02-20
Add support for bz2 compressed sdists from PyPi

### LINKED ISSUE #720: Deploying with Rasa doesn't work.
Hi, 

Really liking chalice - but was trying something the other day that seems to flummox my local test deployment to a folder.

As title says: trying to bundle this with Rasa (http://rasa.com).

But including Rasa seems to result in a missing output file/folder for deployment.

Isolating the issue is possible by doing the following:

```
tim$ chalice --version
chalice 1.1.0

tim$ chalice new-project testrasa

tim$ cd testrasa

tim$ echo "rasa_core" >> requirements.txt 

tim$ chalice package deploy
```

[Time passes....]

```
Creating deployment package.
/var/folders/mn/z19ts48937n8tb2v3n6hl5xc0000gn/T/tmptvnugz3f/Twisted-17.9.0.tar.bz2

tim$ ls deploy
ls: deploy: No such file or directory
```

Similar thing occurs if I use `--single_file` option...

(Might perhaps be something to do with the sheer size of this?)

Thanks. T

stealthycoin:
Support for bz2 has been removed from PyPi, and is explicitly called out
in PEP 527 as not supported. No new packages can be uploaded with the
bz2 format. However; there are still packages that have not been
migrated over, and possibly never will be. This adds support for bz2
since it is a trivial change and increases the number of packages
Chalice can automatically handle.

fixes #720 (partially)


chalice/deploy/packager.py
```diff
@@ -561,7 +561,7 @@ def _unpack_sdist_into_dir(self, sdist_path, unpack_dir):
         # type: (str, str) -> str
         if sdist_path.endswith('.zip'):
             self._osutils.extract_zipfile(sdist_path, unpack_dir)
-        elif sdist_path.endswith('.tar.gz'):
+        elif sdist_path.endswith('.tar.gz') or sdist_path.endswith('.tar.bz2'):
```
dstufft: ``sdist_path.endswith((".tar.gz", ".tar.bz2"))`` is probably cleaner here.

tests/functional/test_package.py
```diff
@@ -804,11 +804,16 @@ def _write_fake_sdist(self, setup_py, directory, ext):
             with zipfile.ZipFile(path, 'w',
                                  compression=zipfile.ZIP_DEFLATED) as z:
                 z.writestr('sdist/setup.py', setup_py)
-        else:
+        elif ext == 'tar.gz':
             with tarfile.open(path, 'w:gz') as tar:
                 tarinfo = tarfile.TarInfo('sdist/setup.py')
                 tarinfo.size = len(setup_py)
                 tar.addfile(tarinfo, io.BytesIO(setup_py.encode()))
+        elif ext == 'tar.bz2':
```
dstufft: Should there be an ``else: raise RuntimeError()`` here? or an ``assert False, "Unknown type"`` or similiar? If we ever add another ext but forget to update this, it'd be a better error message. Not a big deal either way though.

stealthycoin: Good call. I still need one that isn't valid since I need to generate invalid sdist formats so that I can test that the correct error is thrown when I try to read them. I added an `else` statement back to this to make an empty file with whatever extension was passed in.

tests/functional/test_package.py
```diff
@@ -804,11 +804,18 @@ def _write_fake_sdist(self, setup_py, directory, ext):
...
+            with tarfile.open(path, 'w:bz2') as tar:
+                tarinfo = tarfile.TarInfo('sdist/setup.py')
```
jamesls: This is almost identical to the .tar.gz condition, can you consolidate the code here?

PR #723, opened 2018-02-21
Fix the example code of error handling in docs

tocky:
The example code of error handling is written as follows:

```python
return Response(message='Plain text error message',
                headers={'Content-Type': 'text/plain'},
                status_code=400)
```

However, the implementation is:

```python
class Response(object):
    def __init__(self, body, headers=None, status_code=200):
```

There is a difference between document and code, correct name of the first parameter is `body`.


stealthycoin:
Thanks!

PR #724, opened 2018-02-21
Add --verbose to deploy command

### LINKED ISSUE #711: Add full stack trace for chalice deploy errors
Please add a --verbose flag for "chalice deploy", so that if the deploy fails, the user can debug the issue, by receiving the full stack trace of the error.

Is it stands, the user only receives the final line of the error. Though sending test events to the routes in app.py is a work-around, sometime the error isn't triggered by a route.

ridha:
Fixes #711

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -8,6 +8,7 @@ import os
 import sys
 import tempfile
 import shutil
+import traceback
 
 import botocore.exceptions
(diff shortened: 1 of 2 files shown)


stealthycoin:
Thanks for the contribution! I thought that this needed to be the default, rather than an opt in flag given how many people run into issues with this so.

Issue #697, opened 2018-01-31
Getting 'NotFoundError' only when deploying


sbassi:
In local mode works OK, but when deploying I get:

$ chalice deploy
Regen deployment package.
'NotFoundError'

Relevant lines in my code:

```
from chalice import BadRequestError, NotFoundError

@app.route('/api/v1.0/view_tx/{txid}')
def starttx(txid):
    """
    Gets the data from a transaction
    :txid: Transaction number
    """
    # Get payload
    dynamodb = boto3.resource('dynamodb')
    table = dynamodb.Table('user_txs')
    response = table.get_item(Key={'txid': txid}, ConsistentRead=True)
    if 'Item' not in response:
        raise NotFoundError('Transaction not found')
    return response['Item']
```

With other types of error, such as BadRequestError, it works and deploys OK.

joguSD:
Could you post more details about the error by running `chalice --debug deploy`?

sbassi:
I just run it with the `--debug` parameter, but the output if full of sensitive data (like plain text passwords that in my source code are encrypted) so I won't post it here. Can you tell me what to look for so I upload only that?
Here is the last output before the error:

```
2018-01-31 10:30:14,656 - botocore.hooks - DEBUG - Event needs-retry.iam.GetRole: calling handler <botocore.retryhandler.RetryHandler object at 0x7f00bd8b7ba8>
2018-01-31 10:30:14,656 - botocore.retryhandler - DEBUG - No retry needed.
2018-01-31 10:30:14,657 - botocore.hooks - DEBUG - Event after-call.iam.GetRole: calling handler <function json_decode_policies at 0x7f00be983488>
'NotFoundError'

```

sbassi:
The label closing-soon-if-no-response should be removed since I answered.

stealthycoin:
Can you either provide us with more context about the code? The section you pasted is not sufficient to find the issue. Or preferably `pip install -e .` the most recent master branch of the repo after the merging of #727 so that a full traceback off your issue shows up, and then post that?

joguSD:
Closing due to inactivity.

PR #727, opened 2018-02-27
Show tracebacks on errors

### LINKED ISSUE #711: Add full stack trace for chalice deploy errors

stealthycoin:
Will help debug #697

fixes #711 
closes #724

cc @jamesls @kyleknap

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -1,0 +1,38 @@
+import os
+import subprocess
+
+import pytest
+from chalice.utils import OSUtils
+
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+PROJECT_DIR = os.path.join(CURRENT_DIR, 'testapp')
+
+
+@pytest.fixture
+def local_app(tmpdir):
+    temp_dir_path = str(tmpdir)
+    OSUtils().copytree(PROJECT_DIR, temp_dir_path)
+    old_dir = os.getcwd()
+    try:
+        os.chdir(temp_dir_path)
+        yield temp_dir_path
+    finally:
+        os.chdir(old_dir)
+
+
+def test_stack_trace_printed_on_error(local_app):
+    app_file = os.path.join(local_app, 'app.py')
+    with open(app_file, 'w') as f:
+        f.write(
+            'from chalice import Chalice\n'
+            'app = Chalice(app_name="test")\n'
+            'foobarbaz\n'
+        )
+    p = subprocess.Popen(['chalice', 'local'],
+                         stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stderr = p.communicate()[1].decode('ascii')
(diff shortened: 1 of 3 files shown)


tests/integration/test_cli.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+import subprocess
+
+import pytest
+from chalice.utils import OSUtils
+
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+PROJECT_DIR = os.path.join(CURRENT_DIR, 'testapp')
+ROOT_DIR = os.path.dirname(os.path.dirname(CURRENT_DIR))
+MAIN_FILE = os.path.join(ROOT_DIR, 'chalice', 'cli', '__init__.py')
```
jamesls: What's the rationale for doing this instead of just invoking `chalice`?

tests/integration/test_cli.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@pytest.fixture
+def local_app(tmpdir):
+    temp_dir_path = str(tmpdir)
+    OSUtils().copytree(PROJECT_DIR, temp_dir_path)
+    os.chdir(temp_dir_path)
+    sys.path.insert(0, ROOT_DIR)
```
jamesls: Why do we need the `ROOT_DIR` on our path?  This already handled as part of your dev process setup (i.e `pip install -e .`).  Also, this is mutating a global `sys.path` without putting it back afterwards, but I don't think we need this line to begin with.

tests/integration/test_cli.py
```diff
@@ -0,0 +1,38 @@
...
+            'app = Chalice(app_name="test")\n'
+            'foobarbaz\n'
+        )
+    p = subprocess.Popen([sys.executable, MAIN_FILE, 'local'],
+                         stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    _, err = p.communicate()
+    rc = p.returncode
+
+    assert rc == 2
+    assert 'Traceback' in err.decode()
```
jamesls: We generally don't use `.decode()` calls without an encoding specified.  You can also just decode it once from the `communicate()` call above:
```
stderr = p.communicate()[1].decode('ascii')
```

stealthycoin: Fixed.

tests/integration/test_cli.py
stealthycoin: I was not making the assumption that they had done that. If I assume `chalice` is linked to this dev version that makes everything way easier, just wasn't sure that was an acceptable assumption.

PR #735, opened 2018-03-06
[new-deployer] Add support for deploying rest APIs

jamesls:
This adds a new RestAPI model as well as the needed
code to deploy/update/delete an API Gateway RestAPI.

I had to add new opcodes to get this working:

* BuiltinFunction - needed to parse ARNs
* LoadValue - set a value to a var name

Additionally, I expanded the variable substitution to crawl nested structures.
This was needed because the Swagger generation needs a reference to the lambda
arn in a deeply nested structure in the swagger doc.  I also added a new type
of variable, "StringFormat", which accepts a list of variables and renders a
string based on a format string.

This is really just a shorthand.  It's possible to implement this in
terms of a `BuiltinFunction` with the result being stored as a var.

As an example, given this app:

```python
from chalice import Chalice

app = Chalice(app_name='james19')


@app.route('/')
def index():
    return {'hello': 'world'}

```

You'll get this new deployed.json file:

```json
{
  "stages": {
    "dev": {
      "resources": [
        {
          "role_arn": "arn:aws:iam::1234:role/james19-dev",
          "name": "default-role",
          "resource_type": "iam_role"
        },
        {
          "lambda_arn": "arn:aws:lambda:us-west-2:1234:function:james19-dev-james19",
          "name": "james19",
          "resource_type": "lambda_function"
        },
        {
          "rest_api_id": "my_rest_api_id",
          "name": "rest_api",
          "resource_type": "rest_api"
        }
      ]
    }
  },
  "schema_version": "2.0"
}
```

NOTE: If you're trying this out locally, the UI hasn't been plumbed back in, so it won't print out the full URL to test.  This will be a future PR.

### DIFF
--- a/chalice/deploy/newdeployer.py
+++ b/chalice/deploy/newdeployer.py
@@ -94,7 +94,9 @@ from chalice import app  # noqa
 from chalice.deploy.packager import LambdaDeploymentPackager
 from chalice.deploy.packager import PipRunner, SubprocessPip
 from chalice.deploy.packager import DependencyBuilder as PipDependencyBuilder
+from chalice.deploy.swagger import SwaggerGenerator
 from chalice.deploy.planner import PlanStage, Variable, RemoteState
+from chalice.deploy.planner import StringFormat
 from chalice.deploy.planner import UnreferencedResourcePlanner, NoopPlanner
 from chalice.policy import AppPolicyGenerator
 from chalice.constants import LAMBDA_TRUST_POLICY
@@ -103,8 +105,8 @@ from chalice.constants import DEFAULT_LAMBDA_MEMORY_SIZE
 from chalice.awsclient import TypedAWSClient
 
 
-def create_default_deployer(session):
-    # type: (Session) -> Deployer
+def create_default_deployer(session, config):
+    # type: (Session, Config) -> Deployer
     client = TypedAWSClient(session)
     osutils = OSUtils()
     pip_runner = PipRunner(pip=SubprocessPip(osutils=osutils),
@@ -131,10 +133,14 @@ def create_default_deployer(session):
                         osutils=osutils
                     ),
                 ),
+                SwaggerBuilder(
+                    swagger_generator=TemplatedSwaggerGenerator(),
+                )
             ],
         ),
         plan_stage=PlanStage(
-            osutils=osutils, remote_state=RemoteState(client),
+            osutils=osutils, remote_state=RemoteState(
+                client, config.deployed_resources(config.chalice_stage)),
         ),
         sweeper=UnreferencedResourcePlanner(),
         executor=Executor(client),
@@ -161,11 +167,14 @@ class UnresolvedValueError(Exception):
 
     def __init__(self, key, value, method_name):
         # type: (str, models.Placeholder, str) -> None
-        msg = self.MSG % (key, value, method_name)
-        super(UnresolvedValueError, self).__init__(msg)
+        super(UnresolvedValueError, self).__init__()
         self.key = key
         self.value = value
         self.method_name = method_name
+
+    def __str__(self):
+        # type: () -> str
+        return self.MSG % (self.key, self.value, self.method_name)
 
 
 class Deployer(object):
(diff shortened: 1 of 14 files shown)


jamesls:
I pushed some new commits.  I was originally going to send this as a follow up PR to keep the scope down, but it helps explain some of the changes in this PR.

Basically, I removed the whole `DEPLOY_PHASE` idea.  Instead, we load the deployed values from deployed.json in the case of an update, or we take the values directly from the return values of an API call.  These values are then loaded and stored as variables.  This makes all the cases consistent (create vs. update) as well as cases where the only way to knew the deployed value is from the deployed.json.  For example, you don't know the appropriate rest_api_id because Rest APIs can have identical names, so there's nothing in the API you can query to know for sure.  You *have* to consult deployed.json.

chalice/deploy/newdeployer.py
```diff
@@ -517,14 +565,82 @@ def _do_jpsearch(self, instruction):
         result = jmespath.search(instruction.expression, v)
         self.stack.append(result)
 
+    def _do_builtinfunction(self, instruction):
+        # type: (models.BuiltinFunction) -> None
+        # Split this out to a separate class of built in functions
+        # once we add more functions.
+        if instruction.function_name == 'parse_arn':
```
stealthycoin: Are we hoping to keep this to one builtin? Is there a reason that we chose this way rather than just making a parse arn op code other than room for growth?

jamesls: Primarily it just seemed too specific.  I'd like the opcodes somewhat generic, and `parse_arn` just seemed too specialized.

chalice/deploy/newdeployer.py
```diff
@@ -223,8 +232,31 @@ def build(self, config, stage_name):
...
+                               config,        # type: Config
+                               deployment,    # type: models.DeploymentPackage
+                               chalice_app,   # type: app.Chalice
+                               stage_name,    # type: str
+                               ):
+        # type: (...) -> models.RestAPI
+        resource_name = config.app_name
+        lambda_function = self._create_lambda_model(
+            config, deployment, resource_name,
+            'app.app', stage_name
```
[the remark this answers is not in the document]
jamesls: Updated.

chalice/deploy/planner.py
```diff
@@ -279,14 +290,100 @@ def plan_scheduledevent(self, resource):
...
+            models.JPSearch('account_id'),
+            models.StoreValue('account_id'),
+            models.LoadValue('parsed_lambda_arn'),
+            models.JPSearch('region'),
+            models.StoreValue('region_name'),
+            # The swagger doc uses the 'api_handler_lambda_arn'
+            # var name so we need to make sure we populate this variable
+            # before importing the rest API.
+            models.LoadValue(varname),
+            models.StoreValue('api_handler_lambda_arn'),
```
kyleknap: This `shared_plan` is a little hard for me to follow. Is this just taking all of the variables from the parse_arn and storing it as variables to use later? We talked about this offline but I realize that this is suppose to be really low level, but I think in the future it would be nice if we moved over to a register-based implementation. Keeping track of where the stack is at is a little difficult as you have to know how each opcode affects the stack.

chalice/deploy/planner.py
```diff
@@ -279,14 +290,100 @@ def plan_scheduledevent(self, resource):
...
+                    params={'function_name': function_name,
+                            'region_name': Variable('region_name'),
+                            'account_id': Variable('account_id'),
+                            'rest_api_id': Variable('rest_api_id')},
+                ),
+            ]
+        else:
+            deployed = self._remote_state.resource_deployed_values(resource)
+            plan = shared_plan + [
+                models.Push(deployed['rest_api_id']),
```
joguSD: Capturing some offline feedback, I'd be curious to see what these would look like if there was only a stack or only variables.

chalice/deploy/newdeployer.py
```diff
@@ -479,29 +524,47 @@ def _do_storevalue(self, instruction):
...
+
+    def _add_to_deployed_values(self, payload):
+        # type: (Dict[str, str]) -> None
+        key = payload['name']
+        if key not in self._resource_value_index:
+            self._resource_value_index[key] = payload
+            self.resource_values.append(payload)
+        else:
+            # If the key already exists, we merge the new payload
+            # with the existing payload.
```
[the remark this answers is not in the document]
jamesls: If you want to store multiple deployed values for a resource.  If you take a look at this commit I added (https://github.com/aws/chalice/pull/735/commits/5972786a0dfc931c679060cb59ea3b6b5ea871cd) you can see we want to store the role_arn as well as the role_name to avoid having to parse our the name from the arn when we delete a role.

chalice/deploy/planner.py
jamesls: >  Is this just taking all of the variables from the parse_arn and storing it as variables to use later?

Yes.   This is needed for both the create as well as the update case, so it's a shared set of instructions.

chalice/deploy/newdeployer.py
```diff
@@ -479,29 +524,47 @@ def _do_storevalue(self, instruction):
...
+        self._add_to_deployed_values(payload)
+
+    def _add_to_deployed_values(self, payload):
+        # type: (Dict[str, str]) -> None
+        key = payload['name']
+        if key not in self._resource_value_index:
+            self._resource_value_index[key] = payload
+            self.resource_values.append(payload)
+        else:
+            # If the key already exists, we merge the new payload
```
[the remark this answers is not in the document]
jamesls: See comment above https://github.com/aws/chalice/pull/735#discussion_r173009990

chalice/deploy/newdeployer.py
```diff
@@ -517,14 +580,82 @@ def _do_jpsearch(self, instruction):
         result = jmespath.search(instruction.expression, v)
         self.stack.append(result)
 
+    def _do_builtinfunction(self, instruction):
+        # type: (models.BuiltinFunction) -> None
+        # Split this out to a separate class of built in functions
```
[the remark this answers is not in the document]
jamesls: I guess.  It's not really an active TODO though because there's only a single function for now.  The other ones are actionable now.

jamesls:
@kyleknap @joguSD The latest commit (3a5ac4e87de02273ee173839485ddc3cc230c6a1) removes the stack in favor of explicit variables for input/output.  Let me know what you think.

PR #744, opened 2018-03-12
[new-deployer] Add back print statements during deploy

jamesls:
The basic idea is to update the planner to be a python object instead of a list
of instructions.  This allows us to add a `messages` attribute that contains of
mapping of instructions to messages to print.

The original plan was to just the instruction object directly as a key.
However, because it contains mutable attributes, the `id()` of the instruction
is used.  Not opposed to other alternative implementations.

The biggest thing that's missing right now is the formation of the Rest API URL at
the end of the deployment.  The old deployer did this, but I haven't plumbed
this back in.  I was to think about whether I should just be printing out a
more detailed overview of all the resources we created for you instead of just
the Rest API, especially as we add support for more resources.  This will be a
separate pull request.

### Alternatives considered:

* Separate PRINT instructions.  The reason against this is that is clutters the
  plan with unnecessary metadata.
* Have the Executor print every `APICall` instruction.  Too low level and not
  enough control over the contents of the message.

### DIFF
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -66,15 +66,16 @@ class RemoteState(object):
 class UnreferencedResourcePlanner(object):
 
     def execute(self, plan, config):
-        # type: (List[models.Instruction], Config) -> None
-        marked = set(self._mark_resources(plan))
+        # type: (models.Plan, Config) -> None
+        instructions = plan.instructions
+        marked = set(self._mark_resources(instructions))
         deployed = config.deployed_resources(config.chalice_stage)
         if deployed is not None:
             deployed_resource_names = reversed(deployed.resource_names())
             remaining = [
                 name for name in deployed_resource_names if name not in marked
             ]
-            self._plan_deletion(plan, remaining, deployed)
+            self._plan_deletion(instructions, remaining, deployed)
 
     def _mark_resources(self, plan):
         # type: (List[models.Instruction]) -> List[str]
@@ -125,7 +126,7 @@ class PlanStage(object):
         self._osutils = osutils
 
     def execute(self, resources):
-        # type: (List[models.Model]) -> List[models.Instruction]
+        # type: (List[models.Model]) -> models.Plan
         plan = []  # type: List[models.Instruction]
         for resource in resources:
             name = 'plan_%s' % resource.__class__.__name__.lower()
@@ -134,7 +135,7 @@ class PlanStage(object):
                 result = handler(resource)
                 if result:
                     plan.extend(result)
-        return plan
+        return models.Plan(plan, {})
 
     # TODO: This code will likely be refactored and pulled into
     # per-resource classes so the PlanStage object doesn't need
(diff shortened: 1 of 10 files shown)


chalice/deploy/planner.py
```diff
@@ -96,12 +103,16 @@ def _plan_deletion(self,
...
+                messages[id(apicall)] = (
+                    "Deleting function: %s\n" % resource_values['lambda_arn'])
                 plan.append(apicall)
             elif resource_values['resource_type'] == 'iam_role':
                 apicall = models.APICall(
                     method_name='delete_role',
                     params={'name': resource_values['role_name']},
                 )
+                messages[id(apicall)] = (
+                    "Deleting IAM role: %s\n" % resource_values['role_name'])
```
kyleknap: In the old deployer, I think we prompted on whether we wanted to delete the role. Do we want to keep doing that?

jamesls:
@kyleknap Pushed a test fix, it was because py3 `dict_values([]) != list(...)`

chalice/deploy/planner.py
```diff
@@ -125,23 +138,30 @@ def __init__(self, remote_state, osutils):
...
         for resource in resources:
-            name = 'plan_%s' % resource.__class__.__name__.lower()
+            name = '_plan_%s' % resource.__class__.__name__.lower()
             handler = getattr(self, name, None)
             if handler is not None:
                 result = handler(resource)
                 if result:
-                    plan.extend(result)
-        return plan
+                    for single in result:
```
kyleknap: This is getting pretty nested. I wonder if it makes sense to add a helper method to clean this up a bit? Maybe something like`_add_result(self, result, plan, messages)`?

chalice/deploy/planner.py
jamesls: I'm not sure.  I was wondering if it made sense to have a more general pattern to this.  For example:

* Add a `plan` command that pretty prints the plan of the exact API calls we want to make
* Update the `deploy` command to pretty print the plan and confirm if we should proceed or not.

The thing I don't like about the previous approach was that we cherry picked API calls we thought were worth flagging and I'd like something more general.  It also meant that you couldn't know before hand whether or not a deploy was going to prompt you for something (making it harder to script).

Thoughts?

kyleknap: I like the general idea. It definitely was weird that we would cherry-pick the api calls. As to the examples, I think I like the `plan` command the most, but instead what if we added a `--dryrun` or `--dry-run` flag that will pretty print the plan out? The reason I suggest this is that we could add the flag to both `deploy` and `delete`. If we just had a `plan` command, we would have to add an argument to differentiate what chalice command we are planning for. As for pretty printing the `deploy` command and confirming, would that be opt-in? I really like the ability to just run `chalice deploy` and it all works without further interactions.

jamesls: Here's a sketch of what we could do:

* By default, deploy/delete no longer prompt.  This makes it easier for the people scripting so they have a guarantee that we will never prompt by default.
* Add an option to `deploy/delete` (something like `--confirm`).  This would pretty print the plan and require confirmation from the user before executing the plan.  This is opt-in.
* Add a `--dry-run` flag to `deploy` and `delete`.  I think this would be a post-merge to `develop` feature.  It would essentially pretty print the plan just like `confirm`, and then exit.  We might even be able to add some sort of `FakeExecutor` that knows how to stub out API calls so you can actually test out the execution plan (i.e this would catch bugs where the planner emits a bad plan).

chalice/deploy/planner.py
jamesls: Updated.

PR #745, opened 2018-03-12
update README to specify newer version of boto3 requirement

gweedo767:
Calling put_object() fails on boto3 version 1.3.3 due to a change in authorization mechanism requirements.  The error you will get is "Please use AWS4-HMAC-SHA256"


stealthycoin:
Seems to work for me with that version.

stealthycoin:
Closing due to inactivity.

PR #750, opened 2018-03-19
Add implicit HEAD method when creating GET route only

dabio:


jamesls:
Hi, thanks for the pull request.  I'm hesitant to change the default behavior to implicitly add a HEAD request.  Given the current behavior, I think it might be a surprising change for existing users.  Could you talk more about why you'd like this feature?

jamesls:
Closing due to in activity.  If anyone would like this feature, feel free to comment on this PR.

PR #751, opened 2018-03-20
Add cleanups for post deployment processing

jamesls:
This PR cleans up the steps that happen once the deployer is finished.  This
has been broken down into separate commits, but here's a high level summary:

* Move the record of deployed values into the deployer itself.  Previously this
lived in the `deploy()/delete()` command in `chalice/cli/__init__.py` which was always
odd.  This removes duplication that lived in the `deploy/delete` command as
well as the integration tests.
* Add the `backend: api` key to the deployed JSON files.  This was in the
previous deployer and lets us know how these deployed values were generated.
This was done primarily for parity with the old deployer.
* Add the rest_api_url to the list of deployed values.  Previously the old
deployer knew how to construct the URL but it seems more appropriate for the
planner to construct this value.  In order to do this I needed to add a new
builtin function to handle string formatting.
* Add a `DeploymentReporter` to pretty print the relevant resources that were
deployed.  For now this is just lambda functions and the rest api.

Sample output:

```
$ chalice deploy-new
Creating deployment package.
Updating policy for IAM role: james2-dev
Updating lambda function: james2-dev-foo
Updating lambda function: james2-dev
Updating rest API
Resources deployed:
  - Lambda ARN: arn:aws:lambda:us-west-2:123:function:james2-dev-foo
  - Lambda ARN: arn:aws:lambda:us-west-2:123:function:james2-dev
  - Rest API URL: https://abcd.execute-api.us-west-2.amazonaws.com/api/
```

### DIFF
--- a/chalice/deploy/newdeployer.py
+++ b/chalice/deploy/newdeployer.py
@@ -208,11 +208,7 @@ class Deployer(object):
         self._sweeper.execute(plan, config)
         self._executor.execute(plan)
         deployed_values = {
-            'stages': {
-                chalice_stage_name: {
-                    'resources': self._executor.resource_values,
-                },
-            },
+            'resources': self._executor.resource_values,
             'schema_version': '2.0',
         }
         self._recorder.record_results(
--- a/chalice/deploy/newdeployer.py
+++ b/chalice/deploy/newdeployer.py
@@ -597,7 +597,7 @@ class Executor(object):
 
     def _do_builtinfunction(self, instruction):
         # type: (models.BuiltinFunction) -> None
-        # Split this out to a separate class of built in functions
+        # TODO: Split this out to a separate class of built in functions
         # once we add more functions.
         if instruction.function_name == 'parse_arn':
             arg = instruction.args[0]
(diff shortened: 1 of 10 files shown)


chalice/deploy/newdeployer.py
```diff
@@ -203,14 +210,17 @@ def deploy(self, config, chalice_stage_name):
...
-        return {
-            'stages': {
-                chalice_stage_name: {
-                    'resources': self._executor.resource_values,
-                },
-            },
+        deployed_values = {
+            'resources': self._executor.resource_values,
             'schema_version': '2.0',
+            'backend': self.BACKEND_NAME,
```
kyleknap: So I read the overview, but I am still not following why we need the backend parameter. Specifically, how did we use this in the old deployer? Or is this so we can track future deployment methods over time?

jamesls: So we can validate future deployment methods.  I'm thinking specifically for the cfn case.  You can't mix the API based deployer with the cfn deployers so I'd like some way to validate this and warn the user that this isn't supported.

I'd still like the cfn deployer to generate this same file so that things that depend on it (say `chalice url`) don't need to know about specific deployer implementations.

kyleknap: Sounds good. Makes sense.

PR #754, opened 2018-03-20
Create empty deployed resources when no deployment exists

jamesls:
Previously, we returned Optional[DeployedResources], which
made usage awkward because you always had to check if
deployed_resources was not None.  With this change, if there
has been no deployment, we just create an empty DeployedResources.
Consuming code now needs to catch a `ValueError` if they try
to access a resource that was not deployed.

This fixed a bug where `chalice deploy` immediately after a
`chalice delete` would fail.

### DIFF
--- a/chalice/config.py
+++ b/chalice/config.py
@@ -275,7 +275,7 @@ class Config(object):
         return clone
 
     def deployed_resources(self, chalice_stage_name):
-        # type: (str) -> Optional[DeployedResources]
+        # type: (str) -> DeployedResources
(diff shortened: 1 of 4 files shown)


jamesls:
I also pushed a fix for the `chalice deploy && chalice deploy --stage prod`.  It ended up touching the same code and was a small enough fix that I didn't want to bother you all with multiple PRs for this.  https://github.com/aws/chalice/pull/754/commits/c139cd480ebec564b5d7db7cd042251f47600698

PR #756, opened 2018-03-20
Handle upgrading 0.10.0 to the new deployer format

jamesls:


tests/unit/test_config.py
```diff
@@ -540,3 +540,33 @@ def test_can_upgrade_rest_api(self):
...
+            "api_handler_name": "app-dev",
+            "api_handler_arn": (
+                "arn:aws:lambda:us-west-2:123:function:app-dev"),
+            "rest_api_id": "my_rest_api_id",
+            "lambda_functions": {
+                # This is the old < 1.0 style where the
+                # value was just the lambda arn.
+                "app-dev-foo": "my-lambda-arn",
+            },
+            "chalice_version": "1.1.1",
```
kyleknap: Nit pick but the version should be `0.10.0`

jamesls: I'll get that updated before merging.

PR #766, opened 2018-03-25
Chalice was droping query string params in local mode

### LINKED ISSUE #767: Unable to retrieve the request path from event in local mode
This is a two part issue. 

1) The Chalice request object does not expose the original path. I assume this is because the route decorator has already parsed the path in to variables and therefore, you no longer the original. However, there are occasions you might need it. In my case I need to validate a oauth1 signature using the pyLTI module. So this is not really an issue but a feature request. 

2) There is a workaround to the limitation above. API Gateway includes the path in the requestContext which is available from the Chalice request object. Unfortunately, Chalice local mode omits the path (and other keys) from the requestContext. This is a but as local mode should behave the same way as API Gateway whenever possible.  As a result my test cases are failing. Will submit a pull request to fix this.

brianjbeach:
API Gateway includes the path in the requestContext which is available from the Chalice request object. Unfortunately, Chalice local mode omits the path (and other keys) from the requestContext. This is a bug as local mode should behave the same way as API Gateway whenever possible. As a result my test cases are failing. Fixes #767


stealthycoin:
Can you get the tests passing before I take a look at this?

stealthycoin:
Pulled this down and tested it and compared to real API Gateway. Looks good to me.

jamesls:
Rebased and merged via 19be95a.  Thanks for the pull request.

PR #772, opened 2018-03-28
Update package command to work with new deployer

jamesls:
This brings numerous benefits:

1. Full parity with `chalice deploy`.  This means you can use the
package command for pure lambda functions, built in authorizers,
and scheduled events.
2. We'll now get exceptions raised when a new resource isn't supported
in the `package` command instead of silently ignoring the resource
like we do now.

The generated template is slightly different from the old deployer,
notably:

1. The `Events` key is no longer used for `Api` types.  Instead we just
manually create a lambda permission resource that allows api gateway to
invoke the API handler.  This also more closely matches what
`chalice deploy` does.
2. The `Policies` is just the generated policy (when it applies).
Previously the value was a list of policies, but chalice only generates
a single policy.

Here's an worked example with diffs of the old vs. new packager: https://gist.github.com/jamesls/c80880fb7df39bc33ef4d30aca7d6995

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -276,10 +276,13 @@ class TemplatePostProcessor(object):
         # outdir.  That would require plumbing through user
         # provided params such as "outdir" into the build stage
         # somehow, which isn't currently possible.
+        copied = False
         for resource in template['Resources'].values():
             if resource['Type'] != 'AWS::Serverless::Function':
                 continue
             original_location = resource['Properties']['CodeUri']
             new_location = os.path.join(outdir, 'deployment.zip')
-            self._osutils.copy(original_location, new_location)
+            if not copied:
+                self._osutils.copy(original_location, new_location)
+                copied = True
             resource['Properties']['CodeUri'] = './deployment.zip'
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -103,6 +103,10 @@ class SAMTemplateGenerator(object):
                 'MemorySize': resource.memory_size,
             },
         }
+        if resource.environment_variables:
+            resources[cfn_name]['Properties']['EnvironmentVariables'] = {
+                'Variables': resource.environment_variables
+            }
         self._inject_iam_permissions(resource, resources[cfn_name])
 
     def _inject_iam_permissions(self, resource, cfn_resource):
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -1,39 +1,34 @@
 import os
 import copy
 
-from typing import Any, Dict  # noqa
+from typing import Any, Dict, List  # noqa
 
 from chalice.deploy.swagger import CFNSwaggerGenerator
-from chalice.deploy.swagger import SwaggerGenerator  # noqa
-from chalice.deploy.packager import LambdaDeploymentPackager
-from chalice.deploy.packager import DependencyBuilder
-from chalice.deploy.deployer import ApplicationPolicyHandler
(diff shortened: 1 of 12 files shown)


stealthycoin:
Nice I played around with this quite a bit and it seems to work well.

chalice/package.py
```diff
@@ -1,217 +1,285 @@
...
+        # copies the DeploymentPackage.filename over to the
+        # outdir.  That would require plumbing through user
+        # provided params such as "outdir" into the build stage
+        # somehow, which isn't currently possible.
+        for resource in template['Resources'].values():
+            if resource['Type'] != 'AWS::Serverless::Function':
+                continue
+            original_location = resource['Properties']['CodeUri']
+            new_location = os.path.join(outdir, 'deployment.zip')
+            self._osutils.copy(original_location, new_location)
```
kyleknap: Are we copying over the `deployment.zip` for every lambda/SAM resource? Is it possible to make sure we do this once only if we are?

tests/unit/test_package.py
```diff
@@ -13,362 +15,192 @@ def mock_swagger_generator():
...
-        'swagger': 'document'
-    }
-    config = Config.create(chalice_app=sample_app,
-                           api_gateway_stage='dev')
-    template = p.generate_sam_template(config)
-    properties = template['Resources']['RestAPI']['Properties']
-    assert properties['DefinitionBody'] == {'swagger': 'document'}
-
-
-def test_can_inject_environment_vars(sample_app,
```
kyleknap: Looks like we missed this one when porting over the tests. I also cannot find it in the ported over implementation.

chalice/package.py
```diff
@@ -1,217 +1,285 @@
...
+        # type: (models.AutoGenIAMPolicy, Dict[str, Any]) -> None
+        pass
+
+    def _generate_deploymentpackage(self, resource, template):
+        # type: (models.DeploymentPackage, Dict[str, Any]) -> None
+        pass
+
+    def _default(self, resource, template):
+        # type: (models.Model, Dict[str, Any]) -> None
+        raise NotImplementedError(resource)
```
kyleknap: I wonder if it makes sense to add a test for this? It looks like it would work but may be nice to double check.

chalice/package.py
```diff
@@ -1,217 +1,285 @@
...
+        resources = template['Resources']
+        cfn_name = to_cfn_resource_name(resource.resource_name)
+        resources[cfn_name] = {
+            'Type': 'AWS::Serverless::Function',
+            'Properties': {
+                'Runtime': resource.runtime,
+                'Handler': resource.handler,
+                'CodeUri': resource.deployment_package.filename,
+                'Tags': resource.tags,
+                'Timeout': resource.timeout,
```
kyleknap: Some of these values used to have `if is not None` check on it. It may be good to check that these values are actually never `None` to make sure it does not affect the generated cloudformation template.

jamesls: This happens at a separate layer above.  We previously had to check at this layer because we were using the `config` object directly.  I'll double check, but I'm pretty sure it's now impossible for these values to be None now that we're using the models.

tests/unit/test_package.py
jamesls: Good catch.  Updated.

chalice/package.py
jamesls: Yeah we are.  I've updated this.

jamesls:
@kyleknap I believe I've fixed everything.  Let me know if I missed anything.

PR #774, opened 2018-04-02
Update SAM template to have parity with chalice deploy

jamesls:
This fixes issues with the IAM role generation logic used when
creating Lambda function.  Previously, we'd just attach the policy
directly to the serverless function resource.  This resulted in
generating a role per lambda function.  However, this didn't
match the behavior of `chalice deploy`.  The default behavior is
that all the Lambda functions share the same role unless you
explicitly configure things.

In order to achieve this parity, a separate IAM role resource is
now created, and the serverless function resource is updated
to point to this resource.  This allows us to match exactly
what `chalice deploy` does.

There are two user visible changes/consequences from this:

1. I udpated the resource names to be `CamelCased`.  This fixes
a consistency issue where the hard coded names were camel cased
(e.g `APIHandler`) whereas the autogenerated names removed
non-alphanumeric character and added an md5 hash on the end.

2. Because we are explicitly created a named IAM resource, you
need to use the `CAPABILITY_NAMED_IAM` instead of `CAPABILITY_IAM`
when deploying your SAM template.

(2) is a bit iffy, but this is fixing a bug, so we need to decide
if fixing the bugs with the inconsistency between package/deploy
is more important than preserving `CAPABILITY_IAM` in the default
configuration case.

### DIFF
--- a/chalice/utils.py
+++ b/chalice/utils.py
@@ -3,7 +3,6 @@ import os
 import zipfile
 import json
 import contextlib
-import hashlib
 import tempfile
 import re
 import shutil
--- a/chalice/utils.py
+++ b/chalice/utils.py
@@ -29,10 +29,10 @@ def to_cfn_resource_name(name):
     This transform ensures that only alphanumeric characters are used
     and prevent collisions by appending the hash of the original name.
     """
-    alphanumeric_only_name = re.sub(r'[^A-Za-z0-9]+', '', name)
-    return ''.join([
-        alphanumeric_only_name, hashlib.md5(
-            name.encode('utf-8')).hexdigest()[:4]])
+    for word_separator in ['-', '_']:
+        word_parts = name.split(word_separator)
+        name = ''.join([w[0].upper() + w[1:] for w in word_parts])
+    return re.sub(r'[^A-Za-z0-9]+', '', name)
 
 
 def remove_stage_from_deployed_values(key, filename):
--- a/chalice/utils.py
+++ b/chalice/utils.py
@@ -25,11 +25,17 @@ def to_cfn_resource_name(name):
     # type: (str) -> str
     """Transform a name to a valid cfn name.
 
-    This transform ensures that only alphanumeric characters are used
-    and prevent collisions by appending the hash of the original name.
+    This will convert the provided name to a CamelCase name.
+    It's possible that the conversion to a CFN resource name
+    can result in name collisions.  It's up to the caller
+    to handle name collisions appropriately.
+
     """
-    for word_separator in ['-', '_']:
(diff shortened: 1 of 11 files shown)


jamesls:
@kyleknap merged the docfix pr and rebased against master.  Should be good to look at again.

chalice/utils.py
```diff
@@ -29,10 +28,10 @@ def to_cfn_resource_name(name):
     This transform ensures that only alphanumeric characters are used
     and prevent collisions by appending the hash of the original name.
     """
-    alphanumeric_only_name = re.sub(r'[^A-Za-z0-9]+', '', name)
-    return ''.join([
-        alphanumeric_only_name, hashlib.md5(
-            name.encode('utf-8')).hexdigest()[:4]])
+    for word_separator in ['-', '_']:
+        word_parts = name.split(word_separator)
+        name = ''.join([w[0].upper() + w[1:] for w in word_parts])
```
kyleknap: This breaks if you have a resource that starts with underscore (i.e. `_index()`). Do you think we need to account for this? I can see this being helpful.

docs/source/upgrading.rst
```diff
@@ -117,6 +118,31 @@ the format as you deploy a given stage.
...
+non alphanumeric characters and appending an md5 checksum,
+e.g ``my_function -> myfunction3bfc``.  With this new packager
+update, the resource name would be converted as
+``my_function -> MyFunction``.  Note, the ``Outputs`` section
+renames unchanged in order to preserve backwards compatibility.
+In order to fix parity issues with ``chalice deploy`` and
+``chalice package``, we now explicitly create an IAM role
+resource as part of the default configuration.  This means
+the default SAM template requires capability ``CAPABILITY_NAMED_IAM``
+instead of ``CAPABILITY_IAM``.  As always, you still have the
```
kyleknap: It would be good to be more explicit in what people may have to do to potentially fix this if they have the wrong capabilities set.

jamesls:
Quick update:  I'm incorporating review feedback and updating the docs.  I'll update this PR when it's ready to be looked at again.

docs/source/upgrading.rst
jamesls: You know what, I don't actually need to use `CAPABILITY_NAMED_IAM`.  It doesn't really matter if we don't use the same role name used by `chalice deploy`.  I think the important point is that the role to lambda function mapping (and their associated policies) are what matter.  Letting cloudformation autogenerate the role is actually fine.  I'm going to update the PR to account for this.

chalice/utils.py
jamesls: Yeah, I'll also add property tests for this as well.

PR #782, opened 2018-04-04
Fix sqlalchemy

stealthycoin:
Sqlalchemy incorrectly reports itself as needing C extensions. This
whitelists it so that it can be deployed.


stealthycoin:
Merged 1c4a4f11ed879c8687e11b081531fea24f2aef1e

PR #784, opened 2018-04-04
Remove Pure Lambda docs limitation

BigChief45:
Pure Lambda functions no longer need at least one @app.route
decorator.


jamesls:
Good catch!

Issue #748, opened 2018-03-18
Environment variables on chalice local


epicfaace:
Environment variables specified in my config.json are not being used when running ```chalice local```

stealthycoin:
Marking as an enhancement for local mode.

johntmyers:
It works already, "kind of". The environment variables in `config.json` will only be available inside of decorated route handlers, not in the main body of `app.py`. I haven't figured it out yet.  

Here's my test config:

```json
{
  "version": "2.0",
  "app_name": "env-test",
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "environment_variables": {
        "__FOO__": "BAR"
      }
    }
  }
}
```

Here's a modified version of `cli/__init__.py` that prints the env var:

```python
def run_local_server(factory, host, port, stage, env):
    # type: (CLIFactory, str, int, str, MutableMapping) -> None
    config = factory.create_config_obj(
        chalice_stage_name=stage
    )
    # We only load the chalice app after loading the config
    # so we can set any env vars needed before importing the
    # app.
    env.update(config.environment_variables)
    print('run_local_server: ' + os.getenv('__FOO__'))
    app_obj = factory.load_chalice_app()
    # Check that `chalice deploy` would let us deploy these routes, otherwise
    # there is no point in testing locally.
    routes = config.chalice_app.routes
    validate_routes(routes)
    # When running `chalice local`, a stdout logger is configured
    # so you'll see the same stdout logging as you would when
    # running in lambda.  This is configuring the root logger.
    # The app-specific logger (app.log) will still continue
    # to work.
    logging.basicConfig(stream=sys.stdout)
    server = factory.create_local_server(app_obj, config, host, port)
    server.serve_forever()
```

Here's my app.py:

```python
import os

from chalice import Chalice

app = Chalice(app_name='env-test')

print('in app.py: ' + os.getenv('__FOO__', 'None'))

@app.route('/')
def index():
    print(os.getenv('__FOO__'))
    return {'hello': 'world'}
```

Here's my output when I run `chalice local` and also hit the default route:

```
$ chalice local --stage dev
in app.py: None
run_local_server: BAR
Serving on 127.0.0.1:8000
BAR
127.0.0.1 - - [12/Apr/2018 20:24:41] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [12/Apr/2018 20:24:41] "GET /favicon.ico HTTP/1.1" 403 -
```

johntmyers:
[...]

PR #798, opened 2018-04-13
Actually load environment variables before the first loading of the app

johntmyers:
This may address #748 where environment variables are not available outside of route handler functions. The app actually gets loaded on both the config object and as a stand alone object within `run_local_server` when using `chalice local`. It's fairly confusing why this happens, but this change actually sets the OS environment variables to what you have in `config.json` before the _first_ time the app gets loaded (in `create_config_obj`)

### DIFF
--- a/chalice/cli/factory.py
+++ b/chalice/cli/factory.py
@@ -119,8 +119,6 @@ class CLIFactory(object):
                                % err)
 
         self._validate_config_from_disk(config_from_disk)
-        app_obj = self.load_chalice_app()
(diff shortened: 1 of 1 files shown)


BigChief45:
@stealthycoin @jamesls Could we get a review on this? I think this is a very important change that many people need.

jamesls:
Change looks good, I'd just like to get a test added for this.  I feel like we've had this before but we've regressed on this feature.  I'd like something in place to make sure we don't lose this feature again.

PR #809, opened 2018-04-23
Upgrade dependencies

jamesls:
Initially this started with an upgrade to mypy because I wanted to pull in TypedDict, then I ended up upgrading a few of the other test dependencies.  The links to the changelogs are in the individual commit messages.

I also had to fix up some code to account for errors that mypy was previously not flagging.

### DIFF
--- a/chalice/analyzer.py
+++ b/chalice/analyzer.py
@@ -36,7 +36,7 @@ particular ``FunctionDef`` node.
 import ast
 import symtable
 
-from typing import Dict, Set, Any, Optional, List, Union  # noqa
+from typing import Dict, Set, Any, Optional, List, Union, cast  # noqa
 
 
 APICallT = Dict[str, Set[str]]
@@ -126,7 +126,7 @@ class Boto3ClientType(BaseType):
 
     def __eq__(self, other):
         # type: (Any) -> bool
-        if not isinstance(other, self.__class__):
+        if not isinstance(other, Boto3ClientType):
             return False
         return self.service_name == other.service_name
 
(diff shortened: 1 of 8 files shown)


chalice/analyzer.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, service_name):
 
     def __eq__(self, other):
         # type: (Any) -> bool
-        if not isinstance(other, self.__class__):
+        if not isinstance(other, Boto3ClientType):
```
kyleknap: What was the reason for changing these from using `__class__`? It is unfortunate now that if we change the name of this class that we have to remember to change the `isinstance` check. I think this change happens a couple more times as well.

jamesls: Mypy bug: https://github.com/python/mypy/issues/3061

```
$ cat /tmp/a.py
from typing import Any


class A:
    def __init__(self, x):
        # type: (int) -> None
        self.x = x

    def __eq__(self, other):
        # type: (Any) -> bool
        if not isinstance(other, self.__class__):
            return False
        return self.x == other.x

$ mypy /tmp/a.py
/tmp/a.py:13: error: "object" has no attribute "x"
```

vs.

```
$ cat /tmp/a.py
from typing import Any


class A:
    def __init__(self, x):
        # type: (int) -> None
        self.x = x

    def __eq__(self, other):
        # type: (Any) -> bool
        if not isinstance(other, A):
            return False
        return self.x == other.x
$ mypy /tmp/a.py
$
```

jamesls: Sounds good, I'll add a comment about this before I merge.

PR #833, opened 2018-05-14
Always load env vars before importing the chalice app

jamesls:
This builds on #798 and adds a few tests to ensure we don't regress on this.

This change allows the chalice app to be lazy loaded by setting the chalice app to a callable that can create a chalice app.

I also restructured the local tests.  I've moved the integration local tests over to functional tests.  That way they can be run as part of the travis builds.


chalice/config.py
```diff
@@ -126,7 +127,20 @@ def project_dir(self):
...
+        # There's two value we support.  If the value
+        # is a chalice app, we return it as is.
+        # Otherwise, we assume it's a callable that creates
+        # a chalice app.  This is used to lazy load the chalice
+        # app.
+        if isinstance(v, Chalice):
+            return v
+        elif self._chalice_app is not None:
+            return self._chalice_app
+        app = v()
```
joguSD: Do we want to ensure `v` is callable?

jamesls:
@joguSD Updated.

BigChief45:
@jamesls @joguSD Shouldn't this be added to the changelog?

jamesls:
Yes, good catch, I'll get this updated.

tcatrain:
Hello !

It seems like this pull request was lost somehow over the updates. I had to add this line again to the `chalice/cli/factory.py` file in order to load the environment variables :

```python
def create_config_obj(self, chalice_stage_name=DEFAULT_STAGE_NAME,
        ...
        os.environ.update(config.environment_variables)
```

Is there any reason why it was removed ?
If not, could you please remerge ?
If you need a new pull request, feel free to ask me.

Kind regards,

PR #834, opened 2018-05-14
Skip healthcheck too slow when running on travis

jamesls:
Example of failing build: https://travis-ci.org/aws/chalice/jobs/377921987
Docs: http://hypothesis.readthedocs.io/en/latest/healthchecks.html#hypothesis.HealthCheck

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,11 @@
 language: python
 sudo: false
-env: HYPOTHESIS_PROFILE=ci
 matrix:
   include:
   - python: "3.6"
-    env: TEST_TYPE=prcheck
+    env: TEST_TYPE=prcheck HYPOTHESIS_PROFILE=ci
   - python: "2.7"
-    env: TEST_TYPE=prcheck-py2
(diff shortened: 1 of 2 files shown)


jamesls:
Taking a look at the failure.  This fact that this still fails suggests that either a) hypothesis isn't properly disabling the healthcheck or b) we're not using the correct hypothesis profile on travis.

jamesls:
In our travis.yml file we have:

```
language: python
sudo: false
env: HYPOTHESIS_PROFILE=ci
matrix:
  include:
  - python: "3.6"
    env: TEST_TYPE=prcheck
  - python: "2.7"
    env: TEST_TYPE=prcheck-py2
```

It appears that the `env` from the top level params doesn't get merged in each matrix permutation.  I updated the .travis.yml to just print out the env vars and you can see that the `HYPOTHESIS_PROFILE` isn't in the list of env vars: https://travis-ci.org/jamesls/chalice/jobs/378853146.   I think the easiest thing for now is to just define them for each matrix build

PR #837, opened 2018-05-15
Add VPC support to lambda functions

jamesls:
This adds VPC support to lambda functions in chalice.  It pulls in #673 and fleshes out the rest of the deployer support for VPC.  Notably it makes a few changes:

* The VPC configuration can be added per lambda function and per stage.
* The role management is now moved into the deployer and based on the models.  In order to do this I had to add a `traits` flag to the `AutogenPolicy` which flags which policies it needs to add.
* Update the awsclient to handle removal of VPC configuration (that is, if you configure a function for VPC and then remove it from your config.json, it will appropriately remove VPC configuration).
* Added a few more examples and docs on how to configure VPC.

This PR also triggered a few linter warnings so I cleaned up some of the code (as a separate commit).  There was a whole class (with tests) that was no longer used.  It was a part of the legacy deployer.

Thanks to @lime-green for the original PR.


chalice/deploy/deployer.py
```diff
@@ -664,7 +685,10 @@ def handle_filebasediampolicy(self, config, resource):
     def handle_autogeniampolicy(self, config, resource):
         # type: (Config, models.AutoGenIAMPolicy) -> None
         if isinstance(resource.document, models.Placeholder):
-            resource.document = self._policy_gen.generate_policy(config)
+            policy = self._policy_gen.generate_policy(config)
+            if models.RoleTraits.VPC_NEEDED in resource.traits:
+                policy['Statement'].append(VPC_ATTACH_POLICY)
```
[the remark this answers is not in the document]
jamesls: They're different ARNs.  The config ARNs are for the subnet/sg ids, but the VPC policy is to CRUD ENIs.

jamesls:
@stealthycoin Feedback incorporated.

jamesls:
So (1) ends up being a little tricker than I thought and  there's a number of edge cases.  I was initially adding validation to the `chalice.deploy.validate` module.  It seemed better to fast fail before we even make it to the deployer.  However, it didn't seem like the right thing to do.  For example, what about this:

```
        {
            'app_name': 'myapp',
            'security_group_ids': ['sg-1'],
            'stages': {
                'dev': {
                    'lambda_functions': {
                        'foo': {'subnet_ids': ['sn-1']},
                        'bar': {'subnet_ids': ['sn-1']},
                    }
                },
                "prod": {},
            }
        }
```

Deploying to `dev` is fine, but the `prod` stage would "fail" because there's no `subnet_ids` defined.  But it seems wrong to fail config validation because this is only an issue if you deploy to `prod`.  The `dev` stage is actually fine so I would expect `chalice deploy --stage dev` to just work.  Also, what if I had:

```
        {
            'app_name': 'myapp',
            'security_group_ids': ['sg-1'],
            'stages': {
                'dev': {
                    'lambda_functions': {
                        'foo': {'subnet_ids': ['sn-1']},
                        'bar': {},
                    }
                }
            }
        }
```

but my actual app only has a `foo` function:

```
@app.lambda_function()
def foo(...): pass
```

It doesn't seem like we should say the config is invalid because you `chalice deploy` will actually do what we want: deploy the lambda function in a VPC.

I think what I'm going to do is just make this a deployer error.  If we try to create a `models.LambdaFunction` and we're missing a subnet id or a sg then we'll fail.  The error messages won't be as good but I don't want to fail validation when it doesn't actually matter.

What do you think?

jamesls:
@kyleknap feedback incorporated.  I ended up just adding the validation in the AppGraphBuilder.  That way you can catch the error with `chalice package` and you don't actually have to do a deploy.  The error message you get is:

```
ChaliceBuildError: Invalid VPC params for function 'foo', in order to configure VPC for a Lambda function, you must provide the subnet_ids as well as the security_group_ids, got subnet_ids: [u'subnet-1234', u'subnet-2345'], security_group_ids: []
```

chalice/deploy/deployer.py
```diff
@@ -559,9 +581,10 @@ def _build_lambda_function(self,
             config.app_name, config.chalice_stage, name)
         security_group_ids = config.security_group_ids
         subnet_ids = config.subnet_ids
-        if security_group_ids is None or subnet_ids is None:
+        if security_group_ids is None and subnet_ids is None:
```
kyleknap: Is this if statement still needed? It looks like it gets handled in `_get_vpc_params` and is not necessarily needed.

jamesls:
@kyleknap The two examples I gave in my previous comments give you false positives.  One was for an invalid stage that you're not deploying to and one was config for a non-existent function (defined in config but not in app).  By false positive, I mean the "config validation" would fail, but if you actually were to somehow turn off validation and just let `chalice deploy` (or whatever command) go through it would actually deploy correctly.
This happens anytime the config is loaded too which is just about every command, `local`, `package`, `deploy`, etc.

chalice/deploy/deployer.py
```diff
@@ -555,7 +579,13 @@ def _build_lambda_function(self,
         # type: (...) -> models.LambdaFunction
         function_name = '%s-%s-%s' % (
             config.app_name, config.chalice_stage, name)
-        return models.LambdaFunction(
+        security_group_ids = config.security_group_ids
```
kyleknap: Mentioned this before but I think we can remove lines 582-586 because they are handled in the `_get_vpc_params()` method.

jamesls: This is fixed now.

kyleknap:
Thanks. :ship: again

PR #840, opened 2018-05-16
Ewdurbin package local dir as whl

stealthycoin:
Expands on #653 to implement packaging arbitrary directories on your local machine.

Closes #653

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -626,8 +626,12 @@ class SubprocessPip(object):
             import_string = pip_import_string()
         self._import_string = import_string
 
-    def main(self, args, env_vars=None, shim=None):
-        # type: (List[str], EnvVars, OptStr) -> Tuple[int, Optional[bytes]]
+    def main(self,
+             args,           # List[str]
+             env_vars=None,  # EnvVars
+             shim=None       # OptStr
+             ):
+        # type: (...) -> Tuple[int, OptBytes, OptBytes]
         if env_vars is None:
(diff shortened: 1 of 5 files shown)


jamesls:
Seems reasonable, but it looks like this failed the type checker.

chalice/deploy/packager.py
```diff
@@ -700,6 +714,19 @@ def download_all_dependencies(self, requirements_filename, directory):
                 package_name = match.group(1)
                 raise NoSuchPackageError(str(package_name))
             raise PackageDownloadError(error)
+        stdout = out.decode()
+        match = re.search(self._LINK_IS_DIR_PATTERN, stdout)
+        if match:
+            wheel_package_path = str(match.group(1))
```
jamesls: Is the `str()` needed?  The `stdout` you pass `re.search` is already decoded.

stealthycoin: That was to make mypy happy.

stealthycoin:
```
chalice/deploy/packager.py:729: error: Argument 1 to "build_wheel" of "PipRunner" has incompatible type "unicode"; expected "str"
```

We do the same thing a few lines above for the same reason: https://github.com/aws/chalice/blob/master/chalice/deploy/packager.py#L701

jamesls: Ah that's right, the whole `AnyStr` thing.  Yeah, shipit then (don't forget to squash your commits).

PR #846, opened 2018-05-18
Automatically reload dev server when files change

jamesls:
This adds a reloader to chalice local that's enabled by
default.  A new library, watchdog, is pulled in as a dependency
of chalice.  This allows us to efficiently monitor files
for changes using whatever the most efficient OS specific mechanism
(FSEvents, inotify, etc).  I've verified this works on Windows.

The general approach is borrowed from other frameworks such as
django and flask but the implementation is somewhat different.
Two processes are spun up, a worker process and a monitoring
process.  The monitoring process spawns the worker process and
restarts it if it exits with a special RC.  The worker process
launches the dev server in one thread and a file monitor in
a separate thread.  When changes are detected it shuts down the
server and exits with a special RC.

There's a few differences from the other approaches:

* Avoid the use of sys.exit() directly in the reloader.  This
  made the code hard to test and seems odd that a library
  function is calling sys.exit().  However there is a call
  to `sys.exit()` in the `local()` function because this is
  the recommended way in click for a command to exit with
  a specific RC (see the comments for more info).
* We only monitor the project dir for changes.  The other
  reloaders appear to monitor everything in sys.path, which
  seems like overkill here.  The way I see it, stuff on
  sys.path would only be modified if you're installing new
  packages, in which case you'll have to modify requirements.txt
  as well as something in your app.py (in order to import the
  newly installed module).  I dug around their repos for a bit
  and couldn't find a rationale as to why that much
  monitoring is needed.

A note on coverage: because the reloader works by launching
child processes, this isn't going to be tracked via
coverage, so I expect codecov to fail.
There is some support for tracking coverage across subprocesses, but it
involved messing with sitecustomize.py or .pth files.
See http://coverage.readthedocs.io/en/coverage-4.2/subprocess.html
for more info on that.

Supercedes #706.


jamesls:
@kyleknap Feedback incorporated.  There was also a `tests/unit/cli/test_reloader.py` that looks like it wasn't added back when I rebased.  It's there now so the coverage should look better.

Issue #855, opened 2018-05-30
[proposal] Add support for S3 events


jamesls:
We've received numerous requests for S3 event support, but we never had a tracking issue for it.  As we start to flesh out sources for lambda events, I wanted to make sure we have a consistent approach.

I propose adding an `on_s3_event()` decorator, and that all future event sources use the `on_<service>_<something else>` naming convention.  The input parameter is a small abstraction over the event payload to support the most common use case: downloading the object from S3.

Because we don't yet support managed resources (https://github.com/aws/chalice/issues/516), I initially wanted this first version to support pre-existing S3 buckets.  That is, you tell us what existing bucket you want us to configure, and we'll configure the appropriate s3 bucket notifications for you (along with the needed policy updates).  Here's a complete example of the famous s3 image resizer:

```python
from chalice import Chalice

import boto3

app = Chalice(app_name='s3resize')
s3 = boto3.client('s3')


@app.on_s3_event(bucket='my-preexisting-bucket')
def resize_image(event):
    with tempfile.NamedTemporaryFile('w') as f:
        s3_client.download_file(event.bucket, event.key, f.name)
        resized_file = f.name + '.thumbnail'
        with open(Image.open(f.name)) as image:
            image.thumbnail(tuple(x / 2 for x in image.size))
            image.save(resized_file)
        s3_client.upload_file(filename=resized_file,
                              bucket='my-preexisting-dest-bucket',
                              key='%s-thumbnail' % event.key)
```

That's the entire thing.  You run `chalice deploy`, and assuming the source and dest bucket exists, everything just works.

A few things to note in this example:

* There's an optional `filters` and `events` kwarg in the decorator.  The default `events` is `s3:ObjectCreated:*` so it's omitted from the example.
* The `event` param has a `.bucket` and `.key` attribute to make it easy to download objects from s3.  Note this is assuming a single record is sent in the event payload.

## Concerns

The biggest concern is that due to us not managing the S3 bucket for you, you have to use a pre-existing bucket, which means that the **SAM template generation won't work**.  The notification configuration is a property of an `AWS::S3::Bucket` which means you *have* to create the S3 bucket as part of your CFN template.  I see a few possible options here:

1. Just don't support `on_s3_event` in SAM templates, generate a helpful error message.  This feature would only be available via `chalice deploy`.
[...]

PR #860, opened 2018-06-04
Initial commit of S3 events

jamesls:
This commit adds support for triggering a lambda function based
on an S3 event happening such as an object being created or deleted.
This is configured in chalice through a new `on_s3_event` decorator.
Chalice assumes your s3 bucket already exists.  However, it will
configure bucket configuration and function policies for you.
Given this is an existing bucket, we account for existing
notification configurations already in place, and intelligently
merge in the chalice specific lambda configuration as needed.
Same goes for `chalice delete`, it will only remove the lambda
configuration snippet it added to the S3 bucket notification config.

Before this can be merged, we need a plan for this feature
and `chalice package`.

Because you provide chalice with an existing bucket, we aren't
able to describe this in a CFN template (that is, CFN can't
adopt existing resources into a stack).  As a result, the
`chalice package` command fails.  There's a few options here.
We can implement initial support for managed resources so we
can create the bucket for you, or we can create a custom
cfn resource that can apply this config to an existing bucket.
Implementing managed resources doesn't solve the problem though
of what to do with existing buckets.  I think this is an important
enough use case that I don't want to remove support for it just
because CFN can't support this, so for me it boils down to
trying to make this work with custom resources or just not
support S3 events in `chalice package` until we add support for
managed resources.  Then at least customers would have a path
forward if they wanted to use CFN to deploy their app.  They'd
just have to accept the tradeoff that a new bucket would be
created for them.

*Issue #, if available:*

#855

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -450,6 +450,7 @@ class Chalice(object):
         self.log = logging.getLogger(self.app_name)
         self.builtin_auth_handlers = []
         self.event_sources = []
+        self.s3_events = []
         self.pure_lambda_functions = []
         if env is None:
             env = os.environ
@@ -520,6 +521,27 @@ class Chalice(object):
             self.builtin_auth_handlers.append(auth_config)
             return ChaliceAuthorizer(auth_name, auth_func, auth_config)
         return _register_authorizer
+
+    def on_s3_event(self, bucket, events=None,
+                    prefix=None, suffix=None, name=None):
+        def _register_s3_event(event_func):
+            handler_name = name
+            if handler_name is None:
+                handler_name = event_func.__name__
+            trigger_events = events
+            if trigger_events is None:
+                trigger_events = ['s3:ObjectCreated:*']
+            s3_event = S3EventConfig(
+                name=handler_name,
+                bucket=bucket,
+                events=trigger_events,
+                prefix=prefix,
+                suffix=suffix,
+                handler_string='app.%s' % event_func.__name__,
+            )
+            self.s3_events.append(s3_event)
+            return S3EventHandler(event_func)
+        return _register_s3_event
 
     def schedule(self, expression, name=None):
         def _register_schedule(event_func):
@@ -932,3 +954,36 @@ class LambdaFunction(object):
 
     def __call__(self, event, context):
         return self.func(event, context)
+
+
+class S3EventConfig(object):
+    def __init__(self, name, bucket, events, prefix, suffix, handler_string):
+        self.name = name
+        self.bucket = bucket
+        self.events = events
+        self.prefix = prefix
+        self.suffix = suffix
+        self.handler_string = handler_string
+
+
+class S3EventHandler(object):
+    def __init__(self, handler):
+        self.handler = handler
+
+    def __call__(self, event, context):
+        event_obj = self._convert_to_obj(event)
+        return self.handler(event_obj)
+
+    def _convert_to_obj(self, event_dict):
+        return S3Event(event_dict)
+
+
+class S3Event(object):
+    def __init__(self, event):
+        s3 = event['Records'][0]['s3']
+        self.bucket = s3['bucket']['name']
+        self.key = s3['object']['key']
+        self._original_payload = event
+
+    def to_dict(self):
+        return self._original_payload
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -41,7 +41,7 @@ else:
         return unquote_plus(byte_string).decode(encoding)
     # In python 2 there is a base class for the string types that we can check
     # for. It was removed in python 3 so it will cause a name error.
-    _ANY_STRING = (basestring, bytes)
+    _ANY_STRING = (basestring, bytes)  # noqa
 
 
 def handle_decimals(obj):
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -11,25 +11,37 @@ from collections import defaultdict, Mapping
 
 
 __version__ = '1.3.0'
-
+_PARAMS = re.compile(r'{\w+}')
 
 # Implementation note:  This file is intended to be a standalone file
 # that gets copied into the lambda deployment package.  It has no dependencies
 # on other parts of chalice so it can stay small and lightweight, with minimal
-# startup overhead.
+# startup overhead.  This also means we need to handle py2/py3 compat issues
+# directly in this file instead of copying over compat.py
+if sys.version_info.major == 3:
+    from urllib.parse import unquote_plus  # pylint: disable=E0611,E0401
 
+    unquote_str = unquote_plus
 
-_PARAMS = re.compile(r'{\w+}')
-
-try:
-    # In python 2 there is a base class for the string types that
-    # we can check for. It was removed in python 3 so it will cause
-    # a name error.
-    _ANY_STRING = (basestring, bytes)
-except NameError:
     # In python 3 string and bytes are different so we explicitly check
     # for both.
     _ANY_STRING = (str, bytes)
+else:
+    from urllib import unquote_plus
+
+    # This is borrowed from botocore/compat.py
+    def unquote_str(value, encoding='utf-8'):
+        # In python2, unquote() gives us a string back that has the urldecoded
+        # bits, but not the unicode parts.  We need to decode this manually.
+        # unquote has special logic in which if it receives a unicode object it
+        # will decode it to latin1.  This is hard coded.  To avoid this, we'll
+        # encode the string with the passed in encoding before trying to
+        # unquote it.
+        byte_string = value.encode(encoding)
+        return unquote_plus(byte_string).decode(encoding)
+    # In python 2 there is a base class for the string types that we can check
+    # for. It was removed in python 3 so it will cause a name error.
+    _ANY_STRING = (basestring, bytes)
 
 
 def handle_decimals(obj):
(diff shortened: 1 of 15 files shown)


artkay:
@jamesls this is awesome feature!

Is there a plan to deploy this to multiple buckets or to support wildcards in the bucket names? I.e. I would like to do something like:

```python
@app.on_s3_event('mybucket-*', events=['s3:ObjectCreated:Put'],
                 prefix='images/', suffix='.jpg')
```

And it would match `mybucket-123`, `mybucket-test` etc.?

jamesls:
> Is there a plan to deploy this to multiple buckets or to support wildcards in the bucket names?

That's an interesting idea that I hadn't considered.  I'm worried that using a wildcard could result in unintentional configuration, or that `chalice deploy` could configure a new s3 bucket even though nothing in your `app.py` has changed (e.g if you created a new s3 bucket matching the wildcard pattern inbetween chalice deploy).

What about something more explicit where you can give it a list of buckets to configure?

```python
@app.on_s3_event(['mybucket-123', 'mybucket-test'], ...)
```

Not quite as streamlined (you'd have to make the `aws s3 ls` call yourself and copy the bucket values into the `app.py`), but it is more explicit and there wouldn't be any confusion as to which buckets we'll configure for you.

kyleknap:
>> What about something more explicit where you can give it a list of buckets to configure?

I would be in favor of making it an explicit list as well. I think we should take this into consideration for all resource events we create. I can definitely see it being used for hooking up multiple SQS queues or DynamoDB tables to the same lambda function.

Not sure if it needs to be added in this PR, but it would be good to think about now so we can add it in a backwards compatible way. @jamesls in your example, were you thinking of just overloading the `bucket` parameter to be a string or a list?

chalice/app.py
```diff
@@ -932,3 +954,36 @@ def __init__(self, func, name, handler_string):
 
     def __call__(self, event, context):
         return self.func(event, context)
+
+
+class S3EventConfig(object):
+    def __init__(self, name, bucket, events, prefix, suffix, handler_string):
+        self.name = name
+        self.bucket = bucket
+        self.events = events
```
kyleknap: Playing around with the S3 API, it looks like it is possible to have two different lambda configurations to the same lambda function. Specifically, using the CLI some thing like this:
```
$ aws s3api get-bucket-notification-configuration --bucket chalice-s3-notify
{
    "LambdaFunctionConfigurations": [
        {
            "LambdaFunctionArn": "arn:aws:lambda:us-west-2:934212987125:function:s3-notify-dev-index", 
            "Id": "OWQxODY5Y2YtZjA3MC00M2IyLTgwZDItYTg4ODA3MmZlYjdl", 
            "Events": [
                "s3:ObjectCreated:*"
            ]
        }, 
        {
            "Filter": {
                "Key": {
                    "FilterRules": [
                        {
                            "Name": "Prefix", 
                            "Value": "some-prefix"
                        }
                    ]
                }
            }, 
            "LambdaFunctionArn": "arn:aws:lambda:us-west-2:934212987125:function:s3-notify-dev-index", 
            "Id": "YzQwNzc4OTktYzUzNi00ZjAyLWFhYTItOTBlNGNlNWYyN2E1", 
            "Events": [
                "s3:ObjectRemoved:*"
            ]
        }
    ]
}
```
Currently, we cannot support this configuration through the `on_s3_event()` because prefix and suffix rules are applied to a single list of events. To get this to work, it seems like we would need to expose a `configurations` parameter that allows to specify events and suffixes/prefixes per element in the `configurations` list. Does it make sense to support this? The one downside I see is that it makes it a more complicated interface.

docs/source/topics/events.rst
```diff
@@ -66,3 +68,60 @@ event rule.  For example:
...
+    from chalice import Chalice
+
+    app = chalice.Chalice(app_name='s3eventdemo')
+    s3 = boto3.client('s3')
+
+    @app.on_s3_event(bucket='mybucket-name',
+                     events=['s3:ObjectCreated:*'])
+    def handle_s3_event(event):
+        # Download the newly created file locally.
+        s3.downlad_file(event.bucket, event.key, '/tmp/tempfile')
```
kyleknap: Typo: `s3.downlad_file` --> `s3.download_file`. It may be good to try both the examples in the reference and topic guide work if you have not already tried them.

tests/unit/deploy/test_planner.py
```diff
@@ -284,6 +284,55 @@ def test_can_set_variables_when_needed(self):
...
+        bucket_event = models.S3BucketNotification(
+            resource_name='function_name-s3event',
+            bucket='mybucket',
+            events=['s3:ObjectCreated:*'],
+            prefix=None,
+            suffix=None,
+            lambda_function=function,
+        )
+        plan = self.determine_plan(bucket_event)
+        self.assert_apicall_equals(
```
kyleknap: My memory is a little fuzzy. Why do we have these separate `assert_apicall_equals` instead of just having a single `assert plan == expected_plan`?

jamesls:
> were you thinking of just overloading the bucket parameter to be a string or a list?

Yeah, bucket would be `Union[List[str], str]`.  I think I'd save that for a follow up PR.  It's a more involved change and this PR is already large.

BigChief45:
@jamesls  I assume this change will be in the next release? Would be good to add it to the changelog.

chalice/app.py
kyleknap: Yeah seems reasonable.

artkay:
> What about something more explicit where you can give it a list of buckets to configure?

This would save us bunch of functions for sure. What you have in this PR is already very helpful.

> Not sure if it needs to be added in this PR, but it would be good to think about now so we can add it in a backwards compatible way.

Totally - just having stable Chalice with S3 events would be awesome. For now I forked Chalice into our organization and merged the PR to use it for PoC development.

> I would be in favor of making it an explicit list as well.

I understand the concern. We definitely do not need to create S3 buckets from the app, but we have a use case which will benefit from wildcards a lot. We have list of clients and some of them (not everyone!) are getting their personal S3 bucket to share some content with our system. We need Chalice to parse every new uploaded file for each client and we do not know the list of buckets ahead of time - we just have the bucket naming convention. Having explicit list of buckets will require us to re-deploy Chalice app every time new bucket is created - and we are expecting more and more clients using this feature. So having a wildcard will reduce amount of deployments dramatically.

artkay:
I just double checked and you guys are right - Lambda won't allow to create a trigger with a wildcard, at least I was unable to do so unfortunately.

chalice/awsclient.py
```diff
@@ -723,6 +723,109 @@ def add_permission_for_scheduled_event(self, rule_arn,
...
+        # type: (str, str) -> None
+        lambda_client = self._client('lambda')
+        policy = self.get_function_policy(function_arn)
+        bucket_arn = 'arn:aws:s3:::%s' % bucket
+        if self._policy_gives_access(policy, bucket_arn, 's3'):
+            return
+        random_id = self._random_id()
+        # We should be checking if the permission already exists and only
+        # adding it if necessary.
+        lambda_client.add_permission(
```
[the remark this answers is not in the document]
jamesls: Trying to think through how this would happen, if you removed the decorator then the lambda function wouldn't be part of the app anymore, and we'll clean up "properly" because we delete the entire lambda function.  Were you able to get this to leak somehow?  Did you change from `on_s3_event` to `@app.lambda_function()`?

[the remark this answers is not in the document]
kyleknap: Yeah I changed it from `on_s3_event` to `lambda_function`.

jamesls:
@kyleknap Feedback incorporated.  The only thing I didn't incorporate was handling removing permissions if you switch from `on_s3_event` to `lambda_function`.  The problem is more general (we don't handle this for scheduled events to pure lambda function or from API handler to pure lambda function) and will probably end up being a sizable PR so I'd prefer to defer that until later.

kyleknap:
Sounds good about handling the cleanup for later. Might be good to open an issue to track it. So we do not forget.

kyleknap:
Looks good. One more thing that I did notice was that the key is URI encodeded. So using this app.py:
```python
from chalice import Chalice

app = Chalice(app_name='s3-notify')
app.debug = True


@app.on_s3_event(bucket='chalice-s3-notify')
def index(event):
    app.log.debug(event.to_dict())
    return {'hello': event.bucket}
```
I found these logs contained: `u'object': {u'eTag': u'32ff4c9dbd6a328b22031e49b44fe121', u'sequencer': u'005B20811E96F0DC9C', u'key': u'foo+bar', u'size': 284}`

So the question now is do we want to keep it URI encoded? I think for the `event.key` we should URI decode it for the user, but in the `to_dict()` method we preserve it just as it came to the Lambda function. Thoughts?

jamesls:
@kyleknap Good catch.  I've incorporated the feedback.  The `.key` attr automatically urldecodes the key for you, but `to_dict()` keeps everything unchanged.

jamesls:
Hmm, so it looks like a limitation of the linters.  Looks like flake8 is smart enough to check `except` clauses, but doesn't handle conditionals:

```python
try:
    print("The type is: %s" % basestring)
except NameError:
    print("Does not exist")
```

```
$ flake8 /tmp/foo.py
$
```

But if you change to check for python versions:

```python
import sys

if sys.version_info.major == 2:
    print("The type is: %s" % basestring)
else:
    print("Does not exist")
```

If now fails:

```
$ flake8 /tmp/foo.py
/tmp/foo.py:4:31: F821 undefined name 'basestring'
```

I can put it back how it was, but we're relying on the fact that the first line in the `try` will trigger a `NameError`, which could break if we reshuffle the order around.

I'm inclined to just put a `# noqa` on that line.

jamesls:
Looks like there's some flaky issues with the doc link checker.  It passed on the travis run of my fork (https://travis-ci.org/jamesls/chalice/builds/392416776).  I'll restart the job...

samzoozi:
Is there any way to do the same thing using Chalice-cdk? when I use the decorator to capture s3 notifications for a lambda and try `cdk deploy` it give me the following error
`NotImplementedError: Unable to package chalice apps that @app.on_s3_event decorator. CloudFormation does not support modifying the event notifications of existing buckets. You can deploy this app using `chalice deploy`.

Wondering if there is a workaround for that?

PR #862, opened 2018-06-05
Update type hints

jamesls:
This contains two changes:

* Take advantage of mypy's built in support for attrs.  We don't need a separate models.pyi file anymore, you can now add type hints to attr classes directly as of the latest mypy version.
* Update `app.pyi` with the latest type hints.  This file was really out of date.  mypy only type checks `app.py` against code that imports the module, not within the `app.py` file itself.  I noticed this when using pycharm, which has its own type checking implementation and can highlight issues between `app.py` and `app.pyi`.  There's still a few things that pycharm flags, but they're all related to internal methods not having type definitions, and I'm not sure that's worth the effort.


chalice/deploy/planner.py
```diff
@@ -187,12 +187,17 @@ def _plan_lambdafunction(self, resource):
         # Make mypy happy, it complains if we don't "declare" this upfront.
         params = {}  # type: Dict[str, Any]
         varname = '%s_lambda_arn' % resource.resource_name
+        # Not sure the best way to express this via mypy, but we know
+        # that in the build stage we replace the deployment package
+        # name with the actual filename generated from the pip
+        # packager.  For now we resort to a cast.
+        filename = cast(str, resource.deployment_package.filename)
```
joguSD: Might be worthwhile to add a TODO in the comment so it's easy to find places we might want to periodically check back on when mypy updates and begins to support new things. Though, I suppose just looking for where we cast might also be sufficient.

PR #869, opened 2018-06-14
Compile .pyc bytecode cache files when installing wheels

### LINKED ISSUE #868: Compile .pyc files when unpacking wheels into distribution
Wheels don't carry .pyc files, but the Lambda distribution file probably should, to avoid incurring a startup time penalty. `chalice.deploy.packager.DependencyBuilder._install_wheels` should include a call to `compileall.compile_dir`.

### LINKED ISSUE #1314: Add ability to annotate that a function needs access to S3 bucket
We have a function that subscribes to an SNS topic, that broadcasts changes to an S3 bucket. When the function is triggered, it gets the S3 object details and attempts to access the object. However, Chalice does not automatically configure the correct S3 permissions, since we have not indicated that the function needs S3 access.

It would be helpful to somehow 'tell' Chalice that 'inside this function, we need to access S3', so it could automatically configure the correct permissions. This could be done, for example, by a decorator.

kislyuk:
Fixes #868  https://github.com/HumanCellAtlas/data-store/pull/1314

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -3,6 +3,7 @@ import hashlib
 import inspect
 import re
 import subprocess
+import compileall
 from email.parser import FeedParser
 from email.message import Message  # noqa
 from zipfile import ZipFile  # noqa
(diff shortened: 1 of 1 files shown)


kyleknap:
Thanks for the patch and the performance testing deep dive. Looking over the results from https://github.com/aws/chalice/issues/868, there seemed to be not too strong of a performance reason to start compiling the `.pyc` files so I'm hesitant to pull this in right now. For now, I'm closing the pull request to help keep a more accurate list of pull requests we are reviewing on the GitHub repository. However, I am keeping the tracking [GitHub issue open](https://github.com/aws/chalice/issues/868) open. If there ever becomes more +1's on that issue and more people running into performance issues, I say we reopen this PR and get the code merged!

PR #875, opened 2018-06-21
Feature/configure reserved concurrency

### LINKED ISSUE #791: Allow configuration of function concurrency
For high volume Lambda invocations in accounts with multiple Lambdas, it's best practice to set per-function concurrency limits (https://docs.aws.amazon.com/lambda/latest/dg/concurrent-executions.html). This can be done using https://docs.aws.amazon.com/lambda/latest/dg/API_PutFunctionConcurrency.html.

It would be great to have this setting be available alongside things like memory and timeout settings in `config.json`.

xplorld:
Fix #791

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -281,6 +281,20 @@ class TypedAWSClient(object):
             )
             raise self._get_lambda_code_deployment_error(e, context)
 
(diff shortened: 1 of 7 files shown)


jamesls:
Awesome!  Thanks for the pull request, looking over this now.

PR #878, opened 2018-06-29
Add support SNS subscriptions to Lambda

### LINKED ISSUE #488: Pure Lambda: support for SNS trigger
We have a number of pure lambda functions that we deploy through chalice (@app.lambda_function).  

It would be good to be able to specify an sns trigger for the pure lambda function by specifying the sns arn (similar to [serverless sns events]( https://serverless.com/framework/docs/providers/aws/events/sns/))

`functions:
  dispatcher:
    handler: dispatcher.dispatch
    events:
      - sns: arn:xxx`

jamesls:
This adds support for an `on_sns_message` decorator, which allows you to subscribe a lambda function to an SNS topic.  Similar to the S3 events, the SNS topic must already exist.  We have a tracking issue for chalice to manage resources, but for now you provide the topic name you want us to subscribe to.

On the plus side, the `chalice package` command works, because the SNS subscriptions are separate cfn resources.

The PR includes docs along with a tutorial on how to use this feature, but here's a simple demo that creates a lambda function and hooks it up to the SNS topic `mytopic`:

```python
from chalice import Chalice

app = Chalice(app_name='testsns')

@app.on_sns_message(topic_name='mytopic')
def index(event):
    print("Message: %s" % event.message)
    print("Subject: %s" % event.subject)
    print("to_dict: %s" % event.to_dict())
```

There are a few changes that I'd like to make, but it won't affect the existing logic in this PR.  I opted to send the PR it in its current form, as all of the changes are additive and easier to review.  I also tried to break the commits down logically if you'd prefer to go commit by commit.  Planned changes:

* Refactor the event handler code to remove duplication.  The s3 event, scheduled event, and sns handler have similar logic that can be consolidated.
* Add support for SNS topic ARN.  The plan would be to overload the `topic` param to be either a topic name or a topic ARN.  This is an additive change but it does complicate the code so I'll likely hold off on making that change.

Closes #488


tests/unit/deploy/test_planner.py
```diff
@@ -672,6 +798,51 @@ class Foo(models.ManagedModel):
         with pytest.raises(ValueError):
             self.remote_state.resource_exists(foo)
 
+    @pytest.mark.parametrize(
+        'resource_topic,deployed_topic,resource_exists,expected_result', [
```
kyleknap: I wonder if it makes sense to reword `resource_exists` to something to indicate it is the return value from `verify_sns_subscription_current`? I got hung up a little bit on this as just reading the parameter names it seemed redundant to have a `resource_exists` and `expected_result` parameter for something that was testing if the subscription exists. It seemed like only one would have been needed.

chalice/deploy/deployer.py
```diff
@@ -392,24 +392,42 @@ def build(self, config, stage_name):
...
-            bucket_notification = self._create_bucket_notification(
-                config, deployment, s3_event, stage_name)
-            resources.append(bucket_notification)
         return models.Application(stage_name, resources)
 
+    def _create_lambda_event_resources(self, config, deployment, stage_name):
+        # type: (Config, models.DeploymentPackage, str) -> List[models.Model]
+        resources = []  # type: List[models.Model]
+        for event_source in config.chalice_app.event_sources:
+            if isinstance(event_source, app.S3EventConfig):
```
joguSD: I'm assuming this is one of the places you're looking to remove duplication?

chalice/deploy/planner.py
```diff
@@ -337,6 +365,100 @@ def _plan_managediamrole(self, resource):
...
+            models.StoreValue(
+                name=topic_arn_varname,
+                value=StringFormat(
+                    'arn:aws:sns:{region_name}:{account_id}:%s' % (
+                        resource.topic
+                    ),
+                    ['region_name', 'account_id'],
+                ),
+            ),
+        ]  # type: List[_INSTRUCTION_MSG]
```
joguSD: Does this require the type annotation?

docs/source/topics/events.rst
```diff
@@ -122,4 +122,91 @@ configuration needed for the ``handle_s3_event`` Lambda function.
...
+
+    $ chalice deploy
+    Creating deployment package.
+    Creating IAM role: chalice-demo-sns-dev
+    Creating lambda function: chalice-demo-sns-dev-handle_sns_message
+    Subscribing chalice-demo-sns-dev-handle_sns_message to SNS topic my-demo-topic
+    Resources deployed:
+      - Lambda ARN: arn:aws:lambda:us-west-2:123:function:...
+
+And now we can test our app by publishing a few SNS message to our topic.
```
joguSD: messages*

joguSD:
Also it looks like the build is failing because `deployer.py` is too long (max 1k lines).

chalice/app.pyi
```diff
@@ -209,10 +199,25 @@ class LambdaFunction(object):
...
     name = ... # type: str
     bucket = ... # type: str
     events = ... # type: List[str]
     prefix = ... # type: str
     suffix = ... # type: str
     handler_string = ... # type: str
+
+
+class SNSEventConfig(BaseEventSourceConfig):
+    name = ... # type: str
```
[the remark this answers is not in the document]
jamesls: Not needed, I've removed.

chalice/deploy/deployer.py
jamesls: No, this wasn't one of the places I had in mind.  There's not much duplication other than sharing the appending to a list.  The only thing I could see changing in the future is replacing the if/else chain with dynamic dispatch, but 3 if's is ok with me for now.

chalice/deploy/planner.py
jamesls: Yes, otherwise it infers the type based on the elements in the list, and it will incorrectly infer these all to be `Instruction` types because we're not using the tuple form that could also include an optional message we want displayed.

jamesls:
@kyleknap @joguSD ready to look at again.

For removing policies, I only added support for SNS and S3 events.  Given we've shipped scheduled events, api gateway, and custom authorizers, I wanted to spend a little more time vetting those to make sure we won't break people.  Plus this PR's already large.

kyleknap:
@jamesls Changes look fine to me. But it looks like we are still orphaning permissions for scheduled events. I think the awsclient method for it was added but it was never plumbed through to the planner/sweeper.

jamesls:
Yeah I mentioned why in my previous comment.  Are you saying you disagree with what I've proposed?

kyleknap:
Nevermind. I did not see the previous comment. I posted mine soon after you had. That's fine then. Looking over it, it just looked like something you forgot to add given you made the client method for it.

jamesls:
Oh, if that's the issue I don't mind removing the scheduled event method from awsclient until I add support for it.

kyleknap:
I do not feel strongly either way. We are going to need it in the future. I am fine keeping it for now. It is not like it is a lot of extra code.

jamesls:
Yeah I went ahead and removed it.  I think it makes sense to add it once we add support for removing policies of scheduled events.

Issue #884, opened 2018-07-03
[proposal] Add support for SQS event source


jamesls:
This issue proposes adding support for the newly released sqs event source.

While the implementation will be different, the user facing API is almost identical to the [s3 events](https://github.com/aws/chalice/issues/855) and [sns events](https://github.com/aws/chalice/issues/488) we recently added.

### app.py changes

First, we'll add an `@app.on_sqs_message` decorator, which takes a `queue` name and an optional `batch_size` (default is 10, which is the same in the console):

```python

@app.on_sqs_message(queue='myqueue', batch_size=10)
def handler(event):
    for record in event:
        print(record.body)
        print(record.to_dict())
```

Similar to the `topic` arg in the SNS PR, this allows us to eventually add support for Queue URLs and possible queue resources if needed (verified that queue urls and queue names are distinct).

In terms of the event object being passed to the handler, it will still inherit from `BaseLambdaEvent` except that it will also be iterable.  Because we can now get multiple records in a single event, there's no helpers for accessing just the first record.  Each item we iterate over is itself a `BaseLambdaEvent` and will have these values:

```python

class SQSRecord(BaseLambdaEvent):
    body = ...  # type: str

```

I propose making each item a `BaseLambdaEvent` so you also get a `to_dict()` method on each record.

Here's what the raw event looks like (showing just a single record but there can be multiple records):

```
{'Records': [{
        'attributes': {
            'ApproximateFirstReceiveTimestamp': '1530576251596',
            'ApproximateReceiveCount': '1',
            'SenderId': 'sender-id',
            'SentTimestamp': '1530576251595'
        },
        'awsRegion': 'us-west-2',
        'body': 'queue message body',
        'eventSource': 'aws:sqs',
        'eventSourceARN': 'arn:aws:sqs:us-west-2:12345:queue-name',
        'md5OfBody': '754ac2f7a12df38320e0c5eafd060145',
        'messageAttributes': {},
        'messageId': 'message-id',
        'receiptHandle': 'handle'
}]}
```

kadrach:
> ... add support for [...] queue resources

My first thought on seeing both SNS and SQS support was deployment. In a SAM template, I can add CF resources and reference them, e.g. have the queue created as part of the stack or import a value from another stack.

Is this the intent of managed resources here?

kyleknap:
[...]

PR #886, opened 2018-07-05
Add support for SQS events

jamesls:
This implements #884.  It's similar to s3 and sns events in terms of the interface.  The implementation is slightly different because unlike s3 and sns where you set up the configuration in the service emitting the events, you configure sqs integration through the lambda API and its event source mapping.

In addition, because the configuration happens through  lambda, the permissions need to be hooked up through the IAM role we create for you.  Therefore if you use autogen IAM policy, we'll insert a snippet that gives the role access to sqs (this is the `LambdaEventSourcePolicyInjector` class I've added to the build steps).  This is similar to what we do for permissions to cloudwatch logs https://github.com/aws/chalice/blob/master/chalice/constants.py#L72-L80.

Closes https://github.com/aws/chalice/issues/884

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -891,6 +891,56 @@ class TypedAWSClient(object):
                     StatementId=statement['Sid'],
                 )
 
+    def create_sqs_event_source(self, queue_arn, function_name, batch_size):
+        # type: (str, str, int) -> None
+        lambda_client = self._client('lambda')
+        return lambda_client.create_event_source_mapping(
+            EventSourceArn=queue_arn,
+            FunctionName=function_name,
+            BatchSize=batch_size,
+        )['UUID']
+
+    def update_sqs_event_source(self, event_uuid, batch_size):
+        # type: (str, int) -> None
+        lambda_client = self._client('lambda')
+        lambda_client.update_event_source_mapping(
+            UUID=event_uuid,
+            BatchSize=batch_size,
+        )
+
+    def remove_sqs_event_source(self, event_uuid):
+        # type: (str) -> None
+        lambda_client = self._client('lambda')
+        lambda_client.delete_event_source_mapping(UUID=event_uuid)
+
+    def verify_event_source_current(self, event_uuid, resource_name,
+                                    service_name, function_arn):
+        # type: (str, str, str, str) -> bool
+        """Check if the uuid matches the resource and function arn provided.
+
+        Given a uuid representing an event source mapping for a lambda
+        function, verify that the associated source arn
+        and function arn match up to the parameters passed in.
+
+        Instead of providing the event source arn, the resource name
+        is provided along with the service name.  For example, if we're
+        checking an SQS queue event source, the resource name would be
+        the queue name (e.g. ``myqueue``) and the service would be ``sqs``.
+
+        """
+        client = self._client('lambda')
+        try:
+            attributes = client.get_event_source_mapping(UUID=event_uuid)
+            actual_arn = attributes['EventSourceArn']
+            arn_start, actual_name = actual_arn.rsplit(':', 1)
+            return (
+                actual_name == resource_name and
+                arn_start.startswith('arn:aws:%s' % service_name) and
+                attributes['FunctionArn'] == function_arn
+            )
+        except client.exceptions.ResourceNotFoundException:
+            return False
+
     def _random_id(self):
         # type: () -> str
         return str(uuid.uuid4())
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -894,10 +894,13 @@ class TypedAWSClient(object):
     def create_sqs_event_source(self, queue_arn, function_name, batch_size):
         # type: (str, str, int) -> None
         lambda_client = self._client('lambda')
-        return lambda_client.create_event_source_mapping(
-            EventSourceArn=queue_arn,
-            FunctionName=function_name,
-            BatchSize=batch_size,
+        kwargs = {
+            'EventSourceArn': queue_arn,
+            'FunctionName': function_name,
+            'BatchSize': batch_size,
+        }
+        return self._call_client_method_with_retries(
+            lambda_client.create_event_source_mapping, kwargs
         )['UUID']
 
     def update_sqs_event_source(self, event_uuid, batch_size):
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -405,6 +405,12 @@ class ApplicationGraphBuilder(object):
                         config, deployment, event_source, stage_name
                     )
                 )
+            elif isinstance(event_source, app.SQSEventConfig):
+                resources.append(
+                    self._create_sqs_subscription(
+                        config, deployment, event_source, stage_name,
+                    )
+                )
         return resources
 
     def _create_rest_api_model(self,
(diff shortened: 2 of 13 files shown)


chalice/package.py
```diff
@@ -303,6 +303,28 @@ def _generate_snslambdasubscription(self, resource, template):
...
+            sns_cfn_name: {
+                'Type': 'SQS',
+                'Properties': {
+                    'Queue': {
+                        'Fn::Sub': (
+                            'arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:%s' %
+                            resource.queue
+                        )
+                    },
+                    'BatchSize': 5,
```
joguSD: Is this supposed to be hard coded to 5?

chalice/deploy/deployer.py
```diff
@@ -281,7 +282,8 @@ def create_build_stage(osutils, ui, swagger_gen):
             ),
             SwaggerBuilder(
                 swagger_generator=swagger_gen,
-            )
+            ),
+            LambdaEventSourcePolicyInjector(),
```
joguSD: Is this added to the managed policy even if they're not using SQS events?

jamesls: No, the base class does dynamic dispatch based on the class type, so [handle_sqseventsource](https://github.com/aws/chalice/pull/886/files/24ae5b0d0dcd0be19fd69fdff8c5fb2fef12e288#diff-a4c0c52aa7a3b2d98c797c244f0f0e8fR766) is only invoked if that resource is part of your app graph.

chalice/package.py
jamesls: 🤦, I'll fix that.

jamesls:
Pushed up docs, should be good to go now.

chalice/awsclient.py
```diff
@@ -172,25 +172,12 @@ def _create_lambda_function(self, api_args):
...
-                attempts += 1
-                if attempts >= self.LAMBDA_CREATE_ATTEMPTS or \
-                        not self._is_iam_role_related_error(e):
-                    raise
-                continue
-            return response
+    def _is_settling_error(self, error):
+        # type: (botocore.exceptions.ClientError) -> bool
+        message = error.response['Error'].get('Message', '')
+        if re.search('event source mapping.*is in use', message):
```
joguSD: Do you think it would be better to check the `Code`?
```python
response['Error'].get('Code', '') == 'ResourceInUseException'
```

docs/source/api.rst
```diff
@@ -243,6 +243,52 @@ Chalice
...
+                  app.log.info("Message body: %s", record.body)
+
+      :param queue: The name of the SQS queue you want to subscribe to.
+        This is the name of the queue, not the ARN or Queue URL.
+
+      :param batch_size: The maximum number of messages to retrieve
+        when polling for SQS messages.  The event parameter can have
+        multiple SQS messages associated with it.  This is why the
+        event parameter passed to the lambda function is iterable.  The
+        batch size controls how many messages can be in a single event.
```
joguSD: It's in the function signature above but it might be worth mentioning that we default this to 1 here as well.

chalice/awsclient.py
kyleknap: Yeah I would like to know the reasoning of going with message approach as well.

jamesls: This is handled by catching the `ResourceInUseException` here: https://github.com/aws/chalice/pull/886/files/0096cd0ce45857ad2cfad23e8138c50055ca17a0#diff-3aa85da792f26ee5a92aff6a3d060cccR957.

If we knew that this error was only raised when the event source mapping is still settling, then we could remove this check, but I'm not sure if that's a safe assumption.

docs/source/api.rst
jamesls: Pushed a change for this.

chalice/awsclient.py
```diff
@@ -891,6 +880,99 @@ def _remove_lambda_permission_if_needed(self, source_arn, function_arn,
...
+        client = self._client('lambda')
+        attempts = 0
+        if should_retry is None:
+            should_retry = self._is_iam_role_related_error
+        retryable_exceptions = (
+            # We're assuming that if we receive an
+            # InvalidParameterValueException, it's because the role we just
+            # created can't be used by Lambda so retry until it can be.
+            client.exceptions.InvalidParameterValueException,
+            client.exceptions.ResourceInUseException,
```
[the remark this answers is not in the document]
jamesls: Good catch, updated.

docs/source/api.rst
```diff
@@ -243,6 +243,52 @@ Chalice
...
 
+   .. method:: on_sqs_message(queue, batch_size=1, name=None)
+
+      Create a lambda function and configure it to be automatically invoked
+      whenever a message is published to the specified SQS queue.
+
+      The lambda function must accept a single parameter which
+      is of type :class:`SQSEvent`.
+
+      If the decorated function returns without raising any exceptions
```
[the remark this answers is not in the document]
jamesls: Yeah I ran into this as well and meant to document.  I think this makes the most sense in the topic guide rather than the API docs for `on_sqs_message`, thoughts?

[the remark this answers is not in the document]
kyleknap: Yeah that sounds good to me. It is pretty annoying that the default queue you create in the console is incompatible with the default chalice lambda function. It would be nice if we accounted for that with managed resources.

chalice/awsclient.py
kyleknap: I say we keep the message check in place as well. I'm not sure if it is a safe assumption to make either.

chalice/deploy/deployer.py
```diff
@@ -733,6 +762,31 @@ def handle_restapi(self, config, resource):
...
+        # type: () -> None
+        self._policy_injected = False
+
+    def handle_sqseventsource(self, config, resource):
+        # type: (Config, models.SQSEventSource) -> None
+        # The sqs integration works by polling for
+        # available records so the lambda function needs
+        # permission to call sqs.
+        role = resource.lambda_function.role
+        if (not self._policy_injected and
```
kyleknap: Hmm this is a pretty long if statement. I wonder if it is possible to simplify this?

tests/functional/test_awsclient.py
```diff
@@ -1699,3 +1699,189 @@ def test_can_remove_s3_permission(stubbed_session):
...
+    assert result == 'my-uuid'
+
+    stubbed_session.verify_stubs()
+
+
+def test_can_delete_sqs_event_source(stubbed_session):
+    lambda_stub = stubbed_session.stub('lambda')
+    lambda_stub.delete_event_source_mapping(
+        UUID='my-uuid',
+    ).raises_error(
```
kyleknap: It looks like it was not intended to put this raises `ResourceInUseException`? The next test looks the same as the next one.

docs/source/api.rst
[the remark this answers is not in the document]
jamesls: Yeah I pushed a change with doc updates.

tests/functional/test_awsclient.py
jamesls: Yeah good catch, I meant to have one without retries, one with.

chalice/deploy/deployer.py
jamesls: I originally had it simplified, but mypy's flow control doesn't seem to like it and I'd prefer not to explicitly cast.

```python
    def handle_sqseventsource(self, config, resource):
        # type: (Config, models.SQSEventSource) -> None
        # The sqs integration works by polling for
        # available records so the lambda function needs
        # permission to call sqs.
        role = resource.lambda_function.role
        if self._policy_injected:
            # We only need to inject the policy one.
            return
        is_correct_type = (
            isinstance(role, models.ManagedIAMRole) and
            isinstance(role.policy, models.AutoGenIAMPolicy) and
            not isinstance(role.policy.document, models.Placeholder)
        )
        if is_correct_type:
(line 785)  self._inject_trigger_policy(role.policy.document,
                                        SQS_EVENT_SOURCE_POLICY.copy())
            self._policy_injected = True
```

```
$ make typecheck
mypy --py2 --ignore-missing-imports --follow-imports=skip -p chalice --disallow-untyped-defs --strict-optional --warn-no-return
chalice/deploy/deployer.py:785: error: "IAMRole" has no attribute "policy"
make: *** [typecheck] Error 1
```

chalice/awsclient.py
joguSD: Make sense.

chalice/deploy/deployer.py
kyleknap: Yeah that does look better. Unfortunate about the mypy problem. I'm fine keeping it as is.

PR #888, opened 2018-07-09
Give a call out to other event sources in the README

jamesls:
Hopefully this raises awareness of other event sources
chalice can integrate with.


jamesls:
Rendered here if you want to take a look: https://github.com/jamesls/chalice/tree/readme-refresh

PR #889, opened 2018-07-09
Add support for S3 upload_file/download_file/copy in policy generator

jamesls:
This adds a `policy-extra.json` file that's used to track additional methods we add to clients and their corresponding policies needed.

In addition, I've updated the ast transformer to also traverse into the new decorators we added.

The end result is that using `@app.on_s3_event()`, we can detect that uses of `s3.download_file()` should give you `s3:GetObject` permissions.


tests/unit/test_policy.py
```diff
@@ -158,3 +158,19 @@ def test_no_changes():
     first = iam_policy({'s3': {'list_buckets', 'list_objects'}})
     second = iam_policy({'s3': {'list_buckets', 'list_objects'}})
     assert diff_policies(first, second) == {}
+
+
+def test_can_handle_high_level_abstractions():
```
stealthycoin: Can we test all the abstractions supported.

chalice/policy.py
```diff
@@ -136,3 +149,40 @@ def _build_statements_from_client_calls(self, client_calls):
...
+    def _get_actions_from_api_calls(self, service, client_calls):
+        # type: (str, Dict[str, Set[str]]) -> List[str]
+        if service not in self._api_policy_actions:
+            print("Unsupported service for auto policy generation: %s"
+                  % service)
+            return []
+        service_actions = self._api_policy_actions[service]
+        method_calls = client_calls[service]
+        # Next thing we need to do is convert the method_name to
+        # MethodName.  To this reliable we're going to use
```
joguSD: To do this*

kyleknap:
We probably should include `upload_fileobj` and `download_fileobj` as well.

Issue #900, opened 2018-07-15
[Proposal] Chalice invoke command


kyleknap:
Overview
---------
With all of the new event types being released into Chalice, it would be great if Chalice had a built-in way to invoke the deployed Lambda function (and possibly a local version). Right now, I am invoking these functions through the console or the AWS CLI, but it would be much more convenient if Chalice had a built-in `invoke` command that accepts:

* `--name <function-name>`: The logical name of the Lambda function in the chalice app
* `--stage <stage-name>`: To specify the stage you want to target
* `--payload <value>`: A JSON payload to use when invoking the Lambda function

It would then print the result of the invoke to standard output.

Example
--------
Using this deployed `app.py`:
```python
from chalice import Chalice

app = Chalice(app_name='intro')


@app.lambda_function()
def hello(event, context):
    return {'hello': 'world'}


@app.lambda_function()
def hello_name(event, context):
    return {'hello': event['Name']}
```
Users would then be able to invoke the deployed Lambda functions manually:
```
$ chalice invoke --name hello
{"hello": "world"}
$ chalice invoke --name hello_name --payload '{"Name": "Kyle"}'
{"hello": "Kyle"}
```

Future additions
----------------

* Preformatted event payloads. Not sure how this would look quite yet, but essentially you can just provide the only the important event parameters to `chalice invoke` (i.e. bucket/key for S3 event or message for SNS event) and chalice will take the parameters and format an event on your behalf based on what the event source is and invoke the function.

* Local mode. Potentially add a `--local` flag so this will invoke your Lambda function in your Chalice application locally.

* `--context <context-value>`. Add the ability to send a context as well as a payload to invoke the Lambda function.

jamesls:
In general, I like the idea but there's a few edge cases I'm curious about:

* What's the expected behavior if I try this on a resource that's not an `@app.lambda_function()`?
* Payload is required is it not?  I think we can shorten it to just be a positional param.  I would also expect a common use case to load the payload from a file.  I'd almost think that's more common than specifying the JSON inline.
* We should support `-n/--name` for the name arg, consistency with `chalice logs -n`.

I'm also not sure if we want this as a top level command.  Thinking through the direction of improved local testing, it seems like putting this under the `local` command would make the most sense.  What about `chalice local invoke`?  We already have a local http server, this issue would add a "import then run your code" mode, and I could see in the future adding optional support for the docker lambda image assuming there's customer interest.  How would you see the commands for these new features looking like?

[...]

PR #901, opened 2018-07-17
Chalice invoke command

stealthycoin:
Implements #900 

### Sample app
```python
from chalice import Chalice

app = Chalice(app_name='whatever')


@app.route('/')
def index():
    return {'hello': 'world'}


@app.lambda_function()
def error(event, context):
    return context


@app.lambda_function()
def echo(event, context):
    return event
```

### Calling a lambda function
Providing a payload is done from stdin
```
$ echo '"hello world"' | chalice invoke -n echo
"hello world"
```

No payload
```
$ chalice invoke -n echo
{}
```

### Calling a route
You can call a route by using the `api_handler` function name and fake all the values that API Gateway would send. This is a little nosier:
```
$ cat apig.json
{
    "queryStringParameters": [],
    "headers": {},
    "pathParameters": {},
    "body": {},
    "stageVariables": {},
    "requestContext": {
        "resourcePath": "/",
        "httpMethod": "GET"
    }
}
$ cat apig.json | chalice invoke -n api_handler
{"headers": {}, "statusCode": 200, "body": "{\"hello\": \"world\"}"}
```

### Unhandled errors will show a stack trace
Errors will show a formatted stack trace to look like a local python function failed. This prevents the need to wait for cloudwatch logs to propagate.
```
$ chalice invoke -n error
  File "/var/runtime/awslambda/bootstrap.py", line 110, in decimal_serializer
    raise TypeError(repr(o) + " is not JSON serializable")
  File "/var/lang/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/var/lang/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/var/lang/lib/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
TypeError: <__main__.LambdaContext object at 0x7f3fa4d23f28> is not JSON serializable
```

### Handled errors are printed out
Errors thrown by the invoke command itself will be printed out in a pretty format. For example providing a non-json payload causes the `InvalidRequestContentException` to be thrown by lambda.
```
$ echo "invalid json" | chalice invoke -n echo
Error occured when invoking lambda: InvalidRequestContentException
Could not parse request body into json: Unrecognized token 'invalid': was expecting ('true', 'false' or 'null')
 at [Source: [B@30c3127a; line: 1, column: 9]
Aborted!
```

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -195,9 +195,10 @@ class TypedAWSClient(object):
         if payload is not None:
             kwargs['Payload'] = payload
 
-        return self._call_client_method_with_retries(
-            self._client('lambda').invoke, kwargs, max_attempts=1
-        )
+        try:
+            return self._client('lambda').invoke(**kwargs)
+        except RequestsReadTimeout as e:
+            raise ReadTimeout(str(e))
 
     def _is_iam_role_related_error(self, error):
         # type: (botocore.exceptions.ClientError) -> bool
@@ -991,8 +992,6 @@ class TypedAWSClient(object):
                         not should_retry(e):
                     raise
                 continue
-            except RequestsReadTimeout as e:
-                raise ReadTimeout(str(e))
             return response
 
     def _random_id(self):
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -186,8 +186,8 @@ class TypedAWSClient(object):
             return True
         return False
 
-    def invoke_function(self, name, context=None, payload=None):
-        # type: (str, _OPT_STR, bytes) -> Dict[str, Any]
+    def invoke_function(self, name, payload=None):
+        # type: (str, bytes) -> Dict[str, Any]
         kwargs = {
             'FunctionName': name,
             'InvocationType': 'RequestResponse',
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -27,6 +27,8 @@ import botocore.session  # noqa
 from botocore.exceptions import ClientError
 from botocore.vendored.requests import ConnectionError as \
     RequestsConnectionError
+from botocore.vendored.requests.exceptions import ReadTimeout as \
+    RequestsReadTimeout
 from typing import Any, Optional, Dict, Callable, List, Iterator  # noqa
 
 from chalice.constants import DEFAULT_STAGE_NAME
@@ -39,7 +41,6 @@ _OPT_INT = Optional[int]
 _OPT_STR_LIST = Optional[List[str]]
 _CLIENT_METHOD = Callable[..., Dict[str, Any]]
 
-
 _REMOTE_CALL_ERRORS = (
     botocore.exceptions.ClientError, RequestsConnectionError
 )
@@ -47,6 +48,12 @@ _REMOTE_CALL_ERRORS = (
 
 class AWSClientError(Exception):
     pass
+
+
+class ReadTimeout(AWSClientError):
+    def __init__(self, message):
+        # type: (str) -> None
+        self.message = message
 
 
 class ResourceDoesNotExistError(AWSClientError):
@@ -178,6 +185,19 @@ class TypedAWSClient(object):
         if re.search('event source mapping.*is in use', message):
             return True
         return False
+
+    def invoke_function(self, name, context=None, payload=None):
+        # type: (str, _OPT_STR, bytes) -> Dict[str, Any]
+        kwargs = {
+            'FunctionName': name,
+            'InvocationType': 'RequestResponse',
+        }
+        if payload is not None:
+            kwargs['Payload'] = payload
+
+        return self._call_client_method_with_retries(
+            self._client('lambda').invoke, kwargs, max_attempts=1
+        )
 
     def _is_iam_role_related_error(self, error):
         # type: (botocore.exceptions.ClientError) -> bool
@@ -971,6 +991,8 @@ class TypedAWSClient(object):
                         not should_retry(e):
                     raise
                 continue
+            except RequestsReadTimeout as e:
+                raise ReadTimeout(str(e))
             return response
 
     def _random_id(self):
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -19,7 +19,9 @@ from typing import Dict, Any, Optional  # noqa
 from chalice import __version__ as chalice_version
 from chalice.app import Chalice  # noqa
 from chalice.awsclient import TypedAWSClient
+from chalice.awsclient import ReadTimeout
 from chalice.cli.factory import CLIFactory
+from chalice.cli.factory import NoSuchFunctionError
 from chalice.config import Config  # noqa
 from chalice.logs import display_logs
 from chalice.utils import create_zip_file
@@ -30,6 +32,7 @@ from chalice.constants import DEFAULT_STAGE_NAME
 from chalice.constants import DEFAULT_APIGATEWAY_STAGE_NAME
 from chalice.local import LocalDevServer  # noqa
 from chalice.constants import DEFAULT_HANDLER_NAME
+from chalice.invoke import UnhandledLambdaError
 
 
 def create_new_project_skeleton(project_name, profile=None):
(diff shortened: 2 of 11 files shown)


chalice/cli/__init__.py
```diff
@@ -182,6 +189,59 @@ def deploy(ctx, autogen_policy, profile, api_gateway_stage, stage,
...
+                    'This is the logical name of the function. The only '
+                    'exception is to invoke one of the routes the name '
+                    'api_handler should be used.'))
+@click.option('--profile', metavar='PROFILE',
+              help='Override profile at deploy time.')
+@click.option('--stage', metavar='STAGE', default=DEFAULT_STAGE_NAME,
+              help=('Name of the Chalice stage to deploy to. '
+                    'Specifying a new chalice stage will create '
+                    'an entirely new set of AWS resources.'))
+@click.option('--connection-timeout',
```
kyleknap: We should just read this from the `.chalice/config.json` instead

chalice/cli/__init__.py
```diff
@@ -182,6 +189,59 @@ def deploy(ctx, autogen_policy, profile, api_gateway_stage, stage,
...
+@cli.command('invoke')
+@click.option('-n', '--name', metavar='NAME', required=True,
+              help=('The name of the function to invoke. '
+                    'This is the logical name of the function. The only '
+                    'exception is to invoke one of the routes the name '
+                    'api_handler should be used.'))
+@click.option('--profile', metavar='PROFILE',
+              help='Override profile at deploy time.')
+@click.option('--stage', metavar='STAGE', default=DEFAULT_STAGE_NAME,
+              help=('Name of the Chalice stage to deploy to. '
```
kyleknap: Probably should reword the `--stage` help command here. Looks like this is from the `deploy` command.

chalice/awsclient.py
```diff
@@ -179,6 +186,25 @@ def _is_settling_error(self, error):
...
+            'InvocationType': 'RequestResponse',
+        }
+        if context is not None:
+            kwargs['ClientContext'] = context
+        if payload is not None:
+            kwargs['Payload'] = payload
+
+        try:
+            return self._call_client_method_with_retries(
+                self._client('lambda').invoke, kwargs, max_attempts=1
```
kyleknap: Is there a specific reason we need `max_retries` on this?

chalice/awsclient.py
```diff
@@ -812,7 +838,7 @@ def _add_lambda_permission_if_needed(self, source_arn, function_arn,
         policy = self.get_function_policy(function_arn)
         if self._policy_gives_access(policy, source_arn, service_name):
             return
-        random_id = self._random_id()
+        random_id = generate_random_id()
```
kyleknap: Let's keep this an internal method unless some other module uses the utility function and find some other way to get the file under 1000 lines.

chalice/cli/__init__.py
```diff
@@ -182,6 +189,59 @@ def deploy(ctx, autogen_policy, profile, api_gateway_stage, stage,
     reporter.display_report(deployed_values)
 
 
+@cli.command('invoke')
+@click.option('-n', '--name', metavar='NAME', required=True,
+              help=('The name of the function to invoke. '
+                    'This is the logical name of the function. The only '
```
kyleknap: Maybe rephrase the second sentence to: If the function is decorated by `app.route`, use `api_handler` as the name instead

chalice/invoke.py
```diff
@@ -0,0 +1,88 @@
...
+    _TRACEBACK_HEADING = 'Traceback (most recent call last):\n'
+
+    def __init__(self, ui):
+        # type: (UI) -> None
+        self._ui = ui
+
+    def format_response(self, response):
+        # type: (Dict[str, Any]) -> None
+        payload = response[self._PAYLOAD_KEY].read()
+        if self._ERROR_KEY in response and \
```
kyleknap: Can be refactored to:

`if response.get(self._ERROR_KEY) == self._ERROR_VALUE`

chalice/awsclient.py
stealthycoin: Yes its a new required parameter of that method after I rebased it.

chalice/awsclient.py
```diff
@@ -13,6 +13,7 @@
 this class to get improved type checking across chalice.
 
 """
+# pylint: disable=too-many-lines
```
kyleknap: We can probably remove this line. This is under 1000 lines

tests/functional/cli/test_cli.py
```diff
@@ -354,3 +356,52 @@ def test_can_provide_lambda_name_for_logs(runner, mock_cli_factory):
...
+    invoke_handler = mock.Mock(spec=LambdaInvokeHandler)
+    mock_cli_factory.create_lambda_invoke_handler.return_value = invoke_handler
+    mock_cli_factory.create_payload_from_stdin.return_value = 'barbaz'
+    with runner.isolated_filesystem():
+        cli.create_new_project_skeleton('testproject')
+        os.chdir('testproject')
+        result = _run_cli_command(runner, cli.invoke, ['-n', 'foo'],
+                                  cli_factory=mock_cli_factory)
+        assert result.exit_code == 0
+    assert invoke_handler.invoke.call_args == mock.call('barbaz')
```
kyleknap: I wonder if it makes sense to assert the output here in the functional tests. Not sure how feasible it is, but it seems to worthwhile to look at more than just the return code for these tests.

kyleknap:
We should probably add it as a ``feature`` in the changelog as well.

chalice/awsclient.py
```diff
@@ -179,6 +186,19 @@ def _is_settling_error(self, error):
             return True
         return False
 
+    def invoke_function(self, name, context=None, payload=None):
```
jamesls: What's the `context` param used for?

chalice/awsclient.py
```diff
@@ -179,6 +186,19 @@ def _is_settling_error(self, error):
...
+        # type: (str, _OPT_STR, bytes) -> Dict[str, Any]
+        kwargs = {
+            'FunctionName': name,
+            'InvocationType': 'RequestResponse',
+        }
+        if payload is not None:
+            kwargs['Payload'] = payload
+
+        return self._call_client_method_with_retries(
+            self._client('lambda').invoke, kwargs, max_attempts=1
```
jamesls: I might have missed this in the review comments but what's the point of using `_call_client_method_with_retries` with a retry of 1?

chalice/cli/factory.py
```diff
@@ -157,6 +187,39 @@ def create_log_retriever(self, session, lambda_arn):
...
+        try:
+            resource = deployed.resource_values(name)
+            arn = resource['lambda_arn']
+        except (KeyError, ValueError):
+            raise NoSuchFunctionError(name)
+
+        function_scoped_config = config.scope(stage, name)
+        session = self.create_botocore_session(
+            read_timeout=function_scoped_config.lambda_timeout,
+            max_retries=0,
```
jamesls: Can you add a comment about why we set `max_retries=0`.  It may not be obvious to others why we want to disable retries when invoking the lambda function.

chalice/awsclient.py
stealthycoin: Missed removing it. The first POC had the ability to pass a custom context as well.

chalice/awsclient.py
stealthycoin: Is there an alternative? The method I was using didn't have that but got removed when I rebased. It looked like the expectation was to use that method. I don't really see a reason to re-write the old method when it was deleted in favor of this one.

jamesls: >Is there an alternative?

I still like I'm missing something, but I'm specifically asking what's the reason for not just using `self._client('lambda').invoke(**kwargs)`?

The call client method with retries is for retries, and if you have a `max_attempts` of 1 that's effectively a noop is it not?

> It looked like the expectation was to use that method. 

I touched on this in my previous comment:

> That method was added specifically because there's some settling time when making config changes (roles, event sources). We shouldn't need this here.


If there's not a good reason to use it, I'd prefer to just keep the code simpler.

PR #904, opened 2018-07-20
Content-Length header should be determined after any bytes encoding in local mode

leplastrier:
In local mode, if a handler returns a Python 3 string (unicode) object as the Response body, then the Content-Length will be incorrect if the string contains any code points > 256. The Content-Length was being determined before the unicode object was encoded into a bytes object, meaning it was counting glyphs and not bytes, which resulted in truncation in a client which was obeying Content-Length.


leplastrier:
It is unclear to me why codecov is reporting a failure for this PR. The commit includes both a unit and functional test which triggered the bug before being fixed and both of these tests exercise the code which was changed. Any advice on how to resolve the codecov failure would be appreciated.

leplastrier:
Fixed by #917

Issue #806, opened 2018-04-20
Latest version of chalice breaking pip (Python 3.6)


rajkunkolienkar:
After installing chalice from pip, it seems to have installed the enum34 library, which breaks my pip. 

`Traceback (most recent call last):`
  `File "/usr/local/bin/pip3.6", line 4, in <module>`
  `  import re`
  `File "/usr/lib/python3.6/re.py", line 142, in <module>`
  `  class RegexFlag(enum.IntFlag):`
`AttributeError: module 'enum' has no attribute 'IntFlag'`

Fixed it by removing the enum34 distribution from site-packages.

dstufft:
Are you able to provide steps to reproduce this? Generally Python ensures the stdlib comes first in ``sys.path`` and when I try to install both pip and enum34 into an environment locally, it all seems tow ork fine for me.

rajkunkolienkar:
I am running the latest EC2 AMI. 

I created a virtualenv with Python 3.6, pip installed chalice. 
chalice downgraded pip from v10.0.1 to v.9.0.3 and installed successfully.
When I run pip after installing chalice, I get the error as stated above.

dnorton:
Can we reopen this issue? I'm seeing the same problem.  It is because chalice is installing the module enum34.

```
Collecting argh>=0.24.1 (from watchdog==0.8.3->chalice->-r requirements.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/06/1c/e667a7126f0b84aaa1c56844337bf0ac12445d1beb9c8a6199a7314944bf/argh-0.26.2-py2.py3-none-any.whl
Installing collected packages: jmespath, six, python-dateutil, docutils, botocore, enum34, attrs, pathtools, PyYAML, argh, watchdog, typing, click, chalice, pbr, mock, PyMySQL, SQLAlchemy, s3transfer, boto3, dnspython, urllib3, certifi, idna, chardet, requests, aws-requests-auth
```

dnorton:
Can we reopen this issue? I am experiencing the same problem with when installing chalice.

enum34 will not work with python 3.6 - https://stackoverflow.com/questions/45896474/installing-enum34-on-python-3-x-breaks-pip-for-python3-x

DigiPutar:
I agree with @dnorton 
I am also facing the same issue while setting up the chalice environment over virtualenv on Amazon Linux machine AMI for Python 3.6

[...]

PR #905, opened 2018-07-20
switching dependency from enum34 to enum-compat

dnorton:
*Issue #, if available:* 

#806

*Description of changes:*

To prevent the enum34 library from breaking python installs 3.4+, I switched to enum-compat which has a very simple [setup.py](https://github.com/jstasiak/enum-compat/blob/master/setup.py):

```python
has_enum = sys.version_info >= (3, 4)
...
install_requires=[] if has_enum else ['enum34'],
```

I have tested this change on python3.6.5 and python 2.7.14 on Amazon Linux (Amazon Linux AMI release 2018.03)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ install_requires = [
     'six>=1.10.0,<2.0.0',
     'pip>=9,<11',
     'attrs==17.4.0',
-    'enum-compat',
+    'enum-compat>=0.0.2',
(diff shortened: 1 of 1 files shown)


jamesls:
Thanks for the pull request, I'll test this out.

jamesls:
Confirmed fixed for me.  I think in the future we could also look into using env markers to only use this package for 2.7 but I think we can go with this for now.

PR #910, opened 2018-07-24
Fixed message truncation bug

allfro:
The length of the utf-8 encoded body may be different than the unicode length of the string. This causes message truncation on the response. This patch fixes the issue.

*Description of changes:*
Moved the code that converts the body to bytes up a few lines to reside before the code that sends the `Content-Length` header


stealthycoin:
Looks good to me. I am going to add a test for this.

PR #920, opened 2018-07-30
Fix pluggy requirement

stealthycoin:
Pluggy released 0.7.0 and 0.7.1and the way pip does dependency
resolution it doesn't catch that pytest requires < 0.7. Hoisting the
requirement of <0.7 up to the top of requirements-dev.txt allows pip to
pull in the correct version of pluggy which is 0.6.0.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,5 +18,8 @@ pygments==2.1.3
 mock==2.0.0
 requests==2.18.4
 hypothesis==3.56.3
+# pip does not catch the <0.7 requirement in pytest so we need to add it to
+# the top level requirements.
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -18,5 +18,6 @@ pygments==2.1.3
 mock==2.0.0
 requests==2.18.4
 hypothesis==3.56.3
+pluggy<0.7
```
dstufft: I would add a comment about why this dependency is being added here, since it's not really a direct dependency and it's likely in N months we'll forget why it's here.

stealthycoin: Will do.

PR #934, opened 2018-08-15
Add debug logs to packager

stealthycoin:
Add debug logs to packager

Added log statements to track the state of all packages in between each
step during the packaging process. A function in the CLI layer was added
to enable this logger when the --debug flag is passed.

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -26,7 +26,6 @@ from chalice.config import Config  # noqa
 from chalice.logs import display_logs
 from chalice.utils import create_zip_file
 from chalice.deploy.validate import validate_routes, validate_python_version
-from chalice.deploy.packager import LOGGER_NAME as PACKAGER_LOGGER_NAME
 from chalice.utils import getting_started_prompt, UI, serialize_to_json
 from chalice.constants import CONFIG_VERSION, TEMPLATE_APP, GITIGNORE
 from chalice.constants import DEFAULT_STAGE_NAME
@@ -36,11 +35,11 @@ from chalice.constants import DEFAULT_HANDLER_NAME
 from chalice.invoke import UnhandledLambdaError
 
 
-def _configure_logging(name, level, format_string=None):
-    # type: (str, int, Optional[str]) -> None
+def _configure_logging(level, format_string=None):
+    # type: (int, Optional[str]) -> None
     if format_string is None:
         format_string = "%(asctime)s %(name)s [%(levelname)s] %(message)s"
-    logger = logging.getLogger(name)
+    logger = logging.getLogger('')
     logger.setLevel(level)
     handler = logging.StreamHandler()
     handler.setLevel(level)
@@ -100,6 +99,8 @@ def cli(ctx, project_dir, debug=False):
     # type: (click.Context, str, bool) -> None
     if project_dir is None:
         project_dir = os.getcwd()
+    if debug is True:
+        _configure_logging(logging.DEBUG)
     ctx.obj['project_dir'] = project_dir
     ctx.obj['debug'] = debug
     ctx.obj['factory'] = CLIFactory(project_dir, debug, environ=os.environ)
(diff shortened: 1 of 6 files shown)


chalice/deploy/packager.py
```diff
@@ -27,6 +28,9 @@
 OptStr = Optional[str]
 OptBytes = Optional[bytes]
 
+LOGGER_NAME = 'packager'
+logger = logging.getLogger(LOGGER_NAME)
```
kyleknap: You can just use `__name__` here instead of `LOGGER_NAME`. This allows the file to get moved around without having to update the string.

tests/integration/test_cli.py
```diff
@@ -36,3 +36,16 @@ def test_stack_trace_printed_on_error(local_app):
     assert rc == 2
     assert 'Traceback' in stderr
     assert 'foobarbaz' in stderr
+
+
+def test_debug_flag_enables_logging(local_app):
+    req_file = os.path.join(local_app, 'requirements.txt')
+    with open(req_file, 'w') as f:
+        f.write('pip==10.0.0\n')
+    p = subprocess.Popen(['chalice', '--debug', 'package', 'out'],
```
kyleknap: So the coverage is down because it is being ran in a subprocess?

chalice/deploy/packager.py
```diff
@@ -405,6 +415,7 @@ def _download_dependencies(self, directory, requirements_filename):
         # compiler.
         compatible_wheels, incompatible_wheels = self._categorize_wheel_files(
             directory)
+        logger.debug("compatible: %s", compatible_wheels)
```
kyleknap: It looks like there are a few `compatible` repeats. It may be good to update it so the statements are unique so it is not ambiguous where the debug line is coming from and worded in a way that I do not have to dive into the code to see at what step the compatible log entry is getting written.

chalice/cli/__init__.py
```diff
@@ -363,6 +380,9 @@ def generate_sdk(ctx, sdk_type, stage, outdir):
 def package(ctx, single_file, stage, out):
     # type: (click.Context, bool, str, str) -> None
     factory = ctx.obj['factory']  # type: CLIFactory
+    debug = ctx.obj['debug']  # type: bool
```
kyleknap: I wonder if it makes sense to instantiate this at the `cli` level so it gets essentially set for all chalice commands? While it is not bad for two commands, I think we should try to avoid having to do this every time to a new command that needs logs displayed when running `--debug`. We also set it when we create the botocore session. So there is definitely some potential refactoring. Also what was the reasoning behind not setting the logger name to an empty string?

chalice/deploy/packager.py
stealthycoin: Yea so I think this is unavoidable. The logs are only useful if I can see the state of the world between each step.

chalice/cli/__init__.py
stealthycoin: This was more of a POC to facilitate this discussion. Not sure how we want to do logging, nor am I sure how it should be done with click or if there is a better pattern available.

chalice/deploy/packager.py
jamesls: I think Kyle's just asking to add something besides "compatible" so that similar lines (such as line 404) can be distinguished without having to dig through the source code.  Maybe something like `compatible wheels after second download pass`, `compatible wheels after categorization`, etc.

chalice/deploy/packager.py
jamesls: Yeah I'm fine with just using `__name__` and having this be part of our standard debug logs.

chalice/cli/__init__.py
jamesls: I think at least for now, we should have a single place that configures logging.  We were previously relying on the `set_debug_logger()` in the botocore session previously, but now that we're adding our own logging in chalice, we should just do the configuration ourselves and not rely on botocore's configuration.

It might be simplest to just call `_configure_logging()` in `cli()` and remove the call to `set_debug_logger()` in the `CLIFactory`.

tests/integration/test_cli.py
stealthycoin: Yes unless i'm missing something I cant see how to even test top level args with the click test runner.

jamesls: Does just passing the `cli` function to the click runner not work?

tests/functional/cli/test_factory.py
```diff
@@ -41,28 +41,12 @@ def assert_has_no_request_body_filter(log_name):
...
-        isinstance(f, factory.LargeRequestBodyFilter) for f in log.filters)
-
-
 def test_can_create_botocore_session():
     session = factory.create_botocore_session()
     assert session.user_agent().startswith('aws-chalice/')
     assert session.get_default_client_config() is None
 
 
-def test_can_create_botocore_session_debug():
```
jamesls: Not sure why this was removed.  Can you add this back (unless you know why it needs to be removed)?

PR #937, opened 2018-08-20
Fix issue with downloading wheels when using python version other than 27 or 36

jtetrault:
Explicitly pass --python 36 (or python 27) to pip download subprocess cmd.
Fixes an issue where no wheels are found if the current python
interpreter is not 27 or 36.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -80,8 +80,9 @@ class LambdaDeploymentPackager(object):
         requirements_filepath = self._get_requirements_filename(project_dir)
         with self._osutils.tempdir() as site_packages_dir:
             try:
+                python_version_numerical = ''.join((s for s in python_version if s.isdigit()))
                 self._dependency_builder.build_site_packages(
-                    requirements_filepath, site_packages_dir)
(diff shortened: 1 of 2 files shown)


stealthycoin:
Hmm I am not sure this is the correct behavior. In any event the CI fails so we can't take the PR. Can you explain the situation you are having where it doesn't work? I would need to experiment with it to see the actual behavior and try a few edge cases. But in general, the way Chalice works you should have a virtualenv that matches the remote python version you intend to target, so there shouldn't be any need for this.

stealthycoin:
@jtetrault the main issue here it isn't separate from your virtualenv. In fact it uses that to determine which version to deploy. It is intended for you to setup a virtual environment that matches the lambda environment. Your CI/CD should _definitely_ be in the same version.

This is documented in this section: https://chalice.readthedocs.io/en/latest/topics/pyversion.html which also shows an example of using a virtual env to control the deployed version.

jamesls:
Closing due to inactivity and based on @stealthycoin last comment.

PR #947, opened 2018-09-18
added wheel dependency

### LINKED ISSUE #928: Packager assumes wheel to be installed
Took me a while to figure out why Chalice fails to build my deployment package properly in one environment, but not the other. Funnily enough while I was trying to pin all transitive dependencies :(

Chalice appears to use [wheel](https://github.com/aws/chalice/blob/9fe9ddc0423acd202c7e3533b74d14691336d8f7/chalice/deploy/packager.py#L693) regardless of whether wheel is installed in the environment. Should wheel be a setup requirement? Or perhaps include a warning in the "Could not install dependencies..." message?

```
$ python -m venv v1
$ source v1/bin/activate
$ pip install chalice  # wheel not installed
$ chalice new-project v1-test
$ cd v1-test
$ echo "markupsafe" >> requirements.txt
$ chalice package out
Creating deployment package.

Could not install dependencies:
markupsafe==1.0
You will have to build these yourself and vendor them in
the chalice vendor folder.

Your deployment will continue but may not work correctly
if missing dependencies are not present. For more information:
http://chalice.readthedocs.io/en/latest/topics/packaging.html

$ pip install wheel

$ chalice package out       
Creating deployment package.
# Success!
```

david-poirier-csn:
Resolves #928 

Adds `wheel` as a dependency

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ install_requires = [
     'attrs==17.4.0',
     'enum-compat>=0.0.2',
     'jmespath>=0.9.3,<1.0.0',
+    'wheel>=0.31.1'
 ]
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,8 @@ install_requires = [
     'attrs==17.4.0',
     'enum-compat>=0.0.2',
     'jmespath>=0.9.3,<1.0.0',
-    'wheel>=0.31.1'
+    'wheel',
+    'setuptools'
 ]
 
 setup(


stealthycoin:
@david-poirier-csn Thanks for the PR. Looking at the broader issue can we also add setuptools as a requirement as a part of this PR. It's less likely its missing, but that's the same reason wheel wasn't included, its just very unlikely its not in the environment.

I'd also like to remove the range constraints. Since this is a very pivotal package, same with setuptools, and we don't want to be changing it on our users for no reason. Pip itself doesn't care and just assumes that both are in the env, so I think we can follow suit and just ensure its in the environment.

david-poirier-csn:
Roger that, will do

PR #951, opened 2018-09-25
Update CORSConfig example to be more complete

### LINKED ISSUE #950: Documentation on CORS for multiple origins
According to the documentation at https://github.com/aws/chalice#tutorial-cors-support, 

"If you need to supply multiple origins you will need to define a custom handler for it that accepts OPTIONS requests and matches the Origin header against a whitelist of origins. " 

then there is an example that does not accept OPTIONS. Part of example:

```
@app.route('/cors_multiple_origins', methods=['GET'])
def supports_cors_multiple_origins():

```
 Is the example wrong or the rest of the documentation wrong?
Algo, the route wrapper has no cors parameter there, and the documentation states that "Every view function must explicitly enable CORS support."

I think both issues should be addressed in the documentation.

stealthycoin:
Fixes #950

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -854,7 +854,7 @@ There's a couple of things to keep in mind when enabling cors for a view:
   in the ``Access-Control-Allow-Origin`` header.
 
   Example:
-  
+
 .. code-block:: python
 
     from chalice import Chalice, Response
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -854,30 +854,37 @@ There's a couple of things to keep in mind when enabling cors for a view:
...
+	    }
+	    origin = app.current_request.headers.get('origin', '')
+	    if origin in _ALLOWED_ORIGINS:
+		headers.update({'Access-Control-Allow-Origin': origin})
+	    return Response(
+		body=None,
+		headers=headers,
+	    )
+	elif method == 'GET':
+	    ... respond with requested resource...
```
kyleknap: I think we should make this entire example at least does not have syntax errors. Most people are just going to copy and paste this. So I could see it being annoying if it does not work. We could just return something simple here.

stealthycoin: Ok sure.

Issue #964, opened 2018-10-12
Query parameter parsing in local dev server


nickjbyrne0:
I've been attempting to integrate python-filterparams into my api which is implemented in a Chalice app, however I think i've found an issue in that RouteMatcher function match_route which reduces the query_params via parse_qs and a dictionary comprehension and drops all but one of the parameters which have the same name. For example

a query string such as 

`/myapi?filter=%5Bparam%5D%5Bname%5D%5Blike%5D%3Dtest%25&filter=%5Bparam%5D%5Bage%5D%5Beq%5D%3D30`

which httpie would be 

`http -v :8000/myapi filter==[param][name][like]=test% filter==[param][age][eq]=30`

The app.current_request.query_params is a dict containing 

`{'filter': '[param][name][like]=test%'}`

Expected output would be

`{'filter': '[param][name][like]=test%', '[param][age][eq]=30'}`

With a little digging around i can see in 

https://github.com/aws/chalice/blob/master/chalice/local.py#L128

the dict comprehension only takes the first value from the list, which is the reason for losing additional parameters. However, not being that familiar with Chalice's development i can't comment on why this decision was taken or if its actually needed, could we just use the result of parse_qs instead?

stealthycoin:
So this looks like it was contributed code so I don't think we really have a reason why it is this way. 

To back up there there is no spec for multiple keys in query (https://tools.ietf.org/html/rfc3986#section-3.4). Its undefined behavior so anything we do is technically correct. You shouldn't be expecting that output you posted above, it totally depends on the web framework/stack what happens to it. 

In my testing it looks like API Gateway goes with last-given rather than first-given which is what we are doing in local mode so we should switch this behavior to match them since that is the whole point of local mode. So I will be marking this as an enhancement for local mode.

PR #965, opened 2018-10-12
Make local mode querystring processing consistent with APIGateway

stealthycoin:
closes: #964


stealthycoin:
Just to show it matches API Gateway:
```python
from chalice import Chalice

app = Chalice(app_name='test-query')


@app.route('/')
def index():
    return 'args: %s' % app.current_request.query_params
```

```
$ curl "localhost:8000/?a=1&a=2"
args: {'a': '2'}
```
```
$ curl $(chalice url)"?a=1&a=2"
args: {'a': '2'}
```

PR #986, opened 2018-11-15
Add support for bytes when binary_types accept application/json

danilobrinu:
```
import json
import gzip

app = Chalice(app_name='gzipcontent')
app.api.binary_types.append('application/json')

@app.route('/')
def index():
    blob = json.dumps({'hello': 'world'}).encode('utf-8')
    payload = gzip.compress(blob)
    return Response(body=payload, headers={'Content-Type': 'application/json', 'Content-Encoding': 'gzip'})
```

Issue #988, when binary_types is **application/json** and you pass bytes in the body the code throws error

*Description of changes:*
This patch check if the body is a string and apply .encode('utf-8') if not let me the possibility to pass bytes from gzip.compress. Before this patch the before code can't work

Note: the example are in the top

### DIFF
--- a/chalice/deploy/swagger.py
+++ b/chalice/deploy/swagger.py
@@ -1,4 +1,5 @@
 import copy
+import inspect
 
 from typing import Any, List, Dict, Optional  # noqa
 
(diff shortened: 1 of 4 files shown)


joguSD:
I've continued this PR by squashing and rebasing against the latest master in #1013

danilobrinu:
Thank you @joguSD

PR #991, opened 2018-11-22
Remove upper bound on pip version

### LINKED ISSUE #981: Incompatible with Pip>18.0
Chalice 1.6.1 is not compatible with latest `pip` version 18.1
Also it depends on previous version of `click` package i.e. 6.6 rather than 7.0.
It becomes difficult to maintain chalice as a part of package environment.

stealthycoin:
Since pip no longer uses semver the upper bound isn't particularly
useful. It also prevents Chalice from using newer versions once pip is
upgraded. Chalice will also downgrade the environment's pip version when
installed which can be a source of annoyance.

fixes #981

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ install_requires = [
     'botocore>=1.10.48,<2.0.0',
     'typing==3.6.4',
     'six>=1.10.0,<2.0.0',
-    'pip>=9,<19.0',
+    'pip>=9,<=18.1',
     'attrs==17.4.0',
     'enum-compat>=0.0.2',
     'jmespath>=0.9.3,<1.0.0',
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ install_requires = [
     'botocore>=1.10.48,<2.0.0',
     'typing==3.6.4',
     'six>=1.10.0,<2.0.0',
-    'pip>=9,<=18',
+    'pip>=9',
     'attrs==17.4.0',
     'enum-compat>=0.0.2',
     'jmespath>=0.9.3,<1.0.0',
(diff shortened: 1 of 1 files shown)


jamesls:
I'm hesitant to have no upper bound on this, especially given we rely on internal pip APIs.  Can we just lock it to a year?

jamesls:
@stealthycoin I pushed a change to lock to <19.  My concern is mostly about future changes to pip that won't be compatible with chalice.  If a user has a working app on a certain version of chalice, then  it's possible that chalice may not deploy correctly in the future because there's no upper bound on pip.

dstufft:
Fundamentally there's nothing different from the 18.X -> 19.0 release than there is for the 18.X -> 18.X+1 release. It seems wrong to me to lock it to a year for the stated reason, because locking it to a year doesn't actually solve the problem of that a release of pip can break chalice. We should probably either keep the ceiling as is, or remove it.

jamesls:
Quick update, I changed the dependency to <=18.1.  This is essentially what we've been doing so far, which is to lock to a version of pip (or less) that we know works with chalice.

Yes, we'll have to keep this updated with new pip releases, but I think the longer term plan should to investigate vendoring this code into chalice.

dmulter:
This is a problem again now that pip 19.0.2 is out.

duaneking:
This defect is exactly why this project seems so dead; You guys are testing the wrong things and this arbitrary limit is overly restrictive and creates more work every time PIP is updated, as well as greatly harms usage of the library because we as developers CANT TRUST IT when we try to install it the first time and find it actively conflicts with our other pip and package installs.

Pip installs should never break packages; there is a very valid reason for venvs and that is why you should use them... but best practice in python is always to use the newest version of pip so there is no excuse for this.

A version of PIP getting upgraded shouldn't even have the possibility of breaking your modules if they are coded correctly; What are you doing that could break based on pip?  It seems to me that is the problem here, remove the sickness of having an upper limit on pip and this sickness of not being able to use chalice for a month or two every time pip updates goes away.

PR #992, opened 2018-11-22
Add support for python 3.7

stealthycoin:
Lambda now supports python3.7. This commit adds support for packaging
your Chalice app, and deploying it to python 3.7.

Note: To support python 3.7 pylint needed to be updated to 2.0.0. This adds a lot of new validations that we fail. So this PR needs another commit that goes through and disables them or fixes our code to pass them. I don't have the energy to play whack a mole with that right now so pushing this up to get thoughts on the relevant code changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: python
 sudo: false
 matrix:
   include:
+  - python: "3.7"
+    dist: xenial
+    sudo: true
+    env: TEST_TYPE=prcheck HYPOTHESIS_PROFILE=ci
   - python: "3.6"
     env: TEST_TYPE=prcheck HYPOTHESIS_PROFILE=ci
   - python: "2.7"
--- a/.pylintrc
+++ b/.pylintrc
@@ -59,7 +59,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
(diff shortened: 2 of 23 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ language: python
 sudo: false
 matrix:
   include:
+  - python: "3.7"
+    dist: xenial
```
jamesls: Why's this needed?

stealthycoin: Last time I checked 3.7 was only available on xenial on travis. That may no longer be the case but its what we do in our other projects as well:
https://github.com/aws/aws-cli/blob/develop/.travis.yml#L23
https://github.com/boto/botocore/blob/develop/.travis.yml#L23
https://github.com/boto/boto3/blob/develop/.travis.yml#L23

jamesls:
Ok I've got everything working *except* figuring out what we want to do with pylint.

I've updated this branch to work with pylint 2.x and on python3 it passes.  However, pylint 2.x doesn't run on python2.  To further complicate things, on py2 we have to use a specific commit of pylint otherwise we'll get lint errors about relative imports.  And one last final complications, we can't use env markers with `-e` (which we use to point to a specific commit).

So I think perhaps our best option is to just silence the relative imports on py2 if we use an officially released 1.x version of pylint.  Otherwise I can see if I can get the code passing without any additional ignore rules.

jamesls:
Yeah so I dont' know what codecov is complaining about.  If you click on it's link and then click on the red areas it doesn't actually show me what's missing.  AFAICT there's no drop in test coverage.

Issue #997, opened 2018-11-30
FileNotFound not properly handled in filewatcher


santtu:
The filewatcher will get an exception in a situation where a symlink to non-existent file exists in the directory it is watching. This will occur because `mtime` method in `utils.py` uses `os.stat` which will fail with `FileNotFound` exception if a symlink that does not point to a real file is detected.

This is pretty easily occurring situation, for example Emacs creates an `.#app.py` file symlinked to a non-existing file in the directory when editing. This is what then happens:

```
$ chalice local
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/Users/santtu/.pyenv/versions/3.6.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Users/santtu/.pyenv/versions/3.6.6/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/santtu/.pyenv/versions/3.6.6/envs/adc-reborn/lib/python3.6/site-packages/chalice/cli/filewatch/stat.py", line 49, in poll_for_changes_until_shutdown
    self._seed_mtime_cache(root_dir)
  File "/Users/santtu/.pyenv/versions/3.6.6/envs/adc-reborn/lib/python3.6/site-packages/chalice/cli/filewatch/stat.py", line 59, in _seed_mtime_cache
    self._mtime_cache[path] = self._osutils.mtime(path)
  File "/Users/santtu/.pyenv/versions/3.6.6/envs/adc-reborn/lib/python3.6/site-packages/chalice/utils.py", line 249, in mtime
    return os.stat(path).st_mtime
FileNotFoundError: [Errno 2] No such file or directory: '/Users/santtu/dev/adc-reborn/adc-lambda/.#app.py'

Serving on 127.0.0.1:8000
```

The file watcher (reloader) will not work from this point on, e.g. any changes to application files won't be reloaded.

You don't need Emacs for this, just run `ln -s foo bar` in the app directory before or during `chalice local`.

This situation is not covered by any test case as the `OSUtils` class itself is not subject to any test.

See this change which will cause `test_stat.py` test to fail: https://github.com/santtu/chalice/commit/ade2a5d6ed9b7af30980caabfa09c1ed5b66a635

(I do not understand the rationale on the `initial_scan` in `FakeOSUtils` --- it seems to assume something on the inner workings of `StatFileWatcher`.)

kyleknap:
[...]

PR #998, opened 2018-11-30
Fix to prevent FileNotFoundErrors from stopping filewatcher  

santtu:
See issue #997.

* Changed the test so it exposes a bug in `StatFileWatcher._seed_mtime_cache` wrt/ `FileNotFoundError`
* Added exception handler to `_seed_mtime_cache` to handle the case (ignore it)

Someone more familiar with these tests should take a look at `test_stat.py`, the use of `initial_scan` in the mock osutils class smells like it is making assumptions about the specific implementation of `StatFileWatcher`.

### DIFF
--- a/tests/unit/cli/filewatch/test_stat.py
+++ b/tests/unit/cli/filewatch/test_stat.py
@@ -9,7 +9,7 @@ class FakeOSUtils(object):
         self.initial_scan = True
 
     def walk(self, rootdir):
-        yield 'rootdir', [], ['bad-file', 'baz']
+        yield 'rootdir', [], ['bad-file', 'no-file', 'baz']
         if self.initial_scan:
             self.initial_scan = False
 
@@ -17,6 +17,8 @@ class FakeOSUtils(object):
         return os.path.join(*parts)
 
     def mtime(self, path):
+        if path.endswith('no-file'):
+            raise FileNotFoundError()
         if self.initial_scan:
             return 1
         if path.endswith('bad-file'):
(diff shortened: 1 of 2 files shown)


santtu:
Okay, I can see that the test now fails on Python 2.

```
/tmp aglaja $ python2 -c 'import os; os.stat("bar")'
  File "<string>", line 1, in <module>
OSError: [Errno 2] No such file or directory: 'bar'
/tmp aglaja $ python3 -c 'import os; os.stat("bar")'
  File "<string>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'bar'
```

In general, the test could just remove the `initial_scan` and fail on the `"baz"` file with `OSError`, and that would cover both Python 2 and Python 3.

santtu:
There is actually a problem that the fix exposed. `_is_changed_file` handles OSError, but it handled the symlink-to-nowhere as a new file every time it was called, causing constant reloads. Fixed in https://github.com/santtu/chalice/commit/46cb9cefe0e5de564c6bd9f210c18e0ab5cd4eaa, OTOH, the this probably warrants a more thorough test case (thinking about it).

tests/unit/cli/filewatch/test_stat.py
```diff
@@ -1,40 +1,113 @@
...
-        time.sleep(0.2)
-    else:
-        raise AssertionError("Expected callback to be invoked but was not.")
+
+    os.wait()
+    assert len(os.calls) == 0
+    assert len(calls) == 0
+
+    for file_mtimes, os_calls, cb_calls in detect_changes_cases:
+        print("test case:", "file_mtimes", file_mtimes,
```
kyleknap: If we use parameterize, it will help us avoid needing a print statement determine the case that is getting ran

PR #1010, opened 2018-12-05
Loading the environment variables before deploying

tcatrain:
*Description of changes:*
I added this line to the `chalice/cli/factory.py` file in order to load the environment variables before deploying an chalice project :

```python
def create_config_obj(self, chalice_stage_name=DEFAULT_STAGE_NAME,
        ...
        os.environ.update(config.environment_variables)
```

By loading these variables, we can now use variables outside of a lambda scope. I needed this using domovoi, in order to use clean variables in cloudwatch event patterns.


kapilt:
this seems useful for static environment variables

Issue #759, opened 2018-03-21
app.current_request is not thread safe


edwinevans:
I'm running into a problem when testing using chalice local and making calls to my service at almost the same time. It seems like app.current_request could get overwritten by a later request by the time I finish handling an earlier one so my current handler is working with a request object with context for the later request (i.e. I'm handling Request 1, Request 2 comes in, Request 3 comes in, my handler for request 2 uses app.current_request and gets the context for request 3). 

Would it be better to pass the request object in to the route?

Is this a potential problem when running in lambda or does each handler get a separate process?

stealthycoin:
`app.current_request` makes the assumption that only 1 is going to be executed at a time since that is how it functions in Lambda and makes everything simpler. We could play with switching to ForkingMixIn to separate them by a process boundary.

edwinevans:
Thanks for your response. Does Lambda with api gateway only execute one request at a time? I have seen something that looks similar now when running in AWS. In one case I got a Chalice error of "invalid_request" with description "invalid concurrent request". What does this mean?

edwinevans:
Is there any way to workaround/force the requests to come in sequentially? I don't see how I can make something reliable if I can't assume that the request body corresponds to the request function and that request bodies won't get lost.

edwinevans:
Regarding "invalid concurrent request" I can't find the complete response any more. I'll let you know if I see it again.

Regarding the originally described issue it's possible to reproduce it by having the browser make multiple requests in parallel using Promise.All. (I'm not doing this any more so the issue will happen less frequently for me but may be worse when it does happen since I won't be expecting it.)

Thanks for looking into this.

clarsen:
> Only one can execute in Lambda at a time, so this is not an issue with production Chalice, it is an issue with local mode properly emulating the production service. The local mode uses threads since browsers can hog the connection and was causing local mode to hang after a request from a browser. 

Hi there, I noticed this issue in my local development too, but attached a PR with a simple fix that has been working well for me over the last few weeks in local mode operation.

PR #1011, opened 2018-12-05
Don’t multi-thread local dev server, but also force connection close to prevent hanging

clarsen:
Prevents request handlers from running concurrently resulting in indeterminate app.current_request during a request.

*Issue #, if available:*
https://github.com/aws/chalice/issues/759

*Description of changes:*
Use HTTPServer instead of ThreadedHTTPServer, but also instruct HTTP client to close keep-alive connection by adding a "Connection: Close" header.

Tested with Chrome web browser and iOS client using actual apps hitting a production HTTP API.

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -554,6 +554,7 @@ class ChaliceRequestHandler(BaseHTTPRequestHandler):
             )
             status_code = response['statusCode']
             headers = response['headers']
+            headers['Connection'] = 'Close'
             body = response['body']
             self._send_http_response(status_code, headers, body)
         except LocalGatewayException as e:
@@ -615,7 +616,7 @@ class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
 class LocalDevServer(object):
     def __init__(self, app_object, config, host, port,
                  handler_cls=ChaliceRequestHandler,
-                 server_cls=ThreadedHTTPServer):
+                 server_cls=HTTPServer):
         # type: (Chalice, Config, str, int, HandlerCls, ServerCls) -> None
         self.app_object = app_object
         self.host = host


clarsen:
It does look like the travis tests aren't passing due to some assumptions about the local server interface -- will fix that shortly.

jamesls:
Hi, thanks for the pull request.  We originally had the code as you have it now, but we changed it to the threaded server in https://github.com/aws/chalice/pull/659.  While some browsers work without the multithreaded server, other browsers still have problems.

I think to address #759 we should either:

a) Only invoke one request at a time (put a lock around the request handler in local mode)
b) Switch to a Fork based server.

Let me know if you'd be interested in pursuing either of those changes.

clarsen:
> Hi, thanks for the pull request. We originally had the code as you have it now, but we changed it to the threaded server in #659. While some browsers work without the multithreaded server, other browsers still have problems.
> 
> I think to address #759 we should either:
> 
> a) Only invoke one request at a time (put a lock around the request handler in local mode)
> b) Switch to a Fork based server.
> 
> Let me know if you'd be interested in pursuing either of those changes.

Hi, yeah, i did briefly look at wrapping the request handling with a lock embedded in a decorator (at the moment of invoking the handler) -- that didn't help of course since the current_request state itself was set up/reset before invoking the handler.

I could see about elevating that option "a" to happen within chalice itself.

Regarding "b", https://github.com/aws/chalice/pull/659#issuecomment-358091606 hints that other libraries that talk to services off the box, e.g. via a persistent socket, might not deal well with having their operations run concurrently even if done among multiple processes.

NomanGul:
any update on this PR? 👀

PR #1029, opened 2019-01-02
yaml-based configuration

2ps:
We've updated all of the json configuration files to be yaml-based instead as yaml is more human readable and manageable and, you know, aws supports it.  unit and functional tests are confirmed passing.  We're also happy to break this up into smaller PRs as needed.


chalice/cli/__init__.py
```diff
@@ -52,7 +52,7 @@ def create_new_project_skeleton(project_name, profile=None):
     # type: (str, Optional[str]) -> None
     chalice_dir = os.path.join(project_name, '.chalice')
     os.makedirs(chalice_dir)
-    config = os.path.join(project_name, '.chalice', 'config.json')
+    config = os.path.join(project_name, '.chalice', 'config.yml')
```
tedivm: Since YAML is a superset of JSON you can make this PR backwards compatible by doing a "does file exist" check for both the `.json` and the `.yml` file and then using the one that exists.

2ps: Great idea.

2ps:
@tedivm :  bump--anything else you;d like on this?  I'd like to use this as the base part of adding layer-building functionality into chalice.

v-weh:
Hi, is there any reason why this change is never merged into the repo? I think from a readibility point of view, having YAML config and `IAM_POLICY_FILE` will be of tremendous help ...

tedivm:
I think it wasn't merged, or responded to, because this project just isn't a priority at all for amazon. I honestly forgot it even existed until you responded.

Issue #1030, opened 2019-01-03
contributing:  allow use of `mock.patch`


2ps:
Unit tests / functional tests should allow for the use of `mock.patch`.  This is a natural pattern especially for mocking calls to `with open(filename, 'r') as f` and would make some of the unit / functional tests cleaner than  constantly mocking out `OSUtils.get_file_contents`.  Using `get_file_contents` makes double work of reading a file (which will be buffered natively) and then reading into another buffer BytesIO or StringIO before being able to use the read file.  Obviously, I don't know the original reasoning behind not allowing use of `mock.patch`, but I'd be happy to learn the reasoning behind it!

kyleknap:
Disallowing patching is just a coding preference that we would prefer to keep in the chalice codebase. In terms of the main reasons we disallow patching, I would say:

* Patching can be tricky and tedious to maintain. It is quite easy to forget to stop a patch and have the leaky patch bite you months later down the road. In other projects that we maintain that we allow patching, I have spent a significant amount of time tracking down patches that were not cleaned up in order to get a test that I wrote passing (even though my test properly cleaned up the patch).  
* Patching can lead to tests that are not very through. I have seen tests before that apply 4+ patches and once all of the patches have been applied the test is not doing very much as most of the objects being interacted with are mocks, which can cover up runtime issues.
* Patching can cover up code smells in terms of defining the correct level of abstraction for a class and determining its scope in functionality. By disallowing patching, it encourages people to better break up their abstractions and rely more on dependency injection, which helps in both testing and reusability of their abstractions.

There are probably more reasons I could list, but those are just some to start. I hope this clears things up.

2ps:
Understood and thank you for the reply.  In this case, I really wanted to patch around the use of open (and that was it) with a `with`.  I take your point about it easily getting out of hand, but isn't that what code review PRs are for -- so that we can catch those issues before they get out of hand?

stealthycoin:
Link to a talk given by @kyleknap at pycon about our opinion on automated code quality checks and how time should be spent reviewing. 

https://www.youtube.com/watch?v=G1lDk_WKXvY

PR #1031, opened 2019-01-03
[ISSUE-1030]  Removed use of the mock patch linter.

2ps:
mock patching lends itself to natural patterns like using
`with mock.path` to patch calls to `open`.  The alternative,
passing around mocked up osutils classes makes for a bad
production pattern since it requires an extra level of buffering.

*Issue #1030*

*Description of changes:*

Removed linter class that prevented use of `mock.patch`

### DIFF
--- a/tests/linter.py
+++ b/tests/linter.py
@@ -4,29 +4,7 @@ from astroid.exceptions import InferenceError
 
 
 def register(linter):
-    linter.register_checker(PatchChecker(linter))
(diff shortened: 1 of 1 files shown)


kyleknap:
Hi @2ps. Thanks for the pull request. I put a reply to the GitHub issue you opened up about it: https://github.com/aws/chalice/issues/1030. Let's keep the conversation over there. But at this point, I do not think we want to remove this check to disallow patching.

Issue #940, opened 2018-08-25
Using --project-dir option


hx2A:
When deploying a package this works fine to deploy a package:

```
$ cd quickdraw-lambdas/
$ chalice deploy --no-autogen-policy
```

but this does not:

```
$ cd ..
$ chalice --project-dir "quickdraw-lambdas/" deploy --no-autogen-policy
```

I get the error message:

`FileNotFoundError: [Errno 2] No such file or directory: 'quickdraw-lambdas/.chalice/config.json'`

Even though that file exists:

```
$ ll quickdraw-lambdas/.chalice/config.json
-rw-rw-r-- 1 jim jim 129 Aug 18 21:07 quickdraw-lambdas/.chalice/config.json
```

I can change that directory to an absolute path to make it work but I want it to be relative. I do see issue #414 . Is it supposed to work this way? Is there a way to specify a relative path?

Thanks

kyleknap:
It looks like as of now it does not support relative paths. However, I am inclined to support relative paths. It is tedious to always specify absolute paths. @jamesls @stealthycoin any thoughts?

It looks like the issue you linked aims to fix it. So I would recommend following that PR.

PR #1037, opened 2019-01-07
now project_dir always will be stored as absolute path in the context.

igorborgest:
Requested here: #940 

*Description of changes:*
Now project_dir always will be stored as absolute path in the context. This will allow use absolute or relative paths in the --project-dir CLI parameter.


igorborgest:
I used the `pytest.mark.parametrize` feature to improve the code coverage. I think that increases the coverage in a elegant way.

PR #1041, opened 2019-01-08
Add websocket support

stealthycoin:
PR to add websocket support to chalice. There is a follow up PR to this prepared here for adding cloudformation support: https://github.com/stealthycoin/chalice/commit/cb566190f3940b5dd43e515a7a76ab3dd24d9297.

Example:
```python
from boto3.session import Session
from chalice import Chalice
from chalice import WebsocketDisconnectedError
app = Chalice(app_name="echo-server")
app.websocket_api.session = Session()
app.experimental_feature_flags.update([
    'WEBSOCKETS'
])

@app.on_ws_message()
def message(event):
  try:
    app.websocket_api.send(
      connection_id=event.connection_id,
      message=event.body,
    )
  except WebsocketDisconnectedError as e:
    pass  # Disconnected so we can't send the message back.

@app.on_ws_connect()
def connect(event):
  print('%s connected' % event.connection_id)

@app.on_ws_disconnect()
def disconnect(event):
  print('%s disconnected' % event.connection_id)
```


igorborgest:
Great work @stealthycoin !
I'm very excited with the implementation of this feature.

1. I prefer the first option. And although I know that "Explicit is better than implicit", I think this is a case where the "implicit" will bring more simplicity and welcome abstraction.
Otherwise, speaking about semantic, I think that "@app.*" is also better because the websocket is part of the same application with the other triggers, so it should use the same decorator pattern.

2. That looks great to me:
```python
import boto3.session.Session

app = Chalice()
app.websocket.session = boto3.session.Session()
```
And will we also have a default session? Because I think that most of the time we will not need custom sessions.

stealthycoin:
@igorborgest I think I'm leaning toward doing the same thing. I don't see multiple APIs per application as being useful. Making a separate app seems like a fair solution to needing multiple apps. They can take a shared dependency if needed.

A default session is problematic since we have no dependency on boto3/botocore currently in . There is also no guarantee  that the environment will always have `boto3` available (or a correct version). So we are basically doing this as a way to inject the dependency on boto3. I suppose we could also do something like trying to import it and making a session if we succeed and throwing an error if we don't.


Also to summarize some out of band discussion I had with @jamesls 
It would good idea to re-organize the handlers from using their name to have special meaning (which maps to the route key right now) and instead have three decorators `on_connect`, `on_disconnect`, `on_message` (or something) and treat them as separate things. 
One thing I noticed when playing with it was that the different route mappings don't really do anything except add complexity in the deployer planning stage. We do all the routing of events to a handler in the runtime code anyway, so having extra routes other than `$default` doesn't really do anything.

igorborgest:
@stealthycoin @jamesls 

"_Making a separate app seems like a fair solution to needing multiple apps. They can take a shared dependency if needed._" 👍 

Now I see the problem with the default session. I had not noticed that boto3/botocore isn't dependencies. So the "trying to import" suggestion sounds good to me.
Another option would be do raw requests against the API Gateway service… Do you think that it will need too much work?

The new idea of create others decorator sounds good as well. Maybe the route should be passed as argument in the `on_message` decorator, with the `$default` as in fact the default argument.

stealthycoin:
Now that experimental feature flags has been merged in I will be getting back to this and making it an experimental feature.

djn658:
Are there any updates on this?

stealthycoin:
Three handlers are now registered for websockets. on_connect,
on_disconnect, and on_message. on_message handles all messages and the
names of these functions is no longer significant. Currently the
on_message is a single function that accepts every message. We could
add a chalice routing layer that allows on_message to be registered
with a `action` kwarg that calls a different handler depending on the
value of the api gateway event action.

To send messages over a websocket the app.websocket_api property is
now available. This is needs two things to be configured properly:
* a boto3 session
* have the configure method called with the domain name and stage of
  the websocket api

The needs to be added explicitly like this:
`app.websocket_api.session = boto3.session.Session()`

This is because Chalice does not take a dependency on boto3 directly,
and in order to send a message we need to use boto3. Giving the app a
session object will allow it to create the appropriate client when
needed.

The configure method pieces together the endpoint to make boto3 client
calls with to. Any event that is triggered from a websocket api will
provide these as a part of the event. For convenience on_message,
on_connect and on_disconnect when called will all automatically plug
these values out of the configure method.

This leaves out the case where you want to send a message over a
websocket in response some other event source. For example, say a
scheduled event that pushes some piece of system state to every
connected client. The handler for this would need to manually call
`app.websocket_api.configure(domain_name, stage)`. Which means the end
user would be responsible for figuring out what those values are. This
can probably be made better so that Chalice always has the
websocket_api preconfigured. Not sure the best way to do this without
querying an API or setting some environment variables during
deployment.

FINALLY, to actually send a message use
`app.websocket_api.send(connection_id, message)`. If the app is not
correctly configured to send messages it will raise a `ValueError`.
If the connection_id is not connected, a `WebsocketDisconnectedError`
will be raised with a `connection_id` property.



Here is a revision of the above app with the new API. Let me know what you think
@jamesls @igorborgest 
```python
import os
import json

import boto3
from boto3.session import Session

from chalice import Chalice
from chalice import WebsocketDisconnectedError

app = Chalice(app_name="chalice-chat")
app.websocket_api.session = Session()
app.experimental_feature_flags.update([
    'WEBSOCKETS'
])

TABLE = boto3.resource('dynamodb').Table(os.environ.get('TABLE'))


@app.on_connect()
def connect(event):
    connection_id = event.connection_id
    TABLE.put_item(
        Item={
            'connectionId': connection_id,
        },
    )


@app.on_disconnect()
def disconnect(event):
    _remove_connection_id(TABLE, event.connection_id)


@app.on_message()
def message(event):
    body = event.json_body
    message = json.dumps({'message': body['message']})
    for connection_id in [i['connectionId'] for i in TABLE.scan()['Items']]:
        try:
            print('Sending %s to %s' % (message, connection_id))
            app.websocket_api.send(
                connection_id=connection_id,
                message=message,
            )
        except WebsocketDisconnectedError as e:
            _remove_connection_id(TABLE, e.connection_id)


def _remove_connection_id(table, connection_id):
    try:
        table.delete_item(
            Key={
                'connectionId': connection_id,
            },
        )
    except Exception as e:
        print(e)

```

For reference the new wrapper event object that is passed to the three websocket handlers is:
```python
class WebsocketEvent(BaseLambdaEvent):
    def __init__(self, event_dict, context):
        super(WebsocketEvent, self).__init__(event_dict, context)
        self._json_body = None

    def _extract_attributes(self, event_dict):
        request_context = event_dict['requestContext']
        self.domain_name = request_context['domainName']
        self.stage = request_context['stage']
        self.connection_id = request_context['connectionId']
        self.body = event_dict.get('body')

    @property
    def json_body(self):
        if self._json_body is None:
            try:
                self._json_body = json.loads(self.body)
            except ValueError:
                raise BadRequestError('Error Parsing JSON')
        return self._json_body

```

chalice/constants.py
```diff
@@ -238,3 +238,12 @@ def index():
     ],
     "Resource": "*",
 }
+
+
+POST_TO_WEBSOCKET_CONNETION_POLICY = {
```
JordonPhillips: *CONNECTION

chalice/deploy/planner.py
```diff
@@ -722,6 +594,170 @@ def _plan_scheduledevent(self, resource):
...
+                            output_var='region_name'),
+            # The swagger doc uses the 'api_handler_lambda_arn'
+            # var name so we need to make sure we populate this variable
+            # before importing the rest API.
+            models.CopyVariable(from_var=varname,
+                                to_var='api_handler_lambda_arn'),
+        ]  # type: List[InstructionMsg]
+
+        # There's also a set of instructions that are needed
+        # at the end of deploying a rest API that apply to both
```
JordonPhillips: a rest api?

chalice/app.py
```diff
@@ -748,16 +833,33 @@ def _register_handler(self, handler_type, name, user_handler,
...
-        # Sometimes the event can be something that's not
-        # what we specified in our request_template mapping.
-        # When that happens, we want to give a better error message here.
-        resource_path = event.get('requestContext', {}).get('resourcePath')
-        if resource_path is None:
-            return error_response(error_code='InternalServerError',
-                                  message='Unknown request.',
-                                  http_status_code=500)
+    def _register_websocket(self, name, user_handler, handler_string,
+                            kwargs, **unused):
```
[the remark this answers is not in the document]
stealthycoin: I think its forwards compat. All the others have it as well.

chalice/app.py
```diff
@@ -1119,6 +1237,13 @@ def __init__(self, name, handler_string, queue, batch_size):
         self.batch_size = batch_size
 
 
+class WebsocketConfig(BaseEventSourceConfig):
+    def __init__(self, name, route_key_handled, handler_string, user_handler):
+        super(WebsocketConfig, self).__init__(name, handler_string)
+        self.route_key_handled = route_key_handled
```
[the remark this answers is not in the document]
stealthycoin: Its the route key handled by the handler function.

chalice/awsclient.py
```diff
@@ -968,6 +981,130 @@ def verify_event_source_current(self, event_uuid, resource_name,
...
+            self,
+            api_id,
+            lambda_function,
+    ):
+        # type: (str, str) -> str
+        client = self._client('apigatewayv2')
+        return client.create_integration(
+            ApiId=api_id,
+            ConnectionType='INTERNET',
+            ContentHandlingStrategy='CONVERT_TO_TEXT',
```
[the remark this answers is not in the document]
stealthycoin: At the moment no binary. I imagine we would add this with a feature request for it.

chalice/awsclient.py
```diff
@@ -968,6 +981,130 @@ def verify_event_source_current(self, event_uuid, resource_name,
...
+            RouteResponseSelectionExpression='$default',
+            Target='integrations/%s' % integration_id,
+        )
+
+    def create_route_if_needed(self, api_id, route_key, integration_id,
+                               existing_routes):
+        # type: (str, str, str, Dict[str, str]) -> None
+        client = self._client('apigatewayv2')
+        if route_key not in existing_routes:
+            client.create_route(
```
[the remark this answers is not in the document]
stealthycoin: Yea.

chalice/package.py
```diff
@@ -241,6 +241,17 @@ def _inject_restapi_outputs(self, template):
     # noop methods to indicate we've accounted for these
     # resources.
 
+    def _generate_websocketapi(self, resource, template):
+        # type: (models.WebsocketAPI, Dict[str, Any]) -> None
+        message = (
+            "Unable to package chalice apps that use any websocket "
+            "decorators. @app.on_ws_connect, @app.on_ws_disconnect, "
+            "@app.on_ws_message. CloudFormation does not support "
```
kgutwin: As of Feb 8, this is no longer true - CloudFormation has support for API Gateway v2: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html

kgutwin:
Would you consider adding CloudFormation support in the initial release of this PR? We're very interested in this feature, and our build chain relies on `chalice package`. CloudFormation support for API Gateway v2 was announced a little while ago: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html

chalice/package.py
jamesls: We should look into this.  My concern is that we're using SAM to generate the event handlers to integrate with lambda, and it doesn't look like SAM supports web sockets yet https://github.com/awslabs/serverless-application-model/issues/727.

I'd be interested to see what a sample template looks like, but I'm a little concerned about going down this route especially if part of a resource comes from SAM and part comes from CFN.  It's not clear if we can migrate to the SAM version, when support is added, in a backwards compatible way.

Worth looking into, but given the size of this PR I'd suggest adding it as a follow up PR (assuming we find a reasonable way to do it).

stealthycoin:
@kgutwin I have a branch ready to send another PR once this one is merged. https://github.com/stealthycoin/chalice/commit/cb566190f3940b5dd43e515a7a76ab3dd24d9297

docs/source/topics/websockets.rst
```diff
@@ -0,0 +1,1081 @@
...
+  > foo bar baz
+  < foo bar baz
+  >
+
+
+Every message sent to the server (lines that start with ``>``) result in a
+message sent to us (lines that start with ``<``) with the same content.
+
+
+Chat Server Example
```
kyleknap: Talked offline about this. Just recapping... The example is really cool. I do think it would be worth putting this into the Chalice Workshop and linking to the workshop instead. Mainly, most of the time the examples in the chalice docs are relatively short to get the idea of how to use the feature. If there is a lot of boilerplate code, the important part can often get lost. We can then update the echo server example to log out `on_ws_connect` and `on_ws_disconnect` to show how to use those decorators.

docs/source/topics/websockets.rst
```diff
@@ -0,0 +1,1081 @@
...
+need to import and instantiate a boto3 session. This session is manually
+assigned to ``app.websocket_api.session``.
+This is needed because in order to send websocket responses to API Gateway we
+need to construct a boto3 client. Chalice does not take a direct dependency
+on boto3 or botocore, so we need to provide the Session ourselves.
+
+.. code-block:: python
+
+   from boto3.session import Session
+   app.websocket_api.session = Session()
```
kyleknap: I'm a little hesitant to use the boto3 session directly. Mainly because it does not suit well for users that want to have more granular control over how the client gets instantiated and how the message gets sent back. I think it would be worth exploring how we can have a builtin `WebsocketSession` that you could subclass from or provide your own session to. I will get back to providing more ideas for this once I've had a deeper look at the code.

docs/source/topics/websockets.rst
```diff
@@ -0,0 +1,1081 @@
...
+
+
+   @app.on_ws_message()
+   def message(event):
+   try:
+       app.websocket_api.send(
+	   connection_id=event.connection_id,
+	   message=event.body,
+       )
+   except WebsocketDisconnectedError as e:
```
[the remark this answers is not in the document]
stealthycoin: Itll be an uncaught exception and terminate the lambda and go in the logs.

chalice/app.py
```diff
@@ -820,6 +922,22 @@ def __call__(self, event, context):
...
 
+    def __call__(self, event, context):
+        # This is what's invoked via lambda.
+        # Sometimes the event can be something that's not
+        # what we specified in our request_template mapping.
+        # When that happens, we want to give a better error message here.
+        route_key = event.get('requestContext', {}).get('routeKey')
+        resource_path = event.get('requestContext', {}).get('resourcePath')
+        if route_key is not None:
+            return self._handle_websocket(route_key, event, context)
```
jamesls: I'd prefer we not reuse the same lambda function for both rest APIs and websocket APIs.  We've had feedback that customers would even like the individual route handlers to be separate lambda functions.

I think at a minimum we should have a separate lambda function for websockets, and a separate one for rest APIs.  I wouldn't be surprised if we get asked for separate functions for connect/disconnect as well.

docs/source/topics/experimental.rst
```diff
@@ -84,6 +84,12 @@ The status of an experimental API can be:
     - Trial
     - `#1023 <https://github.com/aws/chalice/pull/1023>`__,
       `#651 <https://github.com/aws/chalice/pull/651>`__
+  * - :doc:`websockets`
+    - ``WEBSOCKETS``
+    - 1.7.1
```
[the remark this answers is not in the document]
jamesls: Yeah this will likely be 1.8.0.

djn658:
Any follow up on this PR? We would appreciate your efforts.

chalice/app.py
kapilt: there's several places this notion is wired in swagger generation that needs to be addressed to resolve that end goal of multiple functions on routes, most of which is fixing the calling abstractions to pass in better context, atm they reference a static string ref due to the separation of context and need to reference a common var. is there an extant issue re that function route separation?

chalice/app.py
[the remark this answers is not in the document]
kapilt: isn't it just a route_key then as a simpler name? its not a boolean which is one semantic interpretation of `_handled`

docs/source/tutorials/websockets.rst
```diff
@@ -0,0 +1,1074 @@
...
+
+   app = Chalice(app_name="echo-server")
+   app.websocket_api.session = Session()
+   app.experimental_feature_flags.update([
+       'WEBSOCKETS'
+   ])
+
+
+   @app.on_ws_message()
+   def message(event):
```
jamesls: We should make sure we can handle redeploys if you don't have a connect/disconnect handler.

docs/source/tutorials/websockets.rst
```diff
@@ -0,0 +1,1074 @@
...
+
+.. code-block:: python
+
+   except WebsocketDisconnectedError as e:
+       pass  # Disconnected so we can't send the message back.
+
+
+To test out the echo server you can install ``websocket-client`` from PyPI::
+
+  pip install websocket-client
```
jamesls: I was getting SSL cert errors on python3.7, but python2.7 worked fine.  Even tried updating their cacert.pem file they bundle but still didn't work for python 3.7.  It'd be good to see if you're seeing this error and if so, see if we need use something else.

docs/source/tutorials/websockets.rst
```diff
@@ -0,0 +1,1074 @@
...
+  < foo
+  > bar
+  < bar
+  > foo bar baz
+  < foo bar baz
+  >
+
+
+Every message sent to the server (lines that start with ``>``) result in a
+message sent to us (lines that start with ``<``) with the same content.
```
jamesls: It'd be helpful for this tutorial to follow the structure of the tutorial below, where you say when to write out the requirements files, and the command to run to deploy the echo server.

chalice/app.py
```diff
@@ -495,6 +500,49 @@ def default_binary_types(self):
...
+                'Assign app.websocket_api.session to a boto3 session before '
+                'using the WebsocketAPI'
+            )
+        if self._client is not None:
+            return
+        endpoint = self._WEBSOCKET_ENDPOINT_TEMPLATE.format(
+            domain_name=domain_name,
+            stage=stage,
+        )
+        self._client = self.session.client(
```
jamesls: It'd be nice to defer this client creation until the `send()` call given the `client()` call can have side effects.

docs/source/tutorials/websockets.rst
```diff
@@ -0,0 +1,1074 @@
...
+
+
```
jamesls: We should add a note about  `chalice logs -n websocket_message` if they see any errors when they deploy their app.

chalice/app.py
jamesls: Just following up here, it'd be nice to have the websocket handler code as a separate class, similar to what we've done for new event handlers.  That way they're decoupled from the `Chalice` object.

docs/source/topics/websockets.rst
jamesls: Chatted offline about this.  The thing that I was hoping to improve was to unify the configuration of the `WebsocketAPI` because you have to do both `websocket_api.session = Session()` as well as `websocket_api.configure(..., ...)` to get the object to a point where `send()` will work correctly.  I get why it's two steps, the `configure()` call actually happens in the chalice runtime, but it would be nice to consolidate.

However, there's the more general question about whether we should use boto3 session's directly.  My take on it is that it seems like a reasonable middle ground for the time being, but I think you're right about people wanting more control over the clients and how they get instantiated.  Either that or they want us to take care of everything.

John had an interesting idea about how chalice could have better automatic support with boto3.  We don't include it right now because it's a large dependency, especially if you don't need it.  A default chalice app is a couple of small .py files, which is nice.  However, what if we know you *do* need it?

The chalice mentality is to do things for the user they we know we need to do, but let the override it if they want (e.g. policy generation).  So one idea (and it's just an idea for now) is to detect if the user needs boto3, and if they do, to just _make it work_.  For example, if they have websocket routes defined, and they're calling `websock_api.send()`, then you _must_ have a boto3 client to do that.  So when we detect that we can check:

* Is boto3 in the deployment package?  If not, automatically include it for them.
* Is there a boto3 `_client` on the websocket api?  If not create one for them.

At any point the user can override this (add boto3 to their requirements.txt file), or manually set a client on the websocket API, but by default we can just try to do the right thing for them.  Obviously there's a lot of details and edge cases to work out, but I think it's a really interesting idea.

For the meantime, given this is an experimental API, as long as we clearly document that this API might change I'm ok with keeping this for now, but I'm interested to hear any alternatives if you have them.

docs/source/tutorials/websockets.rst
stealthycoin: I had no problems on 3.7.

requirements-dev.txt
```diff
@@ -20,6 +20,7 @@ py==1.5.3
 pygments==2.1.3
 mock==2.0.0
 requests==2.20.0
+websocket-client==0.54.0
```
kyleknap: We need to add  boto3 as well if we want to be able to run the tests for the websockets app from scratch,

chalice/package.py
```diff
@@ -252,6 +252,17 @@ def _inject_restapi_outputs(self, template):
...
+            "API Gateway Websocket APIs."
+            "You can deploy this app using `chalice deploy`."
+        )
+        raise NotImplementedError(message)
```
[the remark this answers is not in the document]
stealthycoin: I do have a follow up branch ready. Ill reabase once merged.

docs/source/api.rst
```diff
@@ -622,6 +658,64 @@ APIGateway
...
+         import botocore
+         from boto3.session import Session
+         from chalice import Chalice
+
+         app = Chalice('example')
+         session = botocore.session.Session()
+         session.set_config_variable('retries', {'max_attempts': 0})
+         app.websocket_api.session = Session(botocore_session=session)
+
+   .. method:: configure(domain_name, stage)
```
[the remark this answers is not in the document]
stealthycoin: They may want to in handlers like scheduled events where the triggering event was not based on API Gateway v2. In those cases, if they wanted to send websocket messages they would need to call configure and plumb through the values themselves. Definitely something I see being useful so I want to expose it.

docs/source/api.rst
```diff
@@ -622,6 +658,64 @@ APIGateway
...
+      be used.
+
+      .. code-block:: python
+
+         import botocore
+         from boto3.session import Session
+         from chalice import Chalice
+
+         app = Chalice('example')
+         session = botocore.session.Session()
```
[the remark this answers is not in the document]
stealthycoin: The point here was to show sending custom configuration to the botocore session.

Makefile
```diff
@@ -42,8 +42,9 @@ doccheck:
 	##### DOC8 ######
 	# Correct rst formatting for documentation
 	#
+	# TODO: Remove doc8
 	##
-	doc8 docs/source --ignore-path docs/source/topics/multifile.rst
+	doc8 docs/source --ignore-path docs/source/topics/multifile.rst --ignore-path docs/source/tutorials/websockets.rst
```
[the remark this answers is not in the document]
stealthycoin: It uses things that are valid but doc8 doesn't support.

PR #1044, opened 2019-01-13
Provide Lambda context in event object

### LINKED ISSUE #856: `context` parameter for `@app.schedule` lambdas
Currently we do not provide `context` to `@app.schedule` lambdas. It may be useful in some cases when the user wants to know who I am and maybe check remaining times.
To be backward compatible, the `ScheduledEventHandler.__call__` needs to check client code's signature to support 1-param version (`event`) and 2-param version (`event`, `context`) client code.

By the way, the `event` in `@app.schedule` is different from `event` in `@app.lambda_function`, which may arise confusion. Maybe we can rename the parameter name to things like `schedule_event` or `cloudwatch_event`?

alephyud:
Fixes #856.

Adds a `context` attribute for the `event` argument of Cloudwatch, S3, SNS and SQS Lambda handlers, as suggested [in the issue discussion](https://github.com/aws/chalice/issues/856#issuecomment-395506102). That can be used for logging or other purposes:

```python
@app.on_sns_message(topic='my-demo-topic')
def handle_sns_message(event):
    app.log.debug("Received message with subject: %s, message: %s, request id: ",
                  event.subject, event.message, event.context.aws_request_id)
```

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -992,7 +992,7 @@ class EventSourceHandler(object):
         self.event_class = event_class
 
     def __call__(self, event, context):
-        event_obj = self.event_class(event)
+        event_obj = self.event_class(event, context)
         return self.func(event_obj)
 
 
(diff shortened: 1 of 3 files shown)


jamesls:
Thanks for the pull request, taking a look now.

Issue #1019, opened 2018-12-14
Support experimental/provisional APIs


jamesls:
This is a follow up that originally came from a [discussion on a PR](https://github.com/aws/chalice/pull/651#issuecomment-444710490).

Chalice is now GA (> 1.0) so we maintain backwards compatibility of our API.  Specifically, chalice apps that are written against chalice 1.x will continue to work for all future versions of 1.x.

At the same time, Chalice has a lot of potential for new ideas and new APIs, many of which will take several iterations to get right.  I suspect that in many cases we'll need to implement an initial idea, get customer feedback, try it out ourselves, and make corresponding adjustments.  This may include anything from backwards incompatible changes to outright removal of a feature if it ends up not being a good idea.

There's been a few ideas on how to do this that I'll try to summarize here.

## Ideas

* **Documentation only** - Clearly document if an API is experimental vs. stable, but otherwise make no changes to the code itself.  Features/PRs for experimental features get merged into master and treated like any other code.  This is somewhat similar to how python handles provisional APIs ([pep 411](https://www.python.org/dev/peps/pep-0411/)).

* **Namespace** - Add an `experimental` subpackage in chalice, and add experimental features under that namespace.  You'd then have to import explicitly form the experimental package.  I'm not sure what exactly this would look like (haven't done this before), but maybe if we wanted to add a new decorator we might have `from chalice import experimental; app = Chalice(__name__); experimental.inject_some_new_app_decorators(app)`.  My concern with this is that when a feature graduates to a fully supported feature, that in itself would be a breaking change (we've move it out of experimental), and you'd have to modify your code.

* **Experimental branch** - This option proposes maintaining an `experimental` branch on Github.  A PR would then either target the `master` branch for immediate inclusion or the `experimental` branch.  After some period of time testing, we'd merge `experimental` into `master` after removing any features we want to drop from `experimental`.  The downside to this is that we (the maintainers) have a lot more repo maintenance and merging between branches could be problematic.  It's also possible that not as many people would be willing to try out an `experimental` branch.

[...]

PR #1053, opened 2019-01-28
Add support for experimental features

jamesls:
*Issue #, if available:*

#1019

*Description of changes:*

This adds support for feature flags, as discussed here: https://github.com/aws/chalice/issues/1019#issuecomment-456616101

This PR branches on top of (and targets) the blueprints branch, so after we merge this PR, we can merge blueprints back to master.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -688,6 +688,10 @@ class Chalice(_HandlerRegistration, DecoratorAPI):
         if env is None:
             env = os.environ
         self._initialize(env)
+        self.experimental_feature_flags = set()
+        # This is marked as internal but is intended to be used by
+        # any code within Chalice.
+        self._features_used = set()
 
     def _initialize(self, env):
         if self.configure_logs:
@@ -735,6 +739,7 @@ class Chalice(_HandlerRegistration, DecoratorAPI):
         self.log.setLevel(level)
 
     def register_blueprint(self, blueprint, name_prefix=None, url_prefix=None):
+        self._features_used.add('BLUEPRINTS')
         blueprint.register(self, options={'name_prefix': name_prefix,
                                           'url_prefix': url_prefix})
 
--- a/chalice/deploy/validate.py
+++ b/chalice/deploy/validate.py
@@ -7,6 +7,32 @@ from chalice import app  # noqa
 from chalice.config import Config  # noqa
 
 
+class ExperimentalFeatureError(Exception):
+    _ERROR_MSG = (
+        'You are using experimental features without explicitly opting in.\n'
+        'Experimental features do not guarantee backwards compatibility '
+        'and may be removed in the future.\n'
+        'If you still like to use these '
+        'experimental features, you can opt in by adding this to your '
+        'app.py file:\n\n%s\n\n'
+        'See https://doclink/ for more details.'
+    )
+    def __init__(self, features_missing_opt_in):
+        # type: (Set[str]) -> None
+        self.features_missing_opt_in = features_missing_opt_in
+        msg = self._generate_msg(features_missing_opt_in)
+        super(ExperimentalFeatureError, self).__init__(msg)
+
+    def _generate_msg(self, missing_features):
+        # type: (Set[str]) -> str
+        opt_in_line = (
+            'app.experimental_feature_flags.update([\n'
+            '    %s\n'
+            '])\n' % ', '.join(["'%s'" % feature
+                                for feature in missing_features]))
+        return self._ERROR_MSG % opt_in_line
+
+
 def validate_configuration(config):
     # type: (Config) -> None
     """Validate app configuration.
@@ -23,6 +49,17 @@ def validate_configuration(config):
     _validate_manage_iam_role(config)
     validate_python_version(config)
     validate_unique_function_names(config)
+    validate_feature_flags(config.chalice_app)
+
+
+def validate_feature_flags(chalice_app):
+    # type: (app.Chalice) -> None
+    missing_opt_in = set()
+    for feature in chalice_app._features_used:
+        if feature not in chalice_app.experimental_feature_flags:
+            missing_opt_in.add(feature)
+    if missing_opt_in:
+        raise ExperimentalFeatureError(missing_opt_in)
 
 
 def validate_routes(routes):
--- a/chalice/deploy/validate.py
+++ b/chalice/deploy/validate.py
@@ -5,20 +5,10 @@ from typing import Dict, List, Set, Iterator, Optional  # noqa
 
 from chalice import app  # noqa
 from chalice.config import Config  # noqa
+from chalice.constants import EXPERIMENTAL_ERROR_MSG
 
 
 class ExperimentalFeatureError(Exception):
-    _ERROR_MSG = (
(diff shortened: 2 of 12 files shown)


jamesls:
The doc linkchecker keeps failing on travis, though I've manually verified that the links work.  I've added retries to this link checker: https://github.com/aws/chalice/pull/1054

chalice/app.py
```diff
@@ -688,6 +688,10 @@ def __init__(self, app_name, debug=False, configure_logs=True, env=None):
         if env is None:
             env = os.environ
         self._initialize(env)
+        self.experimental_feature_flags = set()
+        # This is marked as internal but is intended to be used by
+        # any code within Chalice.
+        self._features_used = set()
```
kyleknap: Hmm I'm not sure if I like this trend of marking something as an internal property and then accessing it in the deployer code. I think we did something similar in blueprints. I get why we are doing it right now, but I'm wondering if there are any alternatives to tell users that it is an internal property? Options that I am thinking include: Just documenting it as internal (somewhat similar to some of the standard library) or having like an property like `app.internal_stuff` that could be a grab bag for anything that needs to be accessed in the deployer but should not be used in a user's Chalice app.

jamesls: I think we've learned that lesson from botocore that this isn't a great idea.  It basically comes down to either:

a) We make it public, call it something that's hopefully obvious (e.g. `internal_stuff`) and document that users shouldn't touch it and hope that they don't.

b) Make it actually internal with a `_` prefix, and then we (chalice devs) purposefully access internal attributes we know are safe.

I think (b) is a much safer alternative, especially given that the `_` prefix is a much more explicit mechanism.   We need the concept of "package private" and I would prefer the option that requires chalice devs to be mindful of what they're accessing (e.g that `_features_used` is safe) vs. customers knowing that they shouldn't access an attribute called `internal_features_used`.

stealthycoin: Agree with james. I think the convention for a friend-ish class should just be marked private, its a lot safer for everyone involved.

chalice/deploy/validate.py
```diff
@@ -7,6 +7,34 @@
 from chalice.config import Config  # noqa
 
 
+class ExperimentalFeatureError(Exception):
+    _ERROR_MSG = (
```
stealthycoin: Thoughts on putting this in constants with (some) of our other error messages?

chalice/deploy/validate.py
```diff
@@ -7,6 +7,34 @@
...
+        # type: (Set[str]) -> None
+        self.features_missing_opt_in = features_missing_opt_in
+        msg = self._generate_msg(features_missing_opt_in)
+        super(ExperimentalFeatureError, self).__init__(msg)
+
+    def _generate_msg(self, missing_features):
+        # type: (Set[str]) -> str
+        opt_in_line = (
+            'app.experimental_feature_flags.update([\n'
+            '    %s\n'
```
stealthycoin: Nitpick but since this is all one line if we have a bunch of features this would generate a longer-than-80 line. Probably pretty easily. Also not hard to change to be one element per line so its always pep8.

chalice/deploy/validate.py
```diff
@@ -23,6 +51,18 @@ def validate_configuration(config):
     _validate_manage_iam_role(config)
     validate_python_version(config)
     validate_unique_function_names(config)
+    validate_feature_flags(config.chalice_app)
+
+
+def validate_feature_flags(chalice_app):
```
stealthycoin: Do we want to have some sort of warning/alert that something has been removed from the feature flag list and is now a stable feature? Looks like it doesn't _really_ matter but if I upgrade chalice and redeploy and blueprints was moved to stable. I will still have my `BLUEPRINTS` entry in the experimental feature flags list when not necessary.

docs/source/topics/experimental.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+List of Experimental APIs
+=========================
+
+In the table below, the "Feature Flag Name" column is the value you
+must add to the ``app.experimental_feature_flags`` attribute.
+The status of an experimental API can be:
+
+* ``Trial`` - You must explicitly opt-in to use this feature.
```
stealthycoin: Could we add a column here for a link to context. PR/Feature request thread or something like that. Especially useful for tracking when/why something was rejected/accepted.

chalice/app.py
kyleknap: I guess its fine. The only thing I do not like about the `_` prefix is that it can be hard to know which `_` properties are public for our development only. The only way to really tell is to look for the comment in the code where it is defined. We also have to be diligent in making sure we test the `_` property so we do not accidentally refactor it in a backwards incompatible way (but I think we have a good amount of tests for the ones we have added so that should be fine).

docs/source/topics/experimental.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+If you use an experimental API without opting-in, you will receive
+a message whenever you run a Chalice CLI command.  The error message
+tells you which feature flags you need to add::
+
+    $ chalice deploy
+    You are using experimental features without explicitly opting in.
+    Experimental features do not guarantee backwards compatibility and may be removed in the future.
+    If you still like to use these experimental features, you can opt-in by adding this to your app.py file:
```
JordonPhillips: If you would

chalice/deploy/validate.py
```diff
@@ -7,6 +7,34 @@
 from chalice.config import Config  # noqa
 
 
+class ExperimentalFeatureError(Exception):
+    _ERROR_MSG = (
+        'You are using experimental features without explicitly opting in.\n'
+        'Experimental features do not guarantee backwards compatibility '
+        'and may be removed in the future.\n'
+        'If you still like to use these '
```
JordonPhillips: If you would

chalice/deploy/validate.py
JordonPhillips: +1, though it doesn't necessarily need to be added now since we won't be using it yet.

chalice/deploy/validate.py
jamesls: Don't really have a preference, we're not super consistent with this (some of the deployer error messages are in deployer.py).

docs/source/topics/experimental.rst
JordonPhillips: +1

JordonPhillips:
Looks great!

chalice/deploy/validate.py
jamesls: I'd prefer not to.  I think warnings should only be for things that may cause issues for a user (i.e. a deprecation of something).   In this case it's just unnecessary code that runs.

Perhaps in the future we can have a some sort of linting command that can check these things, but I don't think it needs to happen at deploy time.

jamesls:
Not sure what codecov/patch is complaning about, otherwise travis is passing now.

cc @kyleknap @stealthycoin

chalice/deploy/validate.py
stealthycoin: Linting command seems good. Agreed it doesn't need to be deploy time.

Issue #1001, opened 2018-12-02
Implement layers


onurmatik:
AWS recently announced "layers" for sharing code and libraries among lambda functions:
https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html

This eases maintenance when sharing custom code and reduces deployment time when using libraries that are not pure Python. 

Existing `chalicelib` and `vendor` directories can be deployed as layers. Also, a configuration option would be useful for cross-project sharing of resources via `layers` among Chalice projects.

jamesls:
Interesting idea on having chalicelib and vendor as layers.  What I was thinking was maybe we could have *all* your dependencies outside of `app.py` and `chalicelib/` in a single layer.  That has several benefits:

* You only have to update that layer if your requirements.txt or vendor/ directory change.
* Subsequent app deployments are *only* your application code, which would significantly speed up deployments if your app code is much smaller than the code for your dependencies.

The cross project sharing I'd have to think about.  We've definitely ran into cases where a "conceptual" serverless app consists of multiple chalice apps, and it would be really helpful to be able to share/reuse across multiple chalice apps.

I think the next step here would be to get a more fleshed out proposal/spec for layers support.

owenbrown:
Thank you both for working on this exciting project! 

To weigh in, our team has would love to be able to share code, as a layer, across several chalice applications.  

To be specific about our use case, we build chatbots. Different chatbots have different backend functions. But they all use the same "core", which is a cross platform (Actions on Google, Facebook, Web-client) SDK. Layers seemed like a good way to share the core engine with the different agents.

Another use case we have is for analytics. We have many functions that write to AWS RDS. We're a Python shop, and we use SQLAlchemy as an ORM to connect to RDS. Uploading SQLAlchemy with every deploy takes a while. Getting this into a layer would speed our deploys considerably.

sciurus:
Seperate from the larger effort to convert chalice into a layer, it would be helpful if chalice let you declare existing layers to be loaded by the ,ambda it deploys.  For instance, so you could enable the [datadog layer](https://docs.datadoghq.com/integrations/amazon_lambda/#installing-and-using-the-datadog-layer) and then send metrics to datadog from your chalice application.

cedricyau:
+1 for adding Lambda Layer option to config.json

cedricyau:
[...]

PR #1063, opened 2019-02-06
Add layers support

darlev91:
See Issue #1001 

Description of changes: Added layer support to Config and Lambda Function objects.

### DIFF
--- a/chalice/deploy/deployer.py
+++ b/chalice/deploy/deployer.py
@@ -581,6 +581,11 @@ class ApplicationGraphBuilder(object):
                                             security_group_ids)
             )
 
+    def _get_lambda_layers(self, config):
+        # type: (Config) -> List[str]
+        layers = config.layers
+        return layers if layers else []
+
     def _build_lambda_function(self,
                                config,        # type: Config
                                name,          # type: str
@@ -592,6 +597,7 @@ class ApplicationGraphBuilder(object):
         function_name = '%s-%s-%s' % (
             config.app_name, config.chalice_stage, name)
         security_group_ids, subnet_ids = self._get_vpc_params(name, config)
+        lambda_layers = self._get_lambda_layers(config)
         function = models.LambdaFunction(
             resource_name=name,
             function_name=function_name,
@@ -606,6 +612,7 @@ class ApplicationGraphBuilder(object):
             security_group_ids=security_group_ids,
             subnet_ids=subnet_ids,
             reserved_concurrency=config.reserved_concurrency,
+            layers=lambda_layers
         )
         self._inject_role_traits(function, role)
         return function
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -141,6 +141,7 @@ class TypedAWSClient(object):
                         memory_size=None,            # type: OptInt
                         security_group_ids=None,     # type: OptStrList
                         subnet_ids=None,             # type: OptStrList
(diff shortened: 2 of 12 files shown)


chalice/deploy/deployer.py
```diff
@@ -606,6 +606,7 @@ def _build_lambda_function(self,
             security_group_ids=security_group_ids,
             subnet_ids=subnet_ids,
             reserved_concurrency=config.reserved_concurrency,
+            layers=config.layers
```
jamesls: We should add the same logic we have for subnet/security groups where we default to an empty list if no configuration is provided in `.chalice/config.json`.  One of the things I noticed while trying out this PR was that if I had a layer in my chalice config, deploy my app, then remove the layer from my config, the layer will still be part of the function configuration.

darlev91: @jamesls Thanks for your feedback!
We are using AWS layers to give the possibility to use dependencies of operating system unavailable in lambda functions by default.
For instance, we need to use git and ssh clients which is not installed in AMI.
We install git and ssh from source files and build it statically, after that we compressed
installed data and store it as lambda layers.
Some python libraries require dependencies that need to be installed in OS.
For example, GitPython requires executable git binary.

stealthycoin:
Looks good to me.

jpeters71:
So...how exactly is this supposed to work?  If chalice doesn't remove dependencies that are duplicated between the layer and what it has in the vendor dir (or resolved by the requirements.txt file), it makes the layer pointless, right?

I think we're missing something...

Here's the scenario: we have an API that's fairly simple, but has a lot of code behind it. We're running into the AWS lambda size limit (50mb). The biggest package we use is numpy. We add a reference to the numpy layer from our chalice config.json file and try to deploy.  Same problem as before because chalice doesn't look at the layer during the packaging phase so it doesn't know that it doesn't need to package up numpy.

We've tried a few things; like removing numpy from our requirements.txt file.  But, then, chalice won't package it because it can't load our app.py since, deep down in the code that gets pulled in, is a reference to numpy which python can't resolve.

So, where we're at now is: we're using chalice to generate the package with all the dependencies. Then, we manually unzip, remove the packages we know are in the layer (i.e. numpy); zip it back up; use awscli to deploy it.  That seems really cumbersome.  Again, I have to think we're missing something.

Is this the recommended process for doing this?

skwashd:
@jpeters71 this is the initial implementation of layers support. Chalice attaches the layer/s to the function. 

In our case we use gitpython, which requires the `git` binary. A layer was the best way for us to do this. 

Issue #1001 discusses the next steps for how chalice can take full advantage of layers. The features being discussed there should give you the functionality you're after.

vrinda1410:
@jamesls @stealthycoin Is there a tested & clean approach to this question : https://github.com/aws/chalice/pull/1063#issuecomment-516619913

My scenario is similar wherein I have bundled few functions from my code & published them as a lambda layer. e.g. DB Connection code
But I can't seem to call that as layer in my code as neither "chalice deploy" nor "chalice package" is working. I've tried the hacky solutions mentioned above like emptying out requirements.txt

Error - ModuleNotFoundError: No module named 

If I try to add this code to initial zip & then remove it later, that seem to waste the initial effort of using layer.

Please provide suggestions here or if this functionality is not yet implemented in chalice, can somebody make a note of handling this in future releases. Thank you.

PR #1078, opened 2019-02-21
Handle multiple content types

melevittfl:
This is related to issue #391. 

I've been developing a Chalice app that generates a binary image and returns the image. The app was failing because Chalice was checking if the Response header was binary (in this case image/png), but the Accept header was from a browser so had multiple media types.

However, the Response object only checks the first media type in the request Accept header and ignores all others. This means that a browser-based request fail with binary data as browser is rarely going to have the exact media type first in the list.


stealthycoin:
@melevittfl Thank you for the contribution, I put a few commits on top to add tests and address whitespace in the Content-Type header as well as add a Changelog entry.

New PR: #1116

PR #1091, opened 2019-03-15
Change suggested venv dir in CONTRIBUTING.rst

nkantar:
*Issue #, if available:*
No issue

*Description of changes:*
The Contributing guide currently suggests the user set up the virtual environment in `venv-chalice/`. This directory isn't in `.gitignore`, and thus Git shows it as unstaged. However, as `.gitignore` already has `venv`, the simplest solution is to simply change the instructions, which is what this commit does.


stealthycoin:
Nice catch, thanks for the PR.

Issue #1085, opened 2019-03-06
Project seems dead due to Constantly being Incompatible with newer versions of PIP and its own dependencies.


duaneking:
## We cant use Chalice if we are following python best practices and using the most up to date version of pip.

Python packages shouldn't be dependent on a specific version of pip anyway, that in itself is a serous defect that creates more work consistently as new versions of pip are released, and pip gets updated all the time, so Chalice is broken all the time since its constantly hitting a new version it thinks it cant use.

## This creates more work for everybody.

Every time this happens, somebody tries to be helpful and files a defect, that defect is ignored for weeks while people rage about arbitrary version conflicts, and then that defect is ultimately closed after being marked as a feature request when in reality, its not a feature request, its a basic functional requirement.  Then a day later somebody else updates pip and the cycle repeats.

## The dev team has a history of not being community focused.

Reading the closed bugs that are still reproducing - because the Chalice dev team seems to really enjoy closing open and reproducible defects as "feature requests" without fixing them and i have no idea why this is -  the version limit in place for PIP is in place for the absolute wrong reasons, out of fear instead of metrics that can be acted on, yet other projects dont have this issue and are perfectly working fine on AWS without the same pip version restrictions.

## The old dependencies are getting harder to maintain

Add to this the fact that its using an out of date version of click and other dependencies, and we now have multiple defects spread across the project created due to the increasingly difficult maintenance right now, as this project actively looks unmaintained and dead due to lack of support from AWS on these issues, and as time goes on these issues are getting bigger and bigger as the package versions getting used are getting older and older and start to conflict more and more with a modern python dependency map using up to date versions of basic things like pip and click.  Its a ticking time bomb.

## People cant trust Chalice if it wont even install in pip.
Due to this,  people are not investing in Chalice or using it for projects like they could be, because they can not trust that it will be around and get the correct and proper maintenance support that it requires from AWS in the future, since it looks like AWS has already given up on it, while projects like Zappa, etc are still going strong and don't have these issues with newer versions of PIP like Chalice does, so it doesn't make any sense to us as developers to see other projects not having these issues and Chalice for bad or worse, looks like the black sheep that doesn't work as well.

## Are the developers eating their own dogfood?
[...]

PR #1104, opened 2019-04-09
Fix installation with latest pip version

NaxAlpha:
### Unable to install chalice with latest pip version

Hi, I have tried installing chalice (in a virtual environment). Whenever, I use `pip install -r requirements.txt`, it breaks with following error:

image

I have updated the code so that it will be compatible with latest version of pip and now it works smoothly. Let me know if any other change is required.

Thanks

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ install_requires = [
     'botocore>=1.12.86,<2.0.0',
     'typing==3.6.4;python_version<"3.7"',
     'six>=1.10.0,<2.0.0',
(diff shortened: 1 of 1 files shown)


stealthycoin:
Thanks for the submission. I pushed a few extra commits on top of yours and merged this in #1115. We can't just remove the ceiling on PIP as explained #1085 by both @jamesls and @dtufft, in addition to that I added a changelog entry.

NaxAlpha:
Thanks a lot, it works with the latest pip now.

PR #1119, opened 2019-05-15
Add support for `chalice package --force` option

jmehnle:
When specified, ignores use of `@app.on_s3_event` and other unsupported situations.

My company conceptualizes Chalice as providing three benefits:

* AWS Lambda request handling framework
* Packaging of Lambda functions
* Deploying of Lambda functions and related infrastructure

We, however, have chosen to use only the first two benefits and pass on the third. Instead we use Terraform to manage all our infrastructure, including the Lambda function itself.

Therefore the incompatibility between the use of `@app.on_s3_event` and the use of the `chalice package` command does not concern us. Rather, the fact that it dies from a `NotImplementedError` exception is causing a hindrance for us.

Would you be willing to adopt such a `chalice package --force` option?  If so, I'm willing to bring this PR into an acceptable shape, including unit tests and any documentation that you deem necessary.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Next Release (TBD)
 * Raise TypeError when trying to serialize an unserializable
   type
   (`#1100 <https://github.com/aws/chalice/issues/1100>`__)
+* Add support for ``chalice package --force`` option to ignore use of
+  ``@app.on_s3_event`` and other unsupported situations.
 
 
 1.8.0
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -374,22 +374,26 @@ def generate_sdk(ctx, sdk_type, stage, outdir):
                     "this argument is specified, a single "
                     "zip file will be created instead."))
 @click.option('--stage', default=DEFAULT_STAGE_NAME)
+@click.option('--force', is_flag=True, default=False,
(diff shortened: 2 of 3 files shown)


kapilt:
@jmehnle would this use case be better served by direct terraform support in #1121 (if so upvotes appreciated)?

kapilt:
fwiw, i went ahead and added in direct terraform support in #1129. 

Wrt to on_s3_event, i think a better pattern in general for s3 events is using sns notifications, so multiple consumers can subscribe to notifications, s3 event notifications don't support configuring multiple notification mechanisms over the same path (lambda, sqs, sns), so configuring say a single lambda function on the root of the bucket, will prevent all other notification mechanisms.

jmehnle:
@kapilt, the general use case would be better served by full Terraform support, yes.

I agree that, in more complex environments, using SNS is often unavoidable, however I would strongly argue that *many* environments are *not* complex enough to require it, and often enough direct S3-to-Lambda bucket notifications are completely fine and the S3 use case is one that many people (including myself for *parts*, though not all, of my company's infrastructure) will want to use.

kapilt:
yeah, for the simple cases, keeping it simple sounds good (direct notification integration). just wanted to offer up an option that would work with the existing feature set. fwiw, the terraform support pr does support this use case (bucketnotification configuration).

Issue #1124, opened 2019-05-24
support for subscribing to cloud watch events (not just scheduled)


kapilt:
Currently chalice supports periodic functions via cloud watch events, but doesn't support subscribing to other events (via event pattern). I think it would be very useful to support the fuller range of cloud watch events. In terms of a syntax proposal to kick off discussion.

```python
@app.on_cloud_watch_event({'source': 'aws.ec2'})
def on_instance_state_change(event, context):
     print(event)
```

kapilt:
Fwiw pull request implements this currently as `app.on_cw_event(`

jamesls:
Closing out old proposal, PR was merged a while back.

PR #1126, opened 2019-05-25
cloud watch events support

kapilt:
Issue #1124

*Support subscribing functions to cloud watch events.*

with this pr the intent is that you can now subscribe to any of the dozen of event types that are sent through cloud watch events https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html

Example usage, in this case subscribing to ec2 instance state change notifications

```python
@sample_app.on_cw_event({'source': ['aws.ec2']})
def instance_change(event, context):
    print(event)
```

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -589,6 +589,13 @@ class DecoratorAPI(object):
             registration_kwargs={'queue': queue, 'batch_size': batch_size}
         )
 
+    def on_cw_event(self, event_pattern, name=None):
+        return self._create_registration_function(
+            handler_type='on_cw_event',
+            name=name,
+            registration_kwargs={'event_pattern': event_pattern}
+        )
+
     def schedule(self, expression, name=None):
         return self._create_registration_function(
             handler_type='schedule',
@@ -653,6 +660,7 @@ class DecoratorAPI(object):
             'on_s3_event': S3Event,
             'on_sns_message': SNSEvent,
             'on_sqs_message': SQSEvent,
+            'on_cw_event': CloudWatchEvent,
             'schedule': CloudWatchEvent,
         }
         if handler_type in event_classes:
@@ -797,8 +805,16 @@ class _HandlerRegistration(object):
         )
         self.event_sources.append(sqs_config)
 
-    def _register_schedule(self, name, handler_string, kwargs, **unused):
+    def _register_on_cw_event(self, name, handler_string, kwargs, **unused):
         event_source = CloudWatchEventConfig(
+            name=name,
+            event_pattern=kwargs['event_pattern'],
+            handler_string=handler_string
+        )
+        self.event_sources.append(event_source)
+
+    def _register_schedule(self, name, handler_string, kwargs, **unused):
+        event_source = ScheduledEventConfig(
             name=name,
             schedule_expression=kwargs['expression'],
             handler_string=handler_string,
@@ -1257,10 +1273,16 @@ class BaseEventSourceConfig(object):
         self.handler_string = handler_string
 
 
-class CloudWatchEventConfig(BaseEventSourceConfig):
+class ScheduledEventConfig(BaseEventSourceConfig):
     def __init__(self, name, handler_string, schedule_expression):
-        super(CloudWatchEventConfig, self).__init__(name, handler_string)
+        super(ScheduledEventConfig, self).__init__(name, handler_string)
         self.schedule_expression = schedule_expression
+
+
+class CloudWatchEventConfig(BaseEventSourceConfig):
+    def __init__(self, name, handler_string, event_pattern):
+        super(CloudWatchEventConfig, self).__init__(name, handler_string)
+        self.event_pattern = event_pattern
 
 
 class ScheduleExpression(object):
--- a/chalice/deploy/models.py
+++ b/chalice/deploy/models.py
@@ -188,7 +188,7 @@ class CloudWatchEventBase(FunctionEventSubscriber):
 
 @attrs
 class CloudWatchEvent(CloudWatchEventBase):
-    resource_type = 'cloud_watch_event'
+    resource_type = 'cloudwatch_event'
     event_pattern = attrib()    # type: str
 
 
--- a/chalice/deploy/models.py
+++ b/chalice/deploy/models.py
@@ -182,16 +182,19 @@ class FunctionEventSubscriber(ManagedModel):
 
 
 @attrs
-class CloudWatchEvent(FunctionEventSubscriber):
-    resource_type = 'cw_event'
+class CloudWatchEventBase(FunctionEventSubscriber):
     rule_name = attrib()        # type: str
+
+
+@attrs
+class CloudWatchEvent(CloudWatchEventBase):
+    resource_type = 'cloud_watch_event'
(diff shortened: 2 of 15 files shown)


chalice/deploy/models.py
```diff
@@ -165,17 +165,28 @@ def dependencies(self):
 
 
 @attrs
-class ScheduledEvent(ManagedModel):
-    resource_type = 'scheduled_event'
-    rule_name = attrib()            # type: str
-    schedule_expression = attrib()  # type: str
-    lambda_function = attrib()      # type: LambdaFunction
+class FunctionEventSubscriber(ManagedModel):
+    lambda_function = attrib()  # type: LambdaFunction
```
kyleknap: If we want to refactor this, we will probably want to define the `dependencies()` method for this base class as well. It looks like it got dropped in the refactoring.

chalice/deploy/planner.py
```diff
@@ -686,6 +686,41 @@ def _plan_s3bucketnotification(self, resource):
...
+
+        function_arn = Variable(
+            '%s_lambda_arn' % resource.lambda_function.resource_name
+        )
+
+        plan = [
+            models.APICall(
+                method_name='get_or_create_rule_arn',
+                params={'rule_name': resource.rule_name,
+                        'event_pattern': resource.event_pattern},
```
kyleknap: It would be great if we could consolidate a good portion of this logic into a single shared method with the `_plan_scheduledevent` method. Looking over the code, really the only difference between these two functions are the use of `event_pattern` as a parameter instead of `schedule_event`.

docs/source/topics/events.rst
```diff
@@ -70,6 +70,33 @@ and configure ``every_hour`` to be invoked once an hour,
...
+
+    app = chalice.Chalice(app_name='foo')
+
+    @app.on_cw_event({"source": ["aws.codecommit"]})
+    def on_code_commit_changes(event):
+        print(event.to_dict())
+
+In this example, we have a single lambda function that we subscribe to all
+events from the AWS Code Commit service. The first parameter to the decorator
+is the event pattern that will be used to filter the events sent to the function.
```
kyleknap: It may be worth adding a link to the documentation for event patterns: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/CloudWatchEventsandEventPatterns.html as it is not too obvious as to all the different options available.

chalice/app.py
```diff
@@ -530,6 +530,13 @@ def on_sqs_message(self, queue, batch_size=1, name=None):
             registration_kwargs={'queue': queue, 'batch_size': batch_size}
         )
 
+    def on_cw_event(self, event_pattern, name=None):
```
kyleknap: So I'm a little hesitant on using the name `on_cw_event`. Mainly because we recently rebranded the CloudWatch Events client in `botocore` to be EventBridge. And given EventBridge is a superset of the CloudWatch Event API, we may want to use event bridge functionality that gets launched in the future for this decorator. I'd prefer if we somehow generalize/abstract the name a bit more to remove the `cw` part of it. Maybe `event_pattern` or `patterned_event`? This will make it a little more consistent with the pre-existing CW event decorator`schedule()`.

@stealthycoin @jamesls Any thoughts about this?

kyleknap:
Other thing I forgot... It would be good to add a changelog entry for this. You can add it under a new `Next Release` section in the `CHANGELOG.rst` file

docs/source/topics/events.rst
```diff
@@ -70,6 +70,33 @@ and configure ``every_hour`` to be invoked once an hour,
...
+
+The function you decorate must accept a single argument,
+which will be of type :class:`CloudWatchEvent`.
```
kyleknap: It may be worth looking at the class reference more and make sure the documentation applies to event patterns as well. The documentation when I was looking at this class was a little specific for scheduled events.

docs/source/topics/events.rst
```diff
@@ -70,6 +70,33 @@ and configure ``every_hour`` to be invoked once an hour,
 and ``every_two_hours`` to be invoked once every two hours.
 
 
+.. _cwe-events:
+
```
kyleknap: We need to also make sure to update the `api.rst`: https://github.com/aws/chalice/blob/master/docs/source/api.rst for this as well for reference documentation on the new decorator.

chalice/app.py
kapilt: i'm open on any name suggestions here..   `on_bridge_event` ?

jamesls: I don't have a strong opinion on the name other than it's consistent with our naming convention of `on_<service>_<resource>`, so I guess that would make it `on_eb_event`, or `on_eventbridge_event`.

kyleknap: Yeah consistency is the most important part for me. While it is a little redundant to have `event` twice in the name and a bit longer, I think I'd vote for `on_eventbridge_event` instead of `on_eb_event` because `eb` is more associated with `elasticbeanstalk` and I don't think `eb` is really an associated shortname to event bridge quite yet.

kapilt: i care much more about usability then consistency, ie. scheduled is already inconsistent by that nomenclature, and there's only like a handful of these methods. on_eventbridge_event feels overly verbose and repetitive (aka on event event ).. alternatively we could just keep on on_cw_event which would be less typing, or remove the duplication with on_bridge_event or on_eventbridge.. frankly the _resource suffix is redundant with the prefix `on_` since its always been `event`, which the prefix denotes.

kyleknap: @jamesls what is your thoughts on `on_eventbridge`? I do agree with @kapilt that it is definitely redundant to have event twice in the name, especially given the service revolves around emitting events. Otherwise, I'm fine just making it `on_eb_event`. I really want to see us come up with a decision on this so we can proceed with the PR

kapilt: per discussion, going to continue with on_cw_event given its predominant extant usage. we may alias this in future to also support event bridge nomenclature.

docs/source/topics/events.rst
kapilt: i've added a little more clarification there that its for both cloud watch events/event bridge

chalice/deploy/planner.py
```diff
@@ -589,51 +596,21 @@ def _plan_cloudwatchevent(self, resource):
             # You need to remove targets (which have IDs)
             # before you can delete a rule.
             models.RecordResourceValue(
-                resource_type='cloudwatch_event',
+                resource_type=resource.resource_type,
```
kyleknap: So I think it makes sense to keep this hardcoded as `cloudwatch_event`. Specifically in the sweeper, we look for `cloudwatch_event` as the resource type when we try to delete the rule. I think right now (I need to double check), the sweeper will not delete the rule resource. Otherwise, we need to go back to the sweeper and add these resources as strings to look for when deleting the rules.

kapilt: good catch, reverted back to cloudwatch_event static string.

kapilt:
rebased and added changelog and terraform packaging support.

kapilt:
wrt to cloudwatch, updated docs and code to reflect the product marketing name.

PR #1129, opened 2019-05-26
terraform support

### LINKED ISSUE #1121: support terraform config generation?
I keep running into lots of orgs that are strongly committed to terraform. I end up doing gross hacks of layering infrastructure provisioning to try incorporate chalice's deploy/package operations in between terraform executions. But I think the simpler answer would be to just have chalice support terraform generation, curious if that's of interest.

kapilt:
Issue: closes #1121 

This adds in terraform support for chalice package. There's a few goals underlying.. one was to provide provisioning flexibility and extensibility instead of having to layer in multiple provisioning executions for an app (ie reference an s3 bucket, or use chalice lambda function for a step function, etc). Another was to enable chalice use more naturally in orgs that mandate/prefer terraform for infrastructure provisioning. 

I've done a few manual e2e tests on a sample app (api gw, 5 lambdas, scheduled func, managed role). The end result is actually a bit faster than chalice's direct api calls for deploy/destroy.

 https://gist.github.com/kapilt/1794c092cd2f17e0faadb3e00a44bc33

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -1,11 +1,13 @@
+# pylint: disable=too-many-lines
+import json
 import os
 import copy
-import json
 
 from typing import Any, Optional, Dict, List, Set, Union  # noqa
 from typing import cast
 
-from chalice.deploy.swagger import CFNSwaggerGenerator
+from chalice.deploy.swagger import (
+    CFNSwaggerGenerator, TerraformSwaggerGenerator)
 from chalice.utils import OSUtils, UI, serialize_to_json, to_cfn_resource_name
 from chalice.config import Config  # noqa
 from chalice.deploy import models
@@ -15,31 +17,41 @@ from chalice.deploy.deployer import BuildStage  # noqa
 from chalice.deploy.deployer import create_build_stage
 
 
-def create_app_packager(config, merge_template=None):
-    # type: (Config, Optional[str]) -> AppPackager
+def create_app_packager(
+        config, package_format='cloudformation', merge_template=None):
+    # type: (Config, str, Optional[str]) -> AppPackager
     osutils = OSUtils()
     ui = UI()
     application_builder = ApplicationGraphBuilder()
     deps_builder = DependencyBuilder()
-    build_stage = create_build_stage(
-        osutils, ui, CFNSwaggerGenerator()
-    )
-    resource_builder = ResourceBuilder(application_builder,
-                                       deps_builder, build_stage)
-    processors = [
-        ReplaceCodeLocationPostProcessor(osutils=osutils),
-        TemplateMergePostProcessor(
-            osutils=osutils,
-            merger=TemplateDeepMerger(),
-            merge_template=merge_template,
-        ),
-    ]
+    post_processors = []  # type: List[TemplatePostProcessor]
+    generator = None  # type: Union[None, TemplateGenerator]
+
+    if package_format == 'cloudformation':
+        build_stage = create_build_stage(
+            osutils, ui, CFNSwaggerGenerator())
+        post_processors.extend([
+            SAMCodeLocationPostProcessor(osutils=osutils),
+            TemplateMergePostProcessor(
+                osutils=osutils,
+                merger=TemplateDeepMerger(),
+                merge_template=merge_template)])
+        generator = SAMTemplateGenerator(config)
+    else:
+        build_stage = create_build_stage(
+            osutils, ui, TerraformSwaggerGenerator())
+        generator = TerraformGenerator(config)
+        post_processors.append(
+            TerraformCodeLocationPostProcessor(osutils=osutils))
+
+    resource_builder = ResourceBuilder(
+        application_builder, deps_builder, build_stage)
+
     return AppPackager(
-        SAMTemplateGenerator(),
+        generator,
         resource_builder,
-        CompositePostProcessor(processors),
-        osutils,
-    )
+        CompositePostProcessor(post_processors),
+        osutils)
 
 
 class UnsupportedFeatureError(Exception):
@@ -70,7 +82,47 @@ class ResourceBuilder(object):
         return resources
 
 
-class SAMTemplateGenerator(object):
+class TemplateGenerator(object):
+
+    template_file = None  # type: str
+
+    def __init__(self, config):
+        # type: (Config) -> None
+        self._config = config
+
+    def dispatch(self, resource, template):
+        # type: (models.Model, Dict[str, Any]) -> None
+        name = '_generate_%s' % resource.__class__.__name__.lower()
+        handler = getattr(self, name, self._default)
+        handler(resource, template)
+
+    def generate(self, resources):
+        # type: (List[models.Model]) -> Dict[str, Any]
+        raise NotImplementedError()
+
+    def _generate_filebasediampolicy(self, resource, template):
+        # type: (models.FileBasedIAMPolicy, Dict[str, Any]) -> None
+        pass
+
+    def _generate_autogeniampolicy(self, resource, template):
+        # type: (models.AutoGenIAMPolicy, Dict[str, Any]) -> None
+        pass
+
+    def _generate_deploymentpackage(self, resource, template):
+        # type: (models.DeploymentPackage, Dict[str, Any]) -> None
+        pass
+
+    def _generate_precreatediamrole(self, resource, template):
+        # type: (models.PreCreatedIAMRole, Dict[str, Any]) -> None
+        pass
+
+    def _default(self, resource, template):
+        # type: (models.Model, Dict[str, Any]) -> None
+        raise UnsupportedFeatureError(resource)
+
+
+class SAMTemplateGenerator(TemplateGenerator):
+
     _BASE_TEMPLATE = {
         'AWSTemplateFormatVersion': '2010-09-09',
         'Transform': 'AWS::Serverless-2016-10-31',
@@ -78,18 +130,19 @@ class SAMTemplateGenerator(object):
         'Resources': {},
     }
 
-    def __init__(self):
-        # type: () -> None
+    template_file = "sam.json"
+
+    def __init__(self, config):
+        # type: (Config) -> None
+        super(SAMTemplateGenerator, self).__init__(config)
         self._seen_names = set([])  # type: Set[str]
 
-    def generate_sam_template(self, resources):
+    def generate(self, resources):
         # type: (List[models.Model]) -> Dict[str, Any]
         template = copy.deepcopy(self._BASE_TEMPLATE)
         self._seen_names.clear()
         for resource in resources:
-            name = '_generate_%s' % resource.__class__.__name__.lower()
-            handler = getattr(self, name, self._default)
-            handler(resource, template)
+            self.dispatch(resource, template)
         return template
 
     def _generate_scheduledevent(self, resource, template):
@@ -295,7 +348,7 @@ class SAMTemplateGenerator(object):
                         ('arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}'
                          ':${WebsocketAPIId}/*'),
                         {'WebsocketAPIId': api_ref},
-                    ]
+                    ],
                 },
             }
         }
@@ -449,22 +502,6 @@ class SAMTemplateGenerator(object):
             }
         }
 
-    def _generate_filebasediampolicy(self, resource, template):
-        # type: (models.FileBasedIAMPolicy, Dict[str, Any]) -> None
-        pass
-
-    def _generate_autogeniampolicy(self, resource, template):
-        # type: (models.AutoGenIAMPolicy, Dict[str, Any]) -> None
-        pass
-
-    def _generate_deploymentpackage(self, resource, template):
-        # type: (models.DeploymentPackage, Dict[str, Any]) -> None
-        pass
-
-    def _generate_precreatediamrole(self, resource, template):
-        # type: (models.PreCreatedIAMRole, Dict[str, Any]) -> None
-        pass
-
     def _generate_s3bucketnotification(self, resource, template):
         # type: (models.S3BucketNotification, Dict[str, Any]) -> None
         message = (
@@ -523,10 +560,6 @@ class SAMTemplateGenerator(object):
             }
         }
 
-    def _default(self, resource, template):
-        # type: (models.Model, Dict[str, Any]) -> None
-        raise NotImplementedError(resource)
-
     def _register_cfn_resource_name(self, name):
         # type: (str) -> str
         cfn_name = to_cfn_resource_name(name)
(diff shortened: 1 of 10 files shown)


kapilt:
Also worth noting.. This targets terraform versions under < 0.12 at the moment. Terraform just had a major release (0.12) last week which has some significant incompatibilities with previous versions.

https://www.terraform.io/upgrade-guides/0-12.html

kapilt:
fwiw, tested against 0.12 seems to work fine, i've removed that restriction in the generated output.

to elaborate on some of the converged/combined provisioning capability, you can create a dynamodb table, sqs queue in terraform, and then direct reference them as env vars for the chalice app.. an example from a sample app i'm using (config.json). 

```
        "environment_variables": {
            "TABLE_GITTER": "${aws_dynamodb_table.gitter.name}",
            "GITTER_TOKEN_PARAM": "${aws_ssm_parameter.gitter_token.name}",
            "ELASTICSEARCH_HOST": "${aws_elasticsearch_domain.gitter.endpoint}"
        }
```

at least for me this removes the need for post processing scripts, or splitting the app into multi stack deployments, etc.  the one item i've been considering for enabling better/forward backward navigation of references is capturing the app name / chalice stage name as terraform variables for easier referencing, the extant generator interface doesn't allow for that at the moment though, so it will require some additional refactoring, perhaps as a followup.

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,128 @@
...
+
+When you run ``chalice deploy``, chalice will deploy your application using the
+`AWS SDK for Python <http://boto3.readthedocs.io/en/docs/>`__).  Chalice also
+provides functionality that allows you to manage deployments yourself using
+terraform.  This is provided via the ``chalice package --pkg-format terraform``
+command.
+
+When you run this command, chalice will generate the AWS Lambda deployment
+package that contains your application as well as a `Terraform <https://terraform.io>`__
+template.  You can then use a terraform to deploy your chalice application.
```
stealthycoin: Extra space and a stray `a`

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,128 @@
...
+
+In this example, we'll create a chalice app and deploy it using
+the AWS CLI.
+
+First install the necessary packages::
+
+    $ virtualenv /tmp/venv
+    $ . /tmp/venv/bin/activate
+    $ pip install chalice awscli
+    $ chalice new-project test-cfn-deploy
```
stealthycoin: `test-tf-deploy` and below as well.

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,128 @@
...
+
+At this point we've installed chalice and the AWS CLI and we have
+a basic app created locally.  Next we'll run the ``package`` command
+and look at its contents::
+
+    $ $ chalice package --pkg-format terraform /tmp/packaged-app/
+    Creating deployment package.
+    $ ls -la /tmp/packaged-app/
+    -rw-r--r--   1 j         wheel  3355270 May 25 14:20 deployment.zip
+    -rw-r--r--   1 j         wheel     3068 May 25 14:20 sam.json
```
stealthycoin: `chalice.tf.json`

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,128 @@
+Terraform Support
```
stealthycoin: We should probably specify what versions of Terraform are supportted here.

stealthycoin:
This would be really cool to add. What do you think about integration tests for terraform?

I will have to go through this in more detail since I'm not very familiar with terraform so I gave it a quick once over. But Ill have to block off some time to play around with it.

I wonder as we add more backends if we need to add some documentation matrix of features to backend support.

docs/source/topics/tf.rst
kapilt: at the moment support is pretty general, i've got a restriction in the generated output that says > 0.11.0 just for sanity which is basically any release in the last 18 months. i've done most of my testing with 0.11, but currently trying out 0.12.

kapilt:
I'm a little unfamiliar with the integration tests in chalice, and how one actually executes them or what the ci harness is for them to run, i've mostly been using `make prcheck` around test running.  i'd be game for one, or even some additional tests for this on the generated output, but terraform itself is pretty stable in its usage. Is there any docs on how to run the integration tests?

Wrt to additional backends, i'm not sure if there's others that are meaningful to support (maybe cdk?), terraform integration covers off on a wide variety of use cases that cfn doesn't handle by itself well. there's definitely costs to each backend, as there will be differential feature support by backend, and new features need to be added to all of them. Your websocket branch is an example of one such, as terraform doesnt yet have support for that.

One note, i do suspect it would be possible to enable automatic transition from deploy/destroy to terraform, just not sure its worth the effort, would require a separate script to transition the client side state from one to the other.

docs/source/topics/tf.rst
kapilt: i've done enough apps/deploys with 0.12 to feel confident there as well. given terraform versions tends to more recent, i don't think we need an explicit doc here, the constraint in the generated output solves for accidental usage of truly ancient versions. if you feel differently let me know.

haduart:
@kapilt Is this already merged? If I don't understand it wrong, with this terraform support feature, aws chalice should be able to not only generate a cloudformation template but also terraform scripts? is this correct?

kapilt:
@haduart its not merged yet, pip has solid support for installing from a branch, example below. yes this branch effectively supports provisioning/deploying a chalice app with just terraform

```
$ virtualenv xyz
$ ./xyz/bin/pip install -e git+https://github.com/kapilt/chalice.git@terraform-package#egg=chalice
$ ./xyz/bin/chalice start-project mydev
$ cd mydev
... do some chalice dev

# now deploy with terraform

$ ./xyz/bin/chalice package --pkg-format terraform tfout
$ cd tfout
$ terraform init
$ terraform apply
```

There's a readme with some examples in this pr, but real world uses. Also fwiw, since its taking some time to merge, i've started to maintain an integration  branch with most of my pull requests here integrated ( cloud watch events support ) etc called `next` albeit thats subject to change, but feedback appreciated re kicking the tires. i'm using it for some projects atm.

chalice/package.py
```diff
@@ -357,15 +386,255 @@ def _register_cfn_resource_name(self, name):
...
+
+    def _fref(self, lambda_function, attr='arn'):
+        # type: (models.ManagedModel, str) -> str
+        return '${aws_lambda_function.%s.%s}' % (
+            lambda_function.resource_name, attr)
+
+    def _arnref(self, arn_template, **kw):
+        # type: (str, str) -> str
+        d = dict(
+            region='${aws_region.chalice.name}',
```
kyleknap: For these arn ref's, I think we need to use `data.aws_region` etc. As is terraform complains that `aws_region` and `aws_caller_identity` are not in the root module.

chalice/package.py
```diff
@@ -357,15 +386,255 @@ def _register_cfn_resource_name(self, name):
...
+        if resource.suffix:
+            bnotify['filter_suffix'] = resource.suffix
+
+        template['resource'].setdefault('aws_s3_bucket_notification', {})[
+            resource.resource_name] = {
+                'bucket': resource.bucket,
+                'lambda_function': bnotify
+        }
+
+        template['resource'].setdefault('aws_lambda_permission', {})[
```
kyleknap: So ran into [this gotcha](https://www.terraform.io/docs/providers/aws/r/s3_bucket_notification.html) when converting my app. Specifically we can only one bucket notification resource per bucket and in my app, I have two lambda function subscribing to the same bucket. I think to support this we will have to consolidate all lambda functions to the same bucket notification if they point to the same bucket. Otherwise, you get in this weird state where only one of the notification configurations will get applied.

chalice/package.py
```diff
@@ -357,15 +386,255 @@ def _register_cfn_resource_name(self, name):
...
+class TerraformGenerator(TemplateGenerator):
+
+    template_file = "chalice.tf.json"
+
+    def generate(self, resources):
+        # type: (List[models.Model]) -> Dict[str, Any]
+        template = {
+            'resource': {},
+            'terraform': {
+                'required_version': '> 0.11.0'
```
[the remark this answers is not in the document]
kyleknap: Yeah let's put a minimum pinning on this as well. It seems also worthwhile to put a major version ceiling on the provider as well. It looks like they are coming out with a major version 3 and we should be careful that it does not break the template we generate.

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,129 @@
...
+  Apply complete! Resources: 14 added, 0 changed, 0 destroyed.
+
+  Outputs:
+
+  EndpointURL = https://7bnxriulj5.execute-api.us-east-1.amazonaws.com/dev
+
+This will take a minute to complete, but once it's done, the endpoint url
+will be available as an output which we can then curl::
+
+    $ http "https://7bnxriulj5.execute-api.us-east-1.amazonaws.com/dev"
```
kyleknap: To make the tutorial more programmatic, it may make sense to use `terraform output` to get the url and then run HTTPie from there. So:
```
 $ http "$(terraform output EndpointURL)"
```

chalice/package.py
kyleknap: Also what's your thoughts on setting the minimum version to the latest major version of troposphere `0.12`?

chalice/package.py
```diff
@@ -357,15 +386,255 @@ def _register_cfn_resource_name(self, name):
...
+
+    def _generate_restapi(self, resource, template):
+        # type: (models.RestAPI, Dict[str, Any]) -> None
+
+        # typechecker happiness
+        swagger_doc = cast(Dict, resource.swagger_doc)
+        template['resource'].setdefault('aws_api_gateway_rest_api', {})[
+            resource.resource_name] = {
+                'body': json.dumps(swagger_doc),
+                # Terraform will diff explicitly configured attributes
```
kyleknap: That's unfortunate. We'll probably have to set these for minimum compression, endpoint type, and policy as well.

chalice/package.py
```diff
@@ -357,15 +386,255 @@ def _register_cfn_resource_name(self, name):
...
+                resource.resource_name] = {
+                    'rule': '${aws_cloudwatch_event_rule.%s.name}' % (
+                        resource.resource_name),
+                    'target_id': resource.resource_name,
+                    'arn': self._fref(resource.lambda_function)
+        }
+        template['resource'].setdefault(
+            'aws_lambda_permission', {})[
+                resource.resource_name] = {
+                    'function_name': self._fref(resource.lambda_function),
```
kyleknap: So I noticed this in the generate terraform template, but for the function name we actually end up referencing the `arn`. Using the arn seemed to be fine, but may be worth double checking if we should be using the name instead.

chalice/package.py
kapilt: thats an inherent limitation of s3 (no overlapped notifications by prefix), not a tooling issue. frankly for any real world usage, i would expect the notifications to come in via sns or trail data events, bucket notify direct to lambda/sqs is a neat trick but its single purpose/usage.

chalice/package.py
kapilt: i'm notionally fine with it, but this integration was tested with 0.11 and works seamlessly across that boundary. historically i witnessed many versions of tf extant in a single org, so aiming for broader reach where it is known compatible.

chalice/package.py
kapilt: yes, all my half-dozen pull requests conflict based on the way the project is structured and will need updates in response to any trunk changes. thats part of the reason contributor pull requests should be prioritized instead of allowed to languish imo. i'll update to this supports whats currently merged, else its a responsibility of feature contributors and maintainers to ensure parity.

chalice/package.py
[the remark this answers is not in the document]
kapilt: there's nothing material to chalice usage in the version 3 release notes (https://www.terraform.io/docs/providers/aws/guides/version-3-upgrade.html) . ceiling pins cause more problems for end users in my experience then benefits unless their done with known cause. terraform is commonly shared ci platform tool install, so this can be quite tedious for end users since they may not control the upgrade schedule or ability to downgrade do to ceiling pins. doing a minimum version pin for what chalice generated output actually needs is good. afaics for the aws provider anything over > 2.0 should be sufficient.

chalice/package.py
kapilt: I think I misunderstood, are both of those lambda functions in the chalice app?  I can add support for that.

chalice/package.py
kapilt: I switched this to name but might  switch back, its actually bit easier on testing with the arn ref as there isn't really a name for the app setup in most tests, so it comes out oddly when doing assertions as while the sample apps have a name the configs for them don't so assertions come out as comparisons to None-function-name

chalice/package.py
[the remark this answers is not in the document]
kapilt: added in a aws provider > 2.0 here.

chalice/package.py
kapilt: added in support for doing multiple function in the same chalice app.

docs/source/topics/tf.rst
```diff
@@ -0,0 +1,129 @@
...
+          283  05-25-17 14:19   chalice/__init__.py
+          796  05-25-17 14:20   app.py
+     --------                   -------
+      9826899                   723 files
+
+    $ head < /tmp/packaged-app/chalice.tf.json
+    {
+      "resource": {
+        "aws_lambda_function": {
+           "function_name": "radical-dev-sfn_origin",
```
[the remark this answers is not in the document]
kapilt: removed this bit.

chalice/package.py
kapilt: left with function name unless parameter name said arn, and updated said tests to generate config with chalice app name for clean comparisons.

chalice/package.py
kapilt: to be honest given the shared ci situation at a lot of large orgs and that has been tested with 0.11 I'd rather leave as is, the 0.12 shift was fairly large but its also fairly recent, and redress start of q4.

kyleknap: Sounds good. We can keep it at `0.11` then.

chalice/package.py
kyleknap: >> Are both of those lambda functions in the chalice app?

Yep. I ran into that when porting this app: https://github.com/kyleknap/chalice-workshop/commit/73bcfecf24bebcbdda910a6e9ff6f78fe0cc95ec. I had one lambda function that ran on uploads and another that ran on deletes. The update looks good.

chalice/package.py
```diff
@@ -530,15 +559,272 @@ def _register_cfn_resource_name(self, name):
...
+            template['resource'][
+                'aws_api_gateway_rest_api'][
+                    resource.resource_name]['policy'] = swagger_doc[
+                        'x-amazon-apigateway-policy']
+        if resource.minimum_compression.isdigit():
+            template['resource'][
+                'aws_api_gateway_rest_api'][
+                    resource.resource_name][
+                        'minimum_compression_size'] = (
+                            resource.minimum_compression)
```
kapilt: pretty sure this needs to an int conversion for terraform

chalice/package.py
```diff
@@ -530,15 +559,272 @@ def _register_cfn_resource_name(self, name):
...
+                resource.role.resource_name)
+        else:
+            # resource is a PreCreatedIAMRole.
+            role = cast(models.PreCreatedIAMRole, resource.role)
+            func_definition['role'] = role.role_arn
+
+        template['resource'].setdefault('aws_lambda_function', {})[
+            resource.resource_name] = func_definition
+
+    def _generate_restapi(self, resource, template):
```
kyleknap: Looking over this again. I just remembered... Did terraform add support for websockets? Right now if you try this on a websocket app you get this traceback:
```
$ chalice package --pkg-format=terraform terraform
Creating deployment package.
  File "/Users/kyleknap/GitHub/chalice/chalice/cli/__init__.py", line 468, in main
    return cli(obj={})
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/core.py", line 722, in __call__
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/core.py", line 895, in invoke
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/core.py", line 535, in invoke
  File "/Users/kyleknap/.pyenv/versions/3.6.1/envs/oscon/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
  File "/Users/kyleknap/GitHub/chalice/chalice/cli/__init__.py", line 397, in package
    packager.package_app(config, out, stage)
  File "/Users/kyleknap/GitHub/chalice/chalice/package.py", line 842, in package_app
    template = self._templater.generate(resources)
  File "/Users/kyleknap/GitHub/chalice/chalice/package.py", line 582, in generate
    self.dispatch(resource, template)
  File "/Users/kyleknap/GitHub/chalice/chalice/package.py", line 85, in dispatch
    handler(resource, template)
  File "/Users/kyleknap/GitHub/chalice/chalice/package.py", line 109, in _default
    raise UnsupportedFeatureError(resource)
chalice.package.UnsupportedFeatureError: WebsocketAPI(resource_name='websocket_api', name='websockets-dev-websocket-api', api_gateway_stage='api', routes=['$default'], connect_function=None, message_function=LambdaFunction(resource_name='websocket_message', function_name='websockets-dev-websocket_message', deployment_package=DeploymentPackage(filename='/Users/kyleknap/GitHub/chalice/sampleapps/websockets/.chalice/deployments/c28cddae9295a6d0d8b267751df45cd9-python3.6.zip'), environment_variables={}, runtime='python3.6', handler='app.message', tags={'aws-chalice': 'version=1.9.1:stage=dev:app=websockets'}, timeout=60, memory_size=128, role=ManagedIAMRole(resource_name='default-role', role_name='websockets-dev', trust_policy={'Version': '2012-10-17', 'Statement': [{'Sid': '', 'Effect': 'Allow', 'Principal': {'Service': 'lambda.amazonaws.com'}, 'Action': 'sts:AssumeRole'}]}, policy=AutoGenIAMPolicy(document={'Version': '2012-10-17', 'Statement': [{'Effect': 'Allow', 'Action': ['logs:CreateLogGroup', 'logs:CreateLogStream', 'logs:PutLogEvents'], 'Resource': 'arn:aws:logs:*:*:*'}, {'Effect': 'Allow', 'Action': ['execute-api:ManageConnections'], 'Resource': 'arn:aws:execute-api:*:*:*/@connections/*'}]}, traits=set())), security_group_ids=[], subnet_ids=[], reserved_concurrency=None, layers=[]), disconnect_function=None)
```
It might be work explicitly throwing an error for this in order to give a friendlier error like we do for cloudformation generation of lambda functions that have s3 events.

chalice/package.py
```diff
@@ -530,15 +559,272 @@ def _register_cfn_resource_name(self, name):
...
+        d = dict(
+            region='${data.aws_region.chalice.name}',
+            account_id='${data.aws_caller_identity.chalice.account_id}')
+        d.update(kw)
+        return arn_template % d
+
+    def _generate_managediamrole(self, resource, template):
+        # type: (models.ManagedIAMRole, Dict[str, Any]) -> None
+        template['resource'].setdefault('aws_iam_role', {})[
+            resource.resource_name] = {
```
kyleknap: Just making a general comment for future development. I went through the resource names we are generating. They look fine too me. The one thing that we should address (but do not need to now) is add tests to make sure that we do not accidentally change `resource_name`s and have that cause a change in the resource name in the generated template. This is especially true for terraform as people will be able to reference resources from the generated chalice template to other templates. The backwards compatibility guarantee will be needed for this PR as well: https://github.com/aws/chalice/pull/1195

chalice/package.py
```diff
@@ -530,15 +559,272 @@ def _register_cfn_resource_name(self, name):
...
+        if resource.prefix:
+            bnotify['filter_prefix'] = resource.prefix
+        if resource.suffix:
+            bnotify['filter_suffix'] = resource.suffix
+
+        # we use the bucket name here because we need to aggregate
+        # all the notifications subscribers for a bucket.
+        template['resource'].setdefault(
+            'aws_s3_bucket_notification', {}).setdefault(
+                resource.bucket + '_notify',
```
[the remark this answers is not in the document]
kapilt: that's odd the resource name for the notification I'm noticing is a literal string in the terraform output of the bucket_name+_notify (ie. foo_notify). are you using a terraform variable for the bucket name in your chalice app s3 event decorator? the CloudFormation uniqueness is actually not whats not needed here, instead its a grouping by bucket name.

chalice/package.py
kapilt: web sockets support is still pending for terraform split across a half-dozen+ pull requests 
https://github.com/terraform-providers/terraform-provider-aws/issues/7004

noted re specific error message here.

chalice/package.py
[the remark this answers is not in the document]
kapilt: nevermind re decorator value, i see it the example config.json, as an env variable terraform ref. I'll need to detect and parse the terraform ref in bucket names for the resource name and use that to group the bucket notifications without variable syntax which is not supported for a resource name.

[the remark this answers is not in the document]
kyleknap: Yep that was what I was referring to. But that approach sounds good to me.

chalice/package.py
```diff
@@ -625,9 +625,15 @@ def _generate_s3bucketnotification(self, resource, template):
 
         # we use the bucket name here because we need to aggregate
         # all the notifications subscribers for a bucket.
+        # Due to cyclic references to buckets created in terraform
+        # we also try to detect and resolve.
+        if '{aws_s3_bucket.' in resource.bucket:
```
kyleknap: I think that having this quick check here is fine for now. But I have a feeling we will have to come up with a more general way to handle terraform references for resource names. Fortunately, we can rely on the name of the function for a lot of these terraform resources instead of the AWS resource the function is hooked up to, but people in the future may want to override the name of the lambda function (i.e. with the `name` parameter) with a terraform reference.

chalice/package.py
```diff
@@ -539,15 +566,290 @@ def _register_cfn_resource_name(self, name):
...
+                resource.role.resource_name)
+        else:
+            # resource is a PreCreatedIAMRole.
+            role = cast(models.PreCreatedIAMRole, resource.role)
+            func_definition['role'] = role.role_arn
+
+        template['resource'].setdefault('aws_lambda_function', {})[
+            resource.resource_name] = func_definition
+
+    def _generate_restapi(self, resource, template):
```
kyleknap: So I'm not sure what the issue is but I'm running into issues when redeploying. Specifically, if I start with the following `app.py`:
```python
from chalice import Chalice, Rate

app = Chalice(app_name='tf')
app.debug = True


@app.route('/')
def index():
    return {'hello': 'world'}
```
I then deployed it using:
```
$ chalice package --pkg-format=terraform out
$ cd out
$ terraform apply
```
And made a request to it and it worked:
```
$ http $(terraform output EndpointURL)
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 17
Date: Sun, 04 Aug 2019 20:07:50 GMT
Via: 1.1 3cc149abcbd8cb50553c09666247bcfa.cloudfront.net (CloudFront)
X-Amz-Cf-Id: xaIvOdgvZbyNX43qQaZ2RHebbWASYpyO_gbgMkeKppBxufo6x_ZWjg==
X-Amz-Cf-Pop: HIO51-C1
X-Amzn-Trace-Id: Root=1-5d473b15-672b38e5fe6ae2f0e7875408;Sampled=0
X-Cache: Miss from cloudfront
x-amz-apigw-id: d6YrZGNOPHcFldQ=
x-amzn-RequestId: 88a2a7ab-b6f3-11e9-aa00-f7175e1fd48e

{
    "hello": "world"
}
```

Then I updated the `app.py` to and redeployed it:
```python
from chalice import Chalice, Rate

app = Chalice(app_name='tf')
app.debug = True


@app.route('/', methods=['POST'])
def index():
    return {'hello': 'world'}
```
And then I got a missing authentication token when I redeployed:
```
$ http POST $(terraform output EndpointURL)
HTTP/1.1 403 Forbidden
Connection: keep-alive
Content-Length: 42
Date: Sun, 04 Aug 2019 20:13:06 GMT
Via: 1.1 bb14b7a4f45e3e0b1ae5065c59f7826f.cloudfront.net (CloudFront)
X-Amz-Cf-Id: nxQo7BEVTn-D2Qff9UuTf3AmnwTUgnwJshmliHAgn3K5yD7yQNaQxQ==
X-Amz-Cf-Pop: HIO51-C1
X-Cache: Error from cloudfront
x-amz-apigw-id: d6Zc5GclPHcFUSA=
x-amzn-ErrorType: MissingAuthenticationTokenException
x-amzn-RequestId: 4576a6d9-b6f4-11e9-859c-8361244f785a

{
    "message": "Missing Authentication Token"
}
```

The weird part is that if I then run `terraform destroy` and then `terraform apply`, it then works:
```
$ http POST $(terraform output EndpointURL)
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 17
Date: Sun, 04 Aug 2019 20:16:55 GMT
Via: 1.1 7e81687e34febf53ccb2929005d5ef12.cloudfront.net (CloudFront)
X-Amz-Cf-Id: zDynWia8aB3I1Fw0bYcD8BUdzfqgd8HucfZEIuWil-oPGxUe_01JsA==
X-Amz-Cf-Pop: HIO51-C1
X-Amzn-Trace-Id: Root=1-5d473d37-3831e3d2e82333a598e64473;Sampled=0
X-Cache: Miss from cloudfront
x-amz-apigw-id: d6aAmEiqPHcFazQ=
x-amzn-RequestId: cdae530f-b6f4-11e9-8e63-8f62322c2646

{
    "hello": "world"
}
```
Have you run into this issue before? I want to dive into why the API is not being setup correctly. I suspect it is a terraform bug though given it works after tearing it down and redeploying it.

kapilt:
re 3. squashing commits should be done ideally on merge within the GitHub ui, and can be setup as mandatory for pull requests merge on GitHub project settings. The chalice project has not historically done this to date. Doing it on pull requests is typically considered an anti-pattern by reviewers once they've commented/reviewed as they have to reconsider the whole diff on changes as to opposed post review changes.

currently looking into 1 & 2

chalice/package.py
kapilt: reproduced, the issue it appears is that terraform needs to create a new deployment on the method changes and doesn't because none of the deployment's attributes changed.

chalice/package.py
kapilt: this is basically the only place we use a user supplied var name in the resource, as we're using the user input for the name instead of the chalice resource name for it (aka python function name) to enable bucket level grouping here for the multiple lambda functions on a bucket scenario you had posited previously, else we're using the chalice resource name for all other resources.

chalice/package.py
kapilt: resolved via removing explicit stage creation, and implicitly creating one via deployment, and forcing redeployments on swagger changes by linking in the checksum of the swagger.

chalice/package.py
```diff
@@ -539,15 +566,290 @@ def _register_cfn_resource_name(self, name):
...
+        }
+        template['resource'].setdefault('aws_lambda_permission', {})[
+            resource.resource_name] = {
+                'function_name': resource.lambda_function.function_name,
+                'action': 'lambda:InvokeFunction',
+                'principal': 'sns.amazonaws.com',
+                'source_arn': topic_arn
+        }
+
+    def _generate_scheduledevent(self, resource, template):
```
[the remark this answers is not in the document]
kapilt: while its good to note in the pr, this should be filed as a new issue on the repo so it can be tracked for future work. at least for terraform we could indeed be outputing the deployed.json resources with these resources, and technically that may allow some degree of bidirectional compatibility for deploy/package w/ terraform.

chalice/package.py
kapilt: wrt to name parameter, is there a scenario where that would include a terraform reference ?

kyleknap: That is what I'm trying to think if it it would make sense ever to include a terraform reference in the name. So far I have not come up with any use case.

kapilt:
I’ve been noodling again if we should be outputting the stage name as a variable here, so that other resources can use it as a prefix when provisioning to have isolated stacks. It effectively it would become a terraform variable, but we’d have to note that it can’t be user supplied via terraform in the chalice docs though.

kyleknap:
Could you elaborate or give an example on how the stage variable would look in the generated terraform template?

kyleknap:
Otherwise. I don't think I have any more feedback on the code. Confirmed that the updates resolve the two issues I ran into.

chalice/package.py
```diff
@@ -798,17 +803,11 @@ def _generate_restapi(self, resource, template):
...
-                    resource.resource_name),
-                'stage_name': resource.api_gateway_stage,
-                'deployment_id': '${aws_api_gateway_deployment.%s.id}' % (
-                    resource.resource_name)
-        }
-
         template['resource'].setdefault('aws_api_gateway_deployment', {})[
             resource.resource_name] = {
+                'stage_name': resource.api_gateway_stage,
+                'stage_description': (
```
kyleknap: Actually it might be worth to put a comment on why we are adding the md5 to the stage description.

kapilt:
re stage-name, so when provisioning physical assets with terraform its useful to prefix them with app-name, stage-name, so that multiple chalice stages can be deployed in the same account.

> Could you elaborate or give an example on how the stage variable would look in the generated terraform template?

as an example in one of my terraform'd chalice apps i have a kinesis delivery stream (non relevant parts omitted), atm its currently using static names, the change would be to introduce chalice stage and app name variables so they can be referenced when constructing physical asset names as below.

```terraform
resource "aws_kinesis_firehose_delivery_stream" "gitter_index" {
  name        = "${var.chalice_app_name}-{var.chalice_stage_name}-gitter-s3-archive"
  destination = "elasticsearch"

  # omit various non germane configuration
}
```
To support this we would add terraform variables with default values to the terraform module.

Generators don't have access to the app name or chalice stage name atm (though they can infer from rest api stage name or lambda info tags). The clean way of doing this would be updating generator class constructor signature to also take a chalice Config.

The caveat is that these chalice prefix'd variables can't be user defined via the cli, because other chalice generate assets embeds the literal values of those variables into output'd resources. The description for the variables and the chalice documentation would both note this restriction.

--

One other item in looking at the chalice package cli, it feels like this needs a bit more help text.


```shell
$ chalice package --help
Usage: chalice package [OPTIONS] OUT

Options:
  --single-file            Create a single packaged file. By default,
                                  the 'out' argument specifies a directory in
                                  which the package assets will be placed.  If
                                  this argument is specified, a single zip
                                  file will be created instead.
  --stage TEXT
  --pkg-format [cloudformation|terraform]
  --merge-template TEXT           Specify a JSON template to be merged into
                                  the generated template. This is useful for
                                  adding resources to a Chalice template or
                                  modify values in the template.
  --help                          Show this message and exit.
```

ie pkg-format `Specify Provisioning Engine to be used for generated template output`

and also note for --single-file and --merge-template options that they only apply to cloudformation engine.

chalice/package.py
kapilt: done.

kapilt:
i've gone ahead and added in terraform variables for chalice app_name and stage_name, and updated the cli help.

chalice/package.py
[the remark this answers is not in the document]
kapilt: given the other pull requests in the queue, and the nature of the code structure creating conflicts, I went ahead and started on this to have the terraform backend be feature complete, so I can context switch to the other features, as incremental on this would naturally conflict with the other queue items.

kapilt:
cli compatibility with via local provisioning of deployed_resources.json stub to support subcommands logs/invoke/url added.

chalice/package.py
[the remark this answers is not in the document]
jamesls: The concern is that we don't know that for certain until we can test against version 3 (my read on the linked docs suggest the full set of changes aren't finalized yet).  And, in theory this applies to *any* future major version of this provider going forward.

While I understand version caps can be difficult, I feel that capping to an entire major version seems like a reasonable tradeoff (vs. say a minor version pin).  The expectation is that major versions will have breaking changes, so I think waiting until we can confirm we're not affected seems like a reasonable approach.

I would say this applies to anywhere we're specifying versions.

jamesls:
Nice!  I've been playing around with this.  I noticed that the `deployed/dev.json` changes are causing issues for me.  I'd like to do more testing on that before we support writing out that json file.  What do you think about holding off on this for now?  We can always add this later given there's no backwards compatibility issues, and I think this PR without the `deployed/dev.json` changes are already useful as is, and I'd like to get an initial version of this merged.

kyleknap:
@kapilt Thanks for the updates. I had a some comments/ideas about the updates:

1) I agree with James we should hold off on the `deployed.json` in this PR so we can get this merged. I think for me, I'd like to see if we can refactor the code so the chalice deployer and the chalice packager can leverage the same abstractions so we can ensure the `deployed.json` file will be close to the same for the chalice deployer backend and the terraform backend.

1) I like adding the idea of exposing the stage name and app name in the Terraform template. While the `package` command does not expose the stage name and app name for CloudFormation, I think it makes sense for terraform as users are put in a situation where they need to the name the resources themselves (which is typically handled for your behalf) and the app name along with stage name will help preserve uniqueness if they have multiple stages. But I'm wondering if this is another addition that we should consider holding off for in **this** PR and do it in a follow up PR? It would be backwards compatible to add the variables in the future and it seems weird right now to have variables that act more like outputs and do not actually affect the generation of the resources in the template. 

1) I think we should be having CLI runtime validations for the  `--merge-template` and `--single-file` parameters when the `--pkg-format` is set to `terraform`. Having it in the help documentation is nice, but it really should raise an error to better help the user when they did not thoroughly read the documentation.

In general, I think it would be nice to address the variable and `deployed.json` changes in a subsequent PRs as an incremental updates. Even before those two updates, we were really close to getting this merged and into a release, and I'd like to see that happen sooner rather than later especially since those two updates would not introduce any backwards compatible changes.

kapilt:
1. happy to back out the deployed/$stage.json, will file as a separate pull request once this pr gets merged, although curious what issues were seen.
2. exposing of the app and stage names is important imo, and there isn't any other way to do this. they are not weird imo, they are used in the exact way variables in terraform are intended to be use, ie for controlling resource generation, the reality is that anyone using the terraform is likely provisioning the rest of their app (buckets, DynamoDB, sqs, sns) alongside. its only backwards compatible if we ignore that everyone who needs this has to rewrite their terraform config if this feature is added later.
3. sounds good re validation.

kyleknap:
Just replying back...

1) I did not try it out but just skimming over the code for what would have been generated there were a couple of things I noticed: It made the assumption that there was always a `route()` in the app so if you had a chalice app with a single `lambda_function`, my guess is that would fail. Also, I believe we capture more than just lambda functions and the api gateway resources in the `deployed.json` such as iam roles and event sources, and while it is not necessarily needed for the `logs`, `url`, and `invoke` commands, it would be nice to output relatively the same file in case we add a command that needs the value of these deployed resources.

2) I guess it may be weird to me because I don't have too much experience with Terraform. I guess is it an accepted practice in Terraform to have read-only input variables, that are really only used for referencing? I looked more into the available functionality such as data sources, local values, or even if there was a resource available for it, but have not really found anything that fit the bill for being a grab bag of read-only data. I'll defer to you as you have more experience with Terraform, but I want to make sure that we explored our options here (if there are any other).

kapilt:
1. There is no compatibility intended for deploy commands and package, so its unclear what value/intent rendering additional bits here serve, the engine attribute serves to denote which system set those values, and ideally should be used as guards. This was intended to preserve the minimal semantics needed to preserve cli compatibility, doing additional is a non goal feature because its non functional/useful. Per your initial comment, Trying to reuse logic from planner around recorded values, breaks encapsulation of features, and frankly is a large refactoring for minimal value chain, given this already establishes the feature (cli compatibility) for the least amount of code, the packager is already operating on deployer models which is the best degree of reuse out of this imo.

2. its very common in both cloudformation and terraform to have parameters/variables with default values that don't need to be set. terraform has a secondary notion of here dynamically sourced data values, i think i see a one option here that removes the ambiguity of variables that can be set by using static data values from a null data provider (https://www.terraform.io/docs/providers/null/data_source.html) . the ergonomics feel worse then using a variable imo, ${var.chalice_app_name} vs ${data.null_data_source.chalice.outputs.app}  but it removes the ambiguity. i can make the change if that feels preferable.

kyleknap:
Despite the longer name, I think I'd prefer the null data source provider here to remove the ambiguity.

kapilt:
switched to null data provider, i'm still of the opinion that variables are better for this usage. the documentation on the null provider reinforces that perspective https://www.terraform.io/docs/providers/null/index.html

i think there's option here for future improvements to make app and stage actual variables by unwinding all the chalice app-stage prefixing in the deployer when packaging (aka regenerating resource names), but that can be done in a backwards compatible way to this null data provider.

also travis seems to be experiencing issues today, i've had numerous build failures do to network issues https://www.traviscistatus.com/incidents/hl4vqb7hvv5n

kyleknap:
>> i think there's option here for future improvements to make app and stage actual variables by unwinding all the chalice app-stage prefixing in the deployer when packaging (aka regenerating resource names), but that can be done in a backwards compatible way to this null data provider.

Yep. I agree.

Looking over the diff. I think the only outstanding point was removing the deployed state file generation for now.

kapilt:
> Looking over the diff. I think the only outstanding point was removing the deployed state file generation for now.

hmm.. sorry thought i had gotten that removed, switching laptops when force pushing is leading to some oddities where it was reintroduced.  just pushed the removal of that again.

chalice/package.py
```diff
@@ -585,8 +585,6 @@ def generate(self, resources):
             },
             'provider': {
                 'aws': {'version': '> 2.0.0'},
-                'local': {'version': '> 1.0'},
-                'template': {'version': '> 2.0'},
```
kyleknap: Hmm. We wanted to keep the `template` version floor constraint right?

kyleknap: Also on a related topic @jamesls wanted to circle back about having a major version ceiling for these providers. He added this comment: https://github.com/aws/chalice/pull/1129#discussion_r310848785 and not sure if you saw it?

kapilt: updated with ceiling

kyleknap: Awesome. Thanks!

jamesls:
Cool, I'll go ahead and squash/rebase and merge.

PR #1149, opened 2019-06-11
avoid downloading related dependencies

robertobarreda:
Description of changes:
Ensure generated deploy package doesn't contain unlisted related dependencies.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -725,7 +725,9 @@ class PipRunner(object):
     def download_all_dependencies(self, requirements_filename, directory):
         # type: (str, str) -> None
         """Download all dependencies as sdist or wheel."""
-        arguments = ['--no-deps', '-r', requirements_filename, '--dest', directory]
+        arguments = [
+            '--no-deps', '-r', requirements_filename, '--dest', directory
+        ]
         rc, out, err = self._execute('download', arguments)
         # When downloading all dependencies we expect to get an rc of 0 back
         # since we are casting a wide net here letting pip have options about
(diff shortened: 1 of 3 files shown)


stealthycoin:
Can you provide some details on what you are trying to accomplish? Merging this would defeat the purpose of the packager, which is to collect your dependencies so your bundle will work properly.

robertobarreda:
I was trying to pack an application with pandas and scikit-learn dependencies.

This two has as dependencies numpy and scipy.

The only way to build this app with the lambda size restrictions is by using an external layer.

Without the modification, pip tries to install all related dependencies.

kapilt:
wouldn't it be easier to just put the layer dependencies in a different requirements file?

robertobarreda:
The problem is the moment you try to install either `pandas` or `scikit-learn`
`pip` will also install its dependencies: `numpy` & `scipy`.

```
$ pipenv graph

pandas==0.24.2
  - numpy [required: >=1.12.0, installed: 1.15.4]
  - python-dateutil [required: >=2.5.0, installed: 2.8.0]
    - six [required: >=1.5, installed: 1.12.0]
  - pytz [required: >=2011k, installed: 2019.1]
scikit-learn==0.21.2
  - joblib [required: >=0.11, installed: 0.13.2]
  - numpy [required: >=1.11.0, installed: 1.15.4]
  - scipy [required: >=0.17.0, installed: 1.1.0]
    - numpy [required: >=1.8.2, installed: 1.15.4]
```

The only way to avoid this behavior is passing the flag `--no-deps` or you end up with `numpy` and `scipy` in the same zip file because `pip` doesn't know you already have it in the system (because we use the layer `arn:aws:lambda:us-east-1:668099181075:layer:AWSLambda-Python37-SciPy1x:2`)

kapilt:
the problem is this pattern breaks common usage for other users of chalice. your particular case would be better suited to manually constructing another additional layer with just your app deps not inclusive of scipy and having an empty requirements for your chalice app.

PR #1155, opened 2019-06-26
schedule: Add a description for cloudwatch event rule

haidaraM:
*Description of changes:*
This PR adds support for CloudWatch Event description when creating a schedule:
```python
from chalice import Chalice, Rate

app = Chalice(app_name="helloworld")

# Automatically runs every 5 minutes
@app.schedule(Rate(5, unit=Rate.MINUTES), description="Hello world event")
def periodic_task(event):
    return {"hello": "world"}
```
By the default, the `description` is set to an empty string. 

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,9 +7,11 @@ Next Release (TBD)
 
 * Add support for stage independent lambda configuration
   (`#1162 <https://github.com/aws/chalice/pull/1162>`__)
-
 * Add support for subscribing to CloudWatch Events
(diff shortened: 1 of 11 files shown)


haidaraM:
I'm wondering if I should `description` everywhere instead of `rule_description` ? I have tried to be close to the existing names...
And I just figured it out that SAM doesn't support description on Schedule :-(

haidaraM:
Someone to take a look at this PR 😄 ? Anything to change ?

jamesls:
Yep, sorry for the delay, looking at this now.

PR #1160, opened 2019-07-01
api gateway endpoint configuration support

### LINKED ISSUE #897: Deploying to "private" API Gateway with dedicated VPC endpoint
Hi,

I have created a dedicated VPC endpoint in order to be able to deploy a Chalice micro-service to a private API Gateway instance (meaning only accessible on a particular VPC within AWS), but I am not seeing anything in the document, nor issues which outlines how it is possible or not to specify the API Gateway resource policy (where you normally specify the VPC endpoint). I am following AWS's documentation here: https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html#apigateway-private-api-set-up-resource-policy

Thank you in advance for any pointers or guidance.

### LINKED ISSUE #956: Support for API Gateway endpoint configuration as REGIONAL
Currently it's very convenient that the API Gateway REST API is created as the EDGE endpoint type. This way the CloudFront distribution is handled by API Gateway.

In a more secure configuration, I would like to create the API as a REGIONAL endpoint and manage the CloudFront distribution myself (outside of Chalice) so I can do things like add WAF on the front-end of requests.

### LINKED ISSUE #976: API Gateway Resource Policy
It seems like support for API Gateway configuration is fairly limited. Even if I set a Resource Policy for the associated API Gateway, subsequent Chalice deploys appear to overwrite it with a blank policy.

Are there technical limitations that restrict Chalice's ability to configure the API Gateway? If not, is more integration with API Gateway on the roadmap for Chalice? If it's not a planned improvement, would a contribution adding it be welcome?

Thanks,
Brian

kapilt:
Closes #956
Closes #897 
Closes #976

Support configuring and updating the api gateway endpoint type (EDGE (default), REGIONAL, PRIVATE). 

Support configure api gateway resource policy, with additional provisions for automatic policy construction in case of PRIVATE endpoints if VPC Endpoint is configured (else a manually specified policy is required).

Note for PRIVATE endpoints this does not create a vpc endpoint, which needs to be created out of band. A single vpc endpoint can be used for multiple apis though, so its one time per region config per account.


docs/source/topics/configfile.rst
```diff
@@ -34,13 +34,16 @@ information about chalice stages.
   key will automatically be created for you.  See the examples
   section below for some stage specific configurations.
 
+* ``api_gateway_endpoint_type``  - The endpoint configuration of the
+  deployed API Gateway, can be EDGE, REGIONAL, PRIVATE. Note this
```
stealthycoin: Might be nice to either expand on what those things mean and what the default is. Or maybe link to the docs where it explains the difference.

kapilt: thanks for the review/feedback. I'll update to note the default and link to endpoint type docs.

kapilt:
ci is failing because with the websocket addition, new additions to planner end up going above the module size limit. I don't see many options here beyond disable this check globally for pylint

```
chalice/deploy/planner.py:2:0: C0302: Too many lines in module (1009/1000) (too-many-lines)
```

chalice/config.py
```diff
@@ -223,6 +223,11 @@ def api_gateway_stage(self):
         return self._chain_lookup('api_gateway_stage',
                                   varies_per_chalice_stage=True)
 
+    @property
+    def api_gateway_endpoint_type(self):
+        # type: () -> str
+        return self._chain_lookup('api_gateway_endpoint_type')
```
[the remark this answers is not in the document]
kapilt: Endpoint type is set on api, not stage, hence config here tries to match that.

.pylintrc
```diff
@@ -59,7 +59,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=R0201,W0613,I0021,I0020,C0111,W1618,W1619,R0902,R0903,W0231,W0611,R0913,W0703,C0330,R0204,I0011,R0904,R0205,R1705,R1710,W0403
+disable=R0201,W0613,I0021,I0020,C0111,W1618,W1619,R0902,R0903,W0231,W0611,R0913,W0703,C0330,R0204,I0011,R0904,R0205,R1705,R1710,W0403,C0302
```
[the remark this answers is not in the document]
kapilt: i'm finding post websocket merge that most of my prs are running afoul of this across multiple modules, i'm just not sure the value chain on a check that we end up squelching per file anytime we hit it vs just disabling it.  i'll go with the per file squelch for now.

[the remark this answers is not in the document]
kyleknap: Which modules are they affecting? I think the check is worthwhile to know about. I guess the big problem is that typically it requires a decent refactoring to fix the issue.

chalice/config.py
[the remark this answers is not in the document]
kyleknap: Oh so there is a difference between api gateway stage and chalice stage. With chalice, it's notion of stage is just an entire separate stack of resources. So a single stage of a chalice app can only have at most one api gateway api and that api can only have one api gateway stage. By setting `varies_per_chalice_stage` it allows users to change the api gateway endpoint type per stack of resources they may have. For example, they may have a beta or dev stack/stage that they want to keep the api as a private endpoint, but a prod stack/stage that is a edge/regional endpoint. In general, chalice does not really utilize api gateway stages. I believe you really can only set the name of it.

.pylintrc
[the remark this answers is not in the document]
kapilt: deployer.py and planner.py are flagging in the cwe branch

kapilt:
i went ahead and disabled the PRIVATE endpoint type (with validation) for a followup branch that will expose api policy and configuring private endpoint types.

chalice/config.py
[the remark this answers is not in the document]
kapilt: gotcha, i hadn't realized chalice stages mapped to different api gw, sounds good.

chalice/deploy/planner.py
```diff
@@ -908,6 +916,14 @@ def _plan_restapi(self, resource):
             ] + shared_plan_epilogue
         else:
             deployed = self._remote_state.resource_deployed_values(resource)
+            if deployed.get('endpoint_type',
```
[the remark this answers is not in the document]
kapilt: working through this at the moment, its going to involve doing a couple of redundant api calls (exists calls get, but changing that to return non boolean is non trivial / invasive) which i can leave for a followup related to #1177, so we'll end up doing at least two api calls non conditionally, one redundantly. testing out of the behavior on apigw for trying a no-op patch, else this would mean growing some additional conditional evaluation onto api method.

kapilt:
fwiw, addressed review comments and did some manual verification, change from review are in the second commit.

kapilt:
i went ahead and added in support for private endpoints and automated or manual policy config to this pr.

chalice/deploy/deployer.py
```diff
@@ -459,15 +459,38 @@ def _create_rest_api_model(self,
...
         )
 
+    def _get_default_private_api_policy(self, config):
+        # type: (Config) -> Dict[str, Any]
+        statements = [{
+            "Effect": "Allow",
+            "Principal": "*",
+            "Action": "execute-api:Invoke",
+            "Resource": (
+                "arn:aws:execute-api:{region_name}:"
```
kyleknap: So when playing with this the first time, I hit this traceback:
```
chalice deploy
Creating deployment package.
Updating policy for IAM role: private-api-dev
Updating lambda function: private-api-dev
Creating Rest API
{'api_handler_lambda_arn': u'arn:aws:lambda:us-west-2:934212987125:function:private-api-dev', 'region_name': u'us-west-2'}
arn:aws:apigateway:{region_name}:lambda:path/2015-03-31/functions/{api_handler_lambda_arn}/invocations
{'region_name': u'us-west-2', 'rest_api_id': u'70196qu3b3', 'account_id': u'934212987125'}
{"Version": "2012-10-17", "Statement": [{"Action": "execute-api:Invoke", "Resource": "arn:aws:execute-api:{region_name}:{account_id}:{rest_api_id}/*", "Effect": "Allow", "Condition": {"StringEquals": {"aws:SourceVpce": "vpce-0a06670cbb04c3863"}}, "Principal": "*"}]}
  File "/Users/kyleknap/GitHub/chalice/chalice/cli/__init__.py", line 466, in main
    return cli(obj={})
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/core.py", line 895, in invoke
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/core.py", line 535, in invoke
  File "/Users/kyleknap/.pyenv/versions/2.7.10/envs/oscon2/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
  File "/Users/kyleknap/GitHub/chalice/chalice/cli/__init__.py", line 202, in deploy
    deployed_values = d.deploy(config, chalice_stage_name=stage)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/deployer.py", line 342, in deploy
    return self._deploy(config, chalice_stage_name)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/deployer.py", line 355, in _deploy
    self._executor.execute(plan)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 31, in execute
    self._default_handler)(instruction)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 41, in _do_apicall
    final_kwargs = self._resolve_variables(instruction)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 117, in _resolve_variables
    api_call.params, self.variables)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 141, in resolve_variables
    final[k] = self.resolve_variables(v, variables)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 149, in resolve_variables
    final_list.append(self.resolve_variables(v, variables))
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 141, in resolve_variables
    final[k] = self.resolve_variables(v, variables)
  File "/Users/kyleknap/GitHub/chalice/chalice/deploy/executor.py", line 132, in resolve_variables
    return value.template.format(**v)
KeyError: '"Version"'
```
It looks like what is happening is that python's builtin format is having troubles doing string format for a json string. I wonder if it makes sense to just not worry about the exact account and region given that these policies are already attached to a specific APIG API? Right now I have this policy on my APIG API and it is working just fine:
```
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": "*",
            "Action": "execute-api:Invoke",
            "Resource": "arn:aws:execute-api:*",
            "Condition": {
                "StringEquals": {
                    "aws:SourceVpce": "vpce-0a06670cbb04c3863"
                }
            }
        }
    ]
}
```
Having a policy like this would simplify this all down a bit.

chalice/deploy/deployer.py
```diff
@@ -459,15 +459,38 @@ def _create_rest_api_model(self,
                 handler_name=auth.handler_string, stage_name=stage_name,
             )
             authorizers.append(auth_lambda)
+        policy = config.api_gateway_policy
+        if (config.api_gateway_endpoint_type == 'PRIVATE' and not policy):
```
kyleknap: I wonder if we should be generating this policy instead of when `config.api_gateway_endpoint_type == 'PRIVATE'` to when `config.api_gateway_endpoint_vpce` is not falsy? I'm mainly asking to see if we want to be able to support setting `api_gateway_endpoint_vpce` without setting an `api_gateway_endpoint_type`? This will allow users to autogenerate a policy for them without forcing the endpoint to be a private.

docs/source/topics/configfile.rst
```diff
@@ -40,11 +40,24 @@ a stage specific configuration value is needed, the ``stages`` mapping
...
   will also be the URL prefix for your API
   (``https://endpoint/prefix/your-api``).
 
+* ``api_gateway_endpoint_type`` - The endpoint configuration of the
+  deployed API Gateway which determines how the API will be accessed,
+  can be EDGE, REGIONAL, PRIVATE. Note this value can only be set as a
+  top level key and defaults to EDGE. For more information see
+  https://amzn.to/2LofApt
+
+* ``api_gateway_endpoint_vpce`` - When configuring a Private API a VPC
```
kyleknap: Do you think it makes sense to have this be a list of vpce_ids instead of a single vpce value? This is consistent with how we expose `subnet_ids` and `security_group_ids` for lambdas running in a vpc. I wonder if it is common if someone wants to have access from mutlitple vpc endpoints? Otherwise if it is not common, I'm fine just deferring to have them write out their own APIG policy.

docs/source/topics/configfile.rst
```diff
@@ -40,11 +40,24 @@ a stage specific configuration value is needed, the ``stages`` mapping
...
+  Endpoint must be specified to configure a default resource policy on
+  the API if an explicit policy is not specified.
+
+* ``api_gateway_policy`` - The IAM resource policy for the REST API.
```
kyleknap: For this variable, I think it makes sense to make the following changes:

* Do not force users to have to specify account_id, region_name, rest_api_id etc. I see this as a really low level configuration option and we should not be forcing constraints on the contents of the policy nor manipulating the contents.
* We should only support reading it from a file. While the `config.json` can get complex, I want to avoid it from being even more complex by adding support for arbitrary dictionaries as a value. This is also how we expose custom IAM policies for the lambda function.
* We rename the configuration option to `api_gateway_policy_file` to be consistent with `iam_policy_file`

Thoughts?

chalice/deploy/deployer.py
kapilt: couple of thoughts.

in general format() doesn't play nice with json, i was handling this in the cfn transform by doing a hunt on individual leaf string nodes to replace.

that simplification of the access policy looks good and avoids a good bit of complexity

i'm a bit unclear for enterprise usage scenarios which need the private endpoints/ vpc if doing away with variable interpolation is going to cut it, ie those vars are pretty fundamental to crafting the arn on a custom policy.  But the resource policy here feels  effectively like a network reachability one, and its unclear if there's anything finer grained (ie principal specification) allowed since authn is a separate mechanism with apigw.  nutshell the change sounds good, and will simplify things in a few places, but i'd like to verify the reachability behavior and interaction with authn if specifying a principal arn.

chalice/deploy/deployer.py
kapilt: the auto gen'd policy and vpce doesn't really have meaning without configuring the endpoint to being private.

docs/source/topics/configfile.rst
kapilt: i think in that context it would be better to defer to the user just providing a policy, ie it feels pretty corner case then common case usage, albeit its an easy enough change to support. actually i'm a bit curious if the vpce identity carries over multiple accounts in which case i think your right it might be nicer to support it directly. i'm not sure how that would interact with the vpc dns resolution (re multi-account) setups. worth exploring.

docs/source/topics/configfile.rst
kapilt: the problem is a user does want to provide a scoped down policy, they need those values to be able to craft the arn correctly which is fundamental to providing custom iam policies, perhaps they have static hard codes for account id and region and do some additional devops to make it work across accounts/regions for principial arns etc, but its rather hard for them to supply rest_api_id outside of the framework providing it. its unclear if that is needed for scope down, most of the examples have it, but its embedded on the resource directly so it shouldn't really grant access to things other than itself.

re separate file sounds fine, but i'll note that lambda iam policy is fairly arbitrary in length and nature while this is fairly constrained from a usage & definition perspective.

overall the suggested changes from this round of reviews sounds good and should enable simplifying this a bit, thanks.

chalice/deploy/deployer.py
kyleknap: I just did some more diving into this... According to the docs: https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-resource-policies-aws-condition-keys.html, you cannot specify a vpce-based policy if the APIG API is not private. Given that is the case, let's keep the logic the way it is.

chalice/deploy/deployer.py
kyleknap: Sounds good to me. In general, it would make sense to get a better understanding of the policies patterns people will use.

docs/source/topics/configfile.rst
kyleknap: >> actually i'm a bit curious if the vpce identity carries over multiple accounts in which case i think your right it might be nicer to support it directly

I was too. So I spun up my private API in another account and allowed access from a VPC endpoint in another account, and I was able to make API calls from that VPC. So it definitely seems doable.

Thinking it over more, I think it is fine as is for the following reasons:

* For the Lambda security groups and subnets, we chose to use a list because that is what the API needs in setting up the Lambda function.
* If we want to support multiple vpce's in the future, we can overload the value to support a list eventually and do this if we get future requests for it.
* Once a user starts wanting to manage multiple vpce's, they will also may want to add more complicated access policies as well. At that point, it would probably make sense to just rely on the policy file.

In the end, I see this parameter as something to get a user up and running in a straightforward fashion with a private API. Then if they want to expand on the policy itself, they can grab the policy, save it in their chalice project, and update it as needed.

docs/source/topics/configfile.rst
kyleknap: >> unclear if that is needed for scope down, most of the examples have it, but its embedded on the resource directly so it shouldn't really grant access to things other than itself.

Yeah. If the APIG policy was a standalone resource that could be shared across APIG API's needing the rest_api_id, it would make sense. However given that the policy is not standalone and can only be attached to a single APIG API, it seems like the `account_id`, `region`, `api_id` is not really useful other than accidentally attaching the wrong policy to an API? In terms of scoping down access to the resources of an APIG API, you would really only want to do it for methods and routes: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-control-access-using-iam-policies-to-invoke-api.html#api-gateway-iam-policy-resource-format-for-executing-api, in which case you can still use `*` for `region`, `account_id`, `api_id` and get the same result if you had specified the literal value instead.

chalice/deploy/deployer.py
kapilt: i simplified this a bit with just static policies sans interpolation, the custom policy escape should hopefully address the majority of customization concerns.

chalice/deploy/planner.py
```diff
@@ -925,19 +944,9 @@ def _plan_restapi(self, resource):
                         'swagger_document': resource.swagger_doc,
                     },
                 ), "Updating rest API\n"),
-                models.APICall(
-                    method_name='deploy_rest_api',
-                    params={'rest_api_id': Variable('rest_api_id'),
-                            'api_gateway_stage': resource.api_gateway_stage},
-                ),
-                models.APICall(
-                    method_name='add_permission_for_apigateway',
```
kapilt: fwiw this api call was a duplicate, hence the removal.

docs/source/topics/configfile.rst
kapilt: switched over to supporting a list or string of vpce ids.

docs/source/topics/configfile.rst
kapilt: removed variable interpolation and switched to user specifying a policy file with the suggested configuration name.

PR #1165, opened 2019-07-08
Make MultiDict mutable

### LINKED ISSUE #1158: query_params now returns a chalice.app.MultiDict object
[This commit](https://github.com/aws/chalice/commit/3a0e15f17a4632bfff3376490ad80a817f967ebf) has broken our app.

app.current_request.query_params now returns a chalice.app.MultiDict object, whereas it used to be a dict. So we could call app.current_request.query_params.pop('xyz', None) but can't do that anymore.

Not a huge deal as we can make changes at our end, but it's slightly annoying since there are quite a few patterns of this usage in our app.

stealthycoin:
fixes #1158 

Make `MultiDict` mutable again. The initial pull request made it readonly which is not backwards compatible with the raw `dict` it used to be.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -22,6 +22,7 @@ _PARAMS = re.compile(r'{\w+}')
 try:
     from urllib.parse import unquote_plus
     from collections.abc import Mapping
+    from collections.abc import MutableMapping
 
     unquote_str = unquote_plus
 
@@ -31,6 +32,7 @@ try:
 except ImportError:
     from urllib import unquote_plus
     from collections import Mapping
+    from collections import MutableMapping
 
     # This is borrowed from botocore/compat.py
     def unquote_str(value, encoding='utf-8'):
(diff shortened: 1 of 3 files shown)


chalice/app.py
```diff
@@ -160,26 +162,55 @@ class MultiDict(Mapping):
...
+                else:
+                    dict_mapping[k] = [v]
+            mapping = dict_mapping
 
         self._dict = mapping
+        self._remapped_keys = set()
 
-    def __getitem__(self, k):
-        values_list = self._dict[k]
+    def add(self, k, v):
```
kapilt: what's this method for?

stealthycoin: Ah a stub I forgot. Initially it was going to be https://werkzeug.palletsprojects.com/en/0.15.x/datastructures/#werkzeug.datastructures.MultiDict.add

Since there is no other way to add a new value to an internal list, other than getting it, appending and resetting. Initially I was going to use it during `__init__` when creating the internal structure from a list of tuples. However it was awkward since the internal map hasn't been initialized so I opted to write it all out there in the `__init__` method so there is no real reason to have it, since it didn't exist before. I just forgot to remove the stub.

tests/unit/test_app.py
```diff
@@ -1962,7 +1965,73 @@ def test_multidict_list_wont_change_source(input_dict):
...
+        'value3',
+        {'key2': []},
+    ),
+])
+def test_multidict_list_can_pop_value(input_dict, key, popped, leftover):
+    d = MultiDict(input_dict)
+    pop_result = d.pop(key)
+    assert popped == pop_result
+
+    print(d._dict)
```
kyleknap: Leftover print statement.

PR #1172, opened 2019-07-12
Set minimumCompressionSize before API gateway stage is deployed

### LINKED ISSUE #1171: minimumCompressionSize is not applied until API Gateway stage is deployed
I've been trying out `minimumCompressionSize` feature and it doesn't seem to work until I manually, e.g. in AWS Console, deploy API gateway stage.

According to [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gzip-compression-decompression.html), API Gateway stage needs to be deployed after `minimumCompressionSize` is set: 
> After the compression is enabled and the API is deployed, the client can receive an API response with a compressed payload if it specifies an appropriate Accept-Encoding header in the method request.

Looking at the [code](https://github.com/aws/chalice/blob/master/chalice/deploy/planner.py#L768), it's set in `shared_plan_epilogue` which runs after `deploy_rest_api` call.

topu:
*Description of changes:*
This makes setting `minimumCompressionSize` happen before API Gateway stage is deployed and fixes #1171.

### DIFF
--- a/chalice/deploy/planner.py
+++ b/chalice/deploy/planner.py
@@ -760,17 +760,6 @@ class PlanStage(object):
         # the update and create case.
         shared_plan_epilogue = [
             models.APICall(
-                method_name='update_rest_api',
-                params={
-                    'rest_api_id': Variable('rest_api_id'),
-                    'patch_operations': [{
-                        'op': 'replace',
-                        'path': '/minimumCompressionSize',
-                        'value': resource.minimum_compression,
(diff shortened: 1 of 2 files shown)


kapilt:
fwiw, #1160 also ends up doing this albeit for a different reason.

jamesls:
https://github.com/aws/chalice/pull/1160 has been merged, closing.

PR #1174, opened 2019-07-16
Add support for packaging Chalice app with websockets

stealthycoin:
Add websocket (API Gateway v2) support to `chalice package` command
through CloudFormation. Removes the test indicating that
CloudFormation does not support websocket APIs.

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -247,21 +247,182 @@ class SAMTemplateGenerator(object):
             }
         }
 
+    def _add_websocket_lambda_integration(
+            self, api_ref, websocket_handler, resources):
+        # type: (Dict[str, Any], str, Dict[str, Any]) -> None
+        resources['%sAPIIntegration' % websocket_handler] = {
+            'Type': 'AWS::ApiGatewayV2::Integration',
+            'Properties': {
+                'ApiId': api_ref,
+                'ConnectionType': 'INTERNET',
+                'ContentHandlingStrategy': 'CONVERT_TO_TEXT',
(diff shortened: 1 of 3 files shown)


chalice/package.py
```diff
@@ -247,22 +247,183 @@ def _inject_restapi_outputs(self, template):
...
+        # the autogen'd names that contain the md5 suffixes.
+        stage_name = template['Resources']['WebsocketAPIStage'][
+            'Properties']['StageName']
+        outputs = template['Outputs']
+        resources = template['Resources']
+        outputs['WebsocketAPIId'] = {
+            'Value': {'Ref': 'WebsocketAPI'}
+        }
+        if 'WebsocketConnect' in resources:
+            outputs['WebsocketConnectHandlerArn'] = {
```
kapilt: minus the web socket url, are these additional outputs useful for consumers?, else it feels like its adding to the public interface outputs (ie. adding compatibility constraints) without clear value

stealthycoin: Since we do not have any sort of useable build hooks there isn't a way to modify our generated cfn template from within chalice. So the canonical way, or at least the way I do it, is to have some wrpper script that calls `chalice package tmp/something` and then reads in the generated template and modifies it. This means the keys for these things need to be stable, and are implicitly a part to for API contract. We have had to change names before to make lambda function names not change, so this is something other people are doing as well. Here I'm making it more explicit. If its a part of the output we cant change it. Well, it is experimental so we still could, but the point is I'm trying to make a stronger statement about these being a part of the external API so noone accidentally changes them, or misunderstands.

PR #1176, opened 2019-07-17
deterministic zip file generation

### LINKED ISSUE #1114: Deterministic zipfile generation
When running `chalice package` multiple times, the zipfile outputs are normally different. This appears to be due to both order of file insertion and also zipfile mtime timestamps. For reference, the package https://github.com/esoule/strip-nondeterminism provides the ability to strip "extended file attributes" (`zip -X`) and reset mtimes, but not to reorder files in a deterministic order.

To enable change/drift detection, it's important to know when a package has meaningfully changed. It would be nice if Chalice built the zipfile in such a way that it's bit for bit reproducible when the input files/wheels have not changed.

kapilt:
closes #1114 

Normalize zip file generation for time stamps and unix permissions.

### DIFF
--- a/chalice/utils.py
+++ b/chalice/utils.py
@@ -14,6 +14,7 @@ import click
 from typing import IO, Dict, List, Any, Tuple, Iterator, BinaryIO, Text  # noqa
 from typing import Optional, Union  # noqa
 from typing import MutableMapping  # noqa
+from typing import cast  # noqa
 
 from chalice.constants import WELCOME_PROMPT
 
@@ -100,23 +101,47 @@ class ChaliceZipFile(zipfile.ZipFile):
     """Support deterministic zipfile generation.
 
     Normalizes datetime and permissions.
+
     """
 
     compression = 0  # Try to make mypy happy.
+    _default_time_time = (1980, 1, 1, 0, 0, 0)
+
+    def __init__(self, *args, **kwargs):
+        # type: (Any, Any) -> None
+        self._osutils = cast(OSUtils, kwargs.pop('osutils', OSUtils()))
+        super(ChaliceZipFile, self).__init__(*args, **kwargs)
 
     # pylint: disable=W0221
     def write(self, filename, arcname=None, compress_type=None):
         # type: (Text, Optional[Text], Optional[int]) -> None
-        # Only supports files, py2.7 and 3 have different signatures
+        # Only supports files, py2.7 and 3 have different signatures.
+        # We know that in our packager code we never call write() on
+        # directories.
+        zinfo = self._create_zipinfo(filename, arcname, compress_type)
+        with open(filename, 'rb') as f:
+            self.writestr(zinfo, f.read())
 
-        info = zipfile.ZipInfo(arcname or filename)
-        # Grant other users permissions to read,
-        # else lambda blows up with mysterious error.
-        # http://unix.stackexchange.com/questions/14705/
-        info.external_attr = 0o644 << 16
-        info.compress_type = compress_type or self.compression
-        with open(filename, 'rb') as fh:
-            self.writestr(info, fh.read())
+    def _create_zipinfo(self, filename, arcname, compress_type):
+        # type: (Text, Optional[Text], Optional[int]) -> zipfile.ZipInfo
+        # The main thing that prevents deterministic zip file generation
+        # is that the mtime of the file is included in the zip metadata.
+        # We don't actually care what the mtime is when we run on lambda,
+        # so we always set it to the default value (which comes from
+        # zipfile.py).  This ensures that as long as the file contents don't
+        # change (or the permissions) then we'll always generate the exact
+        # same zip file bytes.
+        # We also can't use ZipInfo.from_file(), it's only in python3.
+        st = self._osutils.stat(str(filename))
+        if arcname is None:
+            arcname = filename
+        arcname = self._osutils.normalized_filename(str(arcname))
+        arcname = arcname.lstrip(os.sep)
+        zinfo = zipfile.ZipInfo(arcname, self._default_time_time)
+        zinfo.external_attr = (st.st_mode & 0xFFFF) << 16
+        zinfo.file_size = st.st_size
+        zinfo.compress_type = compress_type or self.compression
+        return zinfo
 
 
 def create_zip_file(source_dir, outfile):
@@ -130,7 +155,8 @@ def create_zip_file(source_dir, outfile):
 
     """
     with ChaliceZipFile(outfile, 'w',
-                        compression=zipfile.ZIP_DEFLATED) as z:
+                        compression=zipfile.ZIP_DEFLATED,
+                        osutils=OSUtils()) as z:
         for root, _, filenames in os.walk(source_dir):
             for filename in filenames:
                 full_name = os.path.join(root, filename)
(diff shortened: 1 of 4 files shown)


stealthycoin:
I really like this idea.

kapilt:
mypy seems to have a bug, the signature here exactly matches the base class re
`chalice/utils.py:107: error: Signature of "write" incompatible with supertype "ZipFile"`

kapilt:
so the signature here actually differs from 2.7 to 3.6, the typeshed/mypy seems to only have the 2.7 definition.

kapilt:
hmm.. stat.filemode doesn't exist in 2.7..

jamesls:
Nice!  I was trying this out and initially wasn't seeing consistent checksums of the deployment zips.  It looks like you also need to update the `open_zip` method on `OSUtils`:

https://github.com/aws/chalice/blob/1cfae93d90dc107577e1431c2c0d8c380e5a4e52/chalice/utils.py#L129-L131

I'd also like to add a functional test that actually generates a deployment package twice and verifies that the zip contents are the same.

kapilt:
Thanks for looking over, sounds good re a functional md5 test. its probably worth saying that most of the value chain for this pr initially will be with packaged deploys via terraform/cloudformation, as chalice planner needs a refactor compare md5 prior to upload. Next step would be getting the auto-layer work merged which separates app code and dependency code which remove the bulk of uploads size in conjunction with this work.

jamesls:
> its probably worth saying that most of the value chain for this pr initially will be with packaged deploys via terraform/cloudformation, as chalice planner needs a refactor compare md5 prior to upload.

Yep, I agree.

I'd like to get this in for the next release so we can at least start laying the groundwork for the smarter delta-updates.

jamesls:
I went ahead updated utils.py.  Also added a functional/integration test to verify this works.  I opted not to add a changelog entry as this doesn't have any end user impact just yet.  Ready to be looked at again.

cc @kyleknap

chalice/utils.py
```diff
@@ -96,6 +96,29 @@ def serialize_to_json(data):
...
+    # pylint: disable=W0221
+    def write(self, filename, arcname=None, compress_type=None):
+        # type: (Text, Optional[Text], Optional[int]) -> None
+        # Only supports files, py2.7 and 3 have different signatures
+
+        info = zipfile.ZipInfo(arcname or filename)
+        # Grant other users permissions to read,
+        # else lambda blows up with mysterious error.
+        # http://unix.stackexchange.com/questions/14705/
+        info.external_attr = 0o644 << 16
```
kyleknap: Should we always be setting the permissions to a specific value? Is it possible say a user wants a file in the `vendor` directory to be executable?

jamesls:
Picking this up again...

chalice/utils.py
jamesls: So I looked into this, and yes it's possible that something in vendor is executable, and we should be preserving this.

I think the issue mentioned in the comments should be tackled separately from the issue of deterministic zip file generation.  Specifically if the file isn't readable by lambda, you'll run into issues when you run your lambda function (for example: https://github.com/aws/chalice/issues/1350#issuecomment-585453733).

All we really care about here is that we don't set the `date_time` to the mtime of the file.  So rather than duplicate the logic in `from_file` (https://github.com/python/cpython/blob/master/Lib/zipfile.py#L490) I'd rather just call `from_file` but then set the `date_time` to the default value of `(1980, 1, 1, 0, 0, 0, )` on the `ZipInfo`.  I'll also add a test that checks the file permissions.

As a separate PR, we can add logic that ensures the file is always readable.

jamesls:
Alright, rebased and pushed up changes.  Should be ready to look at again.

chalice/utils.py
```diff
@@ -96,6 +97,53 @@ def serialize_to_json(data):
...
+        # change (or the permissions) then we'll always generate the exact
+        # same zip file bytes.
+        # We also can't use ZipInfo.from_file(), it's only in python3.
+        st = self._osutils.stat(str(filename))
+        if arcname is None:
+            arcname = filename
+        arcname = self._osutils.normalized_filename(str(arcname))
+        arcname = arcname.lstrip(os.sep)
+        zinfo = zipfile.ZipInfo(arcname, self._default_time_time)
+        zinfo.external_attr = (st.st_mode & 0xFFFF) << 16
```
kapilt: can we get a comment in here what this is doing?  or can we use the octal form here (assuming 2.7 is dropped).

originally this was set to ensure files were always readable in lambda, but then the question of executable bins in uploads came up, part of the original was i've come across many enterprise environments where per some cis or stig benchmark they do default restrictive umask, that breaks lambda if just respecting file perms, it also has the bonus of not considering them for file delta. manual construction of the octal with a check of 747 or 644 on the basis of extant file executable would read more cleanly here.

kapilt: confirmed that behavior is still current

```python
import os
from pwd import getpwuid

def lambda_handler(event, context):
    stat = os.stat(__file__)
    print("File Owner %r" % ( getpwuid(stat.st_uid),))
    print("Executor %r" % ( getpwuid(os.geteuid()),))

    return 22
```

```
START RequestId: 4d17de89-800b-4871-8d17-063b5c233aa6 Version: $LATEST
File Owner pwd.struct_passwd(pw_name='slicer', pw_passwd='x', pw_uid=498, pw_gid=497, pw_gecos='', pw_dir='/tmp', pw_shell='/sbin/nologin')
Executor pwd.struct_passwd(pw_name='sbx_user1051', pw_passwd='x', pw_uid=496, pw_gid=495, pw_gecos='', pw_dir='/home/sbx_user1051', pw_shell='/sbin/nologin')
END RequestId: 4d17de89-800b-4871-8d17-063b5c233aa6
```

chalice/utils.py
kapilt: my two cents, we're packaging up the files here, why we wouldn't work to make the packaged files usable here seems odd, there's already tests here around permission preservation.

jamesls: I started down this path, and turns out it's much more complex than this change (especially given the parity differences between zipfile in py2 and py3) and I'd rather have smaller changes going through than block this on getting those details right.  I agree we should fix this, but I tend to prefer smaller more focused PRs.

chalice/utils.py
jamesls: >or can we use the octal form here (assuming 2.7 is dropped).

Py2 is not dropped :(.

> manual construction of the octal with a check of 747 or 644 on the basis of extant file executable would read more cleanly here.

Just to be clear this isn't constructing an octal.  This is a zip file spec quirk that the mode is part of the upper 16 bits of `external_attr` so we just need to shift it up to the right place.  I'll add a comment about that.

> note a 755 perserve on executable would not work in lambda, as the user executing the lambda is not the same as the owner of the files afaicr

Not sure what you mean here.  It will work in the sense that you'll be able to execute a 0o755 file on lambda if we preserve that mode.  I confirmed that this is the case.

jamesls:
It sounds like the remaining point of discussion is ensuring that files in the zip are always readable.  I agree we should change that, but I don't want to block this being merged on that change.  I'm not hearing any feedback on the specifics of this change so I'm going to go ahead and merge this.  I'll try and tackle the always-readable problem soon unless someone else beats me to it.

kapilt:
merged

Issue #1127, opened 2019-05-25
automated layer construction


kapilt:
Chalice supports deploying multiple functions from the same set of app files, but each will effectively end up with a copy of the dependencies. Chalice supports layers via config.json that are constructed out of band by the end user. Given chalice has knowledge of the applications dependencies from requirements.txt and vendor, constructing an app layer and deployment zip and associating it to the functions would be nice. Primary benefits afaics are reducing code storage size on the account for the set of lambdas and making the resulting lambdas editable via console ui.

kapilt:
looks like this is covered by #1001 which covers existing layer configuration (since implemented) and dependency layer automation.

PR #1179, opened 2019-07-17
automatic layers

kapilt:
closes #1063  related #1127 

Automatic construction of layer for a chalice app (by stage) and reused across all an app's lambda functions. Its configurable and enabled by default. Old layers are automatically removed on deploy and delete.

### DIFF
--- a/chalice/deploy/sweeper.py
+++ b/chalice/deploy/sweeper.py
@@ -1,4 +1,4 @@
-from typing import List, Dict  # noqa
+from typing import List, Dict, Tuple  # noqa
 
 from chalice.config import Config, DeployedResources  # noqa
 from chalice.deploy import models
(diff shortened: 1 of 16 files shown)


kapilt:
ready for review/feedback. the test failure on 2.7 seems to be a travis anomaly wrt to docutils.

kapilt:
let's go ahead and add config for it default on, and auto delete.. ordering delete to happen after publish and config on existing lambdas.

chalice/package.py
```diff
@@ -136,10 +151,15 @@ def _generate_lambdafunction(self, resource, template):
             }
             lambdafunction_definition['Properties'].update(
                 reserved_concurrency_config)
-        if resource.layers:
+
+        layers = resource.layers or []  # type: List[Any]
+        if self._chalice_layer:
+            layers.append({'Ref': self._chalice_layer})
```
kapilt: this needs to be inserted at the start, we want user explicit layers to have precedence.

chalice/deploy/sweeper.py
```diff
@@ -78,79 +170,8 @@ def _plan_deletion(self,
         # type: (...) -> None
         for name in remaining:
             resource_values = deployed.resource_values(name)
-            if resource_values['resource_type'] == 'lambda_function':
```
kapilt: Fwiw, This for/if loop refactoring was triggered by hitting flake8  check on cyclomatic complexity limits when adding another conditional.

jamesls:
Just a heads up, I'm going to pick up this work again and try to get this merged.  I'll get it rebased on the latest master and test it out.

jamesls:
I've just about got everything rebased and working, but in testing this out, I'm noticing a difference in behavior with layers.  For layers, the zip file is unzipped to `/opt`, which means that in order for us to get the same behavior as before with no layers, when creating the layer zip file we need to unzip to `python/lib/{python-version}/site-packages` (documented [here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-manage)).

While that works for python files, it's not going to work for `vendor/` files.  That's because vendor files either can be python packages that you were intending to import (i.e custom built python packages), *or* they can be static assets you just want dumped into the lambda function.  That means that today if I have a file that's in `./vendor/my-picture.png`, then in my app code I can say `with open('my-picture.png')` and it'll work.  However, with layers, that file will be over in `/opt/python` somewhere.

Because of this incompatibility, I don't think we'll be able to default to using automatic layers.  This will have to be something you opt into for the time being until we can come up with a way to smooth other this difference.

jamesls:
Closing, https://github.com/aws/chalice/pull/1487 has been merged/released.

PR #1187, opened 2019-07-25
Remove Duplicate Log Entry

subssn21:
Issue # 1183

Remove the second Log entry for the same exception when reporting to app.log.


kyleknap:
@subssn21 Thanks for the PR. It looks like only a test needs to be updated here. Would it be possible to update the string we asserting for the exception? If not, I'd be to build off of this and see it through.

kyleknap:
Thanks! Merging.

Issue #1188, opened 2019-07-26
Unable to get `json_body` when Content-Type is 'application/csp-report'


Veebers:
Currently the only time you can retrieve the `Request.json_body` is when the `Content-Type` header is set to `application/json`. This means one is unable to retrieve the report payload when the request is a Content Security Policy violation report (as per https://www.w3.org/TR/CSP2/#violation-reports).

This limits using Chalice as an endpoint for `report-uri`.

Veebers:
A PR for this is incoming :-D

kyleknap:
@Veebers Thanks for the PR! Looking now.

PR #1189, opened 2019-07-26
Csp report json body

Veebers:
*Issue #, if available:*
https://github.com/aws/chalice/issues/1188

*Description of changes:*

This branch allows `Request`s with either the headers `application/json` or `application/csp-report` to return json data from `Request.json_body` (currently `application/json` is suported).

This is to allow apps created using Chalice to be Content Security Policy report violation endpoints (i.e. `report-uri` in the `Content-Security-Policy` or `Content-Security-Policy-Report-Only` headers).

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -353,6 +353,11 @@ class CORSConfig(object):
 class Request(object):
     """The current request from API gateway."""
 
+    ACCEPTABLE_JSON_CONTENT_TYPES = (
+        'application/json',
+        'application/csp-report'
+    )
+
     def __init__(self, query_params, headers, uri_params, method, body,
                  context, stage_vars, is_base64_encoded):
         self.query_params = None if query_params is None \
@@ -389,7 +394,12 @@ class Request(object):
 
     @property
     def json_body(self):
-        if self.headers.get('content-type', '').startswith('application/json'):
+        content_type = self.headers.get('content-type', '')
+        acceptable = any([
+            True for acceptable_type
+            in self.ACCEPTABLE_JSON_CONTENT_TYPES
+            if content_type.startswith(acceptable_type)])
+        if acceptable:
             if self._json_body is None:
                 try:
                     self._json_body = json.loads(self.raw_body)
--- a/tests/unit/test_app.py
+++ b/tests/unit/test_app.py
@@ -523,7 +523,8 @@ def test_json_body_available_with_right_content_type(create_event):
 def test_json_body_available_with_csp_report_content_type(create_event):
     demo = app.Chalice("demo-app")
 
-    @demo.route("/", methods=["POST"])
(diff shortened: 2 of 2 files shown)


chalice/app.py
```diff
@@ -389,7 +394,12 @@ def raw_body(self):
 
     @property
     def json_body(self):
-        if self.headers.get('content-type', '').startswith('application/json'):
+        content_type = self.headers.get('content-type', '')
+        acceptable = any([
+            True for acceptable_type
+            in self.ACCEPTABLE_JSON_CONTENT_TYPES
+            if content_type.startswith(acceptable_type)])
+        if acceptable:
```
ibolmo: Could probably go with `content_type in self.ACCEPTABLE_JSON_CONTENT_TYPES` and make the `ACCEPTABLE_JSON_CONTENT_TYPES` a frozenset.

Veebers: That would be cleaner but wouldn't work if the `Content-Type` was `application/json; charset=utf-8` :-(

kyleknap:
Thanks for the pull request. To start, I just had some questions about the background/context for the update:

1) For content types of `application/csp-report`, is the body only valid if it is JSON?
2) Is there a particular reason for just not using `json.loads(app.current_request.raw_body)` other than the `app.current_request.json_body` would make it more succinct?

Veebers:
No worries 😄,

1. Yes the body must only be JSON as per the spec, anything else is incorrect
2. In my current usage I could have used the suggested `json.loads(app.current_request.raw_body)` but it was a surprise to me and took me a moment to realise why `app.current_request.json_body` wasn't returning what I expected. 
    It is my desire to prevent future people getting tripped by this as well.

I consider this a suitable fix (due to it being within spec) but I can understand taking a closer look at why it might be a desirable addition 😄.

kyleknap:
Right now. I'm leaning toward holding off on this pull request for a couple of reasons:

1) It is not very much effort to reproduce the `to_json` logic. it is just consists of a `json.loads(app.current_request.raw_body)`. 

2) I feel like it is possible there will be future requests to be able to customize the JSON loading of the request body. Specific customizations that I can see requested include which content types to load on and actual implementation of `json.load` to use. And I think this customization of which content types to load on should be part of this larger story of having more granular control over json loading. The big problem is that we have not collected enough examples of how users want to customize the loader to come up with a good abstraction.

That being said. I propose we keep the feature request open but close this pull request. If enough people ask support for this via the feature request you opened, I think we should reopen this PR and pull in the logic. Furthermore, if enough requests come in concerning how people want to load the json body to warrant coming up with a good customizations story, we should make sure to be able to handle this case in the larger story and potentially reopen/pull in the PR.

Thoughts?

wm842:
Also needs support for `"content-type":"application/reports+json"`

Issue #1120, opened 2019-05-17
Published a plugin for pytest


studio3104:
This is not an issue but just an announcement, I published a plugin just as titled. https://pypi.org/project/pytest-chalice/

I adopted the LocalGateway approach with reference to the methodology discussed in [this issue](https://github.com/aws/chalice/issues/289).
Also, exposed an abstract client fixture like `pytest-django` and `pytest-flask`, for example,

```python
from chalice import Chalice

app = Chalice(__name__)


@app.route('/')
def index:
    return {'hello': 'world'}
```

```python
from http import HTTPStatus


def test_index(client):
    response = client.get('/')
    assert response.status_code == HTTPStatus.OK
    assert response.json == {'hello': 'world'}
```

Any feedback is welcome, I'd be happy if you guys try it out, thanks.

jamesls:
Nice!  What would you think about porting over some of that stuff to chalice directly?  I think the pytest fixtures would have to be separate, but things like the `InternalLocalGateway` and `ResponseHandler` could be added to chalice.

studio3104:
Thanks for your suggestion, I'm more than happy to do that, going to make a PR to port them then!

jamesls:
Closing, #1468 has been merged and is now available in the 1.17.0 release.  Thanks for your work on this!

PR #1193, opened 2019-07-29
 Transplant implementations from `pytest-chalice`

studio3104:
*Issue #, if available:*
https://github.com/aws/chalice/issues/1120

*Description of changes:*
As @jamesls suggested, I transplanted the implementations from `pytest-chalice`
https://github.com/aws/chalice/issues/1120#issuecomment-494598033

I didn't have strong confidence against the way to split the namespaces, so I just added some files at this moment.


studio3104:
@jamesls Though it's been a while since I opened this PR, I've updated my implementation to pass `make prcheck` just before, PTAL at your leisure?

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
```
jamesls: I think we can move this to a separate module, maybe `chalice.testclient` or `chalice.testutils`?

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+logger = getLogger(__name__)
+
+# Python 3.4 or older don't have JSONDecodeError within json module
+try:
+    JSONDecodeError = json.JSONDecodeError  # type: ignore
+except AttributeError:
+    JSONDecodeError = ValueError  # type: ignore
+
+
+class InternalLocalGateway(LocalGateway):
```
jamesls: I think we can update `LocalGateway` to add support for a custom context.  That way we don't need to separate `InternalLocalGateway` class.

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+        # type: (str) -> Any
+        try:
+            return self.values[key]
+        except KeyError:
+            raise AttributeError(
+                "'%s' object has no attribute '%s'" % (
+                    self.__class__.__name__, key))
+
+
+class RequestHandler(object):
```
jamesls: I think having an explicitly named type such as `TestHTTPClient` would be helpful instead of `RequestHandler`.

I'd would like to eventually add support for testing other chalice resources (`@app.lambda_function()`, `@app.schedule()`, `@app.on_s3_event()`), so having an explicit type of `TestHTTPClient` would be great.

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+            InternalLocalGateway, self)._generate_lambda_event(
+                method, path, headers, body)
+        event['requestContext'].update(self.custom_context)
+        return event
+
+
+UPPERCASE_PATTERN = re.compile('([A-Z])')
+
+
+class ResponseHandler(object):
```
jamesls: I think we should have an abstraction on top of the response dictionary returned.  A user doesn't return this dictionary directly, Chalice automatically creates this value so I'm concerned in exposing this directly as the API you assert against.  This means that the dictionary occasionally changes when API Gateway adds support for more keys.  For example, we now have support for `multiValueHeaders`.

If we have our own response class, we could add these changes in a backwards compatible way.

What do you all think?   

cc @stealthycoin

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+    METHODS = (
+        'get', 'head', 'post', 'options', 'put',
+        'delete', 'trace', 'patch', 'link', 'unlink')
+
+    def __init__(self, app):
+        # type: (Chalice) -> None
+        self.local_gateway = InternalLocalGateway(app, Config())
+
+    @property
+    def custom_context(self):
```
jamesls: Instead of having this `custom_context` property, what if we just make that a parameter to the client?

```
client.get('/', custom_context={...})
```

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class RequestHandler(object):
+    METHODS = (
+        'get', 'head', 'post', 'options', 'put',
+        'delete', 'trace', 'patch', 'link', 'unlink')
+
+    def __init__(self, app):
+        # type: (Chalice) -> None
+        self.local_gateway = InternalLocalGateway(app, Config())
```
jamesls: Let's make this an internal attribute `self._local_gateway` if this doesn't need to be public.

tests/unit/local/test_internal.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from chalice import Chalice
+from chalice.local.internal import RequestHandler
+
+
+@pytest.fixture
+def sample_app():
```
jamesls: Can you add some additional test cases?

I'm thinking:

* One for returning a `Response()` object.
* One that has multi value headers (a value in the `headers` dict is a list).
* A test for a non 200 response.
* A test where the view function raises an unexpected exception.

chalice/local/internal.py
```diff
@@ -0,0 +1,126 @@
...
+from chalice.local import LocalGateway
+from chalice.local import EventType, HeaderType  # noqa
+
+logger = getLogger(__name__)
+
+# Python 3.4 or older don't have JSONDecodeError within json module
+try:
+    JSONDecodeError = json.JSONDecodeError  # type: ignore
+except AttributeError:
+    JSONDecodeError = ValueError  # type: ignore
```
jamesls: I'm ok with just always catching `ValueError` instead of `JSONDecodeError` to keep things simple.

chalice/local/internal.py
studio3104: Renamed the class as you suggested!  https://github.com/aws/chalice/pull/1193/commits/2f783df65ba59ad7ff962d3780c550e0a70b6138

chalice/local/internal.py
studio3104: Made this internal. https://github.com/aws/chalice/pull/1193/commits/779febb95316e14cd83b7a8f0a7713ca6024f09e

chalice/local/internal.py
studio3104: `custom_context` is not a request parameter but an object to override the response, so I worried about how I should implement this interface.
And, I thought that I wanted to make this an object member of the client instance because I just wanted to write tests just like following,

```python
@pytest.fixture
def admin_client(client):
    client.custom_context = {...}  # e.g.) dummy Cognito's context for an user which belongs to `admin` group
    return client


def test_admin_behavior(admin_client):
    response = admin_client.get(...)
    # Assetions...
```

I guess this allows me to write the test simply, but how do you think??

chalice/local/internal.py
studio3104: Updated https://github.com/aws/chalice/pull/1193/commits/1adab1be1390c4c4ac6f2117ad3e27f836457a62

chalice/local/internal.py
studio3104: Change the namespaces https://github.com/aws/chalice/pull/1193/commits/2c7ce41e7987ebd12e9ebde67300c64b1d1d0ad9

Sytten:
Would it possible to allow testing for events from other sources? I feel this is a big missing point of the framework.

tests/unit/local/test_internal.py
studio3104: Added only test cases that you suggested,

- One for returning a Response() object.
  - https://github.com/aws/chalice/pull/1193/commits/4e413ee2dd1cd2e944c018109fc8a3138954974b
- One that has multi value headers (a value in the headers dict is a list).
  - https://github.com/aws/chalice/pull/1193/commits/838f061217f152c62c77e843209b93d20db12b5c
- A test for a non 200 response.
  - https://github.com/aws/chalice/pull/1193/commits/7d1873f55c2438f1cb23b6ea507f7ce282af4c1c
- A test where the view function raises an unexpected exception.
  - https://github.com/aws/chalice/pull/1193/commits/9b1aef71b7515b7d7e5e8e1954cef350c4583595

chalice/local/internal.py
studio3104: I may not be able to understand what you suggest right now.
You mean the class must not generate the values dynamically within the instance but map specific keys for each items, right?
Sorry to bother you but could you please show me a concrete example if not so...?

chalice/local/internal.py
studio3104: When I considered the porting again, I felt that the context should not be overwritten in this way.
The reason why I added `InternalLocalGateway` class was, because `LocalGateway` doesn't support `CognitoUserPoolAuthorizer` as of the current latest version.
The imeplementation I added is just a workaround, I am thinking that we would better to consider how the context should be treated when considering to add `CognitoUserPoolAuthorizer` handling feature on `LocalGateway` for now.

So, I am going to update the implementations like following,

- Abolish `InternalLocalGateway` class
- Add a new parameter to stub override context into `TestHTTPClient`
  - As @jamesls suggested https://github.com/aws/chalice/pull/1193#discussion_r330903200

chalice/local/internal.py
studio3104: Ref: https://github.com/aws/chalice/pull/1193#discussion_r352496402

chalice/local/internal.py
studio3104: I have considered carefully again about the stubbing custom context feature, but I feel like we would better not to add it at this moment.
The motivation why I added this was, because Chalice doesn't support Cognito's context for the local gateway for now, but I wanted to handle the response depending on Cognito's context. But if I imagine that we update `LocalGateway`, it would become a non-well-designed class, I guess.
I hope this will be considered again when `LocalGateway` actually support Cognito's authorization context, to make everything simple, I think it would better to stop adding it now.

chalice/local/internal.py
studio3104: Ref: https://github.com/aws/chalice/pull/1193#discussion_r352997311

chalice/local/internal.py
studio3104: https://github.com/aws/chalice/pull/1193/commits/160f6f7f8fae549af2b26980fe7ba39fa2f60631

studio3104:
@jamesls I responded against all review comments you left, so could you please check my response and updated commits?

studio3104:
@jamesls kindly ping :bow:

studio3104:
@jamesls Gentle ping 👶

knowsuchagency:
Any progress on this feature?

mazulo:
Hi @jamesls @stealthycoin first of all, thank you for all your hard work on this lib. It's incredibly helpful. Do you folks think is there anything missing on this PR? I believe this will be a great step to get `chalice` more mature and reliable.

jamesls:
Hey everyone, I've put together a proposal that builds on this over at: https://github.com/aws/chalice/issues/1468 along with the corresponding PR at https://github.com/aws/chalice/issues/1469.

The main thing I added was the ability to test all the event handlers (SNS, SQS, S3, websockets, pure lambdas, etc) in addition to just the REST APIs.  Thanks for your work on this.  Please let me know if you have any feedback on this proposal.

jamesls:
Closing, #1468 has been merged and is now available in the 1.17.0 release. Thanks for your work on this!

PR #1195, opened 2019-07-30
Add package argument to deep-merge in a JSON file during packaging

stealthycoin:
Its a common use case to need to modify a Chalice generated template
before deployment. Often to inject resources, values, or
configurations that are not supported directly by Chalice. It will
always be the case that something on AWS is not supported by Chalice
directly that a consumer may want to interact with. This PR opens up
an interface to address that gap more easily in Chalice itself.

The package command can now be invoked with a new argument:

  $ chalice package --merge-template extras.json out

This extras.json file should be a JSON formatted file which will be
deep-merged on top of the sam.json that is generated by Chalice.

For a simple example lets assume that we have the default new Chalice
project and that extras.json has the following content:

```
{
    "Resources": {
	"APIHandler": {
	    "Properties": {
		"Environment": {
		    "Variables": {
			"foo": "bar"
		    }
		}
	    }
	}
    }
}
```

The generated template located at out/sam.json will have the this
environment variable injected into it:

```
...
...
    "APIHandler": {
      "Type": "AWS::Serverless::Function",
      "Properties": {
        "Runtime": "python3.6",
        "Handler": "app.app",
        "CodeUri": "./deployment.zip",
        "Tags": {
          "aws-chalice": "version=1.9.1:stage=dev:app=test"
        },
        "Timeout": 60,
        "MemorySize": 128,
        "Role": {
          "Fn::GetAtt": [
            "DefaultRole",
            "Arn"
          ]
        },
        "Environment": {
          "Variables": {
            "foo": "bar"
          }
        }
      }
    },
...
...
```

Obviously this simple example could have been achieved using the
config file for the environment variables. But using the extras.json
file we could have injected a DynamoDB table into the template, and
injected the table name as a REF into the APIHandler's environment
variables.

For the time being the merge functionality is relatively simple. It can
be extended over time as we discover what types of merge extensions
are needed.

*Issue #, if available:*

*Description of changes:*


kapilt:
fwiw, I like the idea, it came up in discussion last week with @jamesls and @kyleknap, partly in reference to the terraform pr which also supports forward/backward refs. the main usage caveat is that most folks are using yaml these days, for which it would be nice to support directly (at least on the read side). I'll also note that using these refs also tends to enshrine that apigw is backed by a single lambda function (apihandler) which may not be always be the case, but the general thought is that the template constructs become part of the chalice public interface if folks start depending on them via reference in their automation templates.

chalice/package.py
```diff
@@ -594,3 +605,67 @@ def _fixup_deployment_package(self, template, outdir):
...
+                result[key] = value
+            else:
+                result[key] = self._merge(value, existing_value)
+            if result[key] is None:
+                del result[key]
+        return result
+
+    def _merge_replace(self, src, dst):
+        # type: (Any, Any) -> None
+        # Default merge behavior is to take the source value. This behaivor is
```
kapilt: speling typo

chalice/package.py
```diff
@@ -594,3 +605,67 @@ def _fixup_deployment_package(self, template, outdir):
...
+    def _merge_dict(self, src, dst):
+        # type: (Dict[str, Any], Dict[str, Any]) -> Dict[str, Any]
+        result = dst.copy()
+        for key, value in src.items():
+            existing_value = dst.get(key)
+            if existing_value is None:
+                result[key] = value
+            else:
+                result[key] = self._merge(value, existing_value)
+            if result[key] is None:
```
jamesls: Why do we want this?  I think this no longer makes this a "deep merge" if certain values have special semantics.

chalice/package.py
```diff
@@ -594,3 +605,67 @@ def _fixup_deployment_package(self, template, outdir):
...
+        try:
+            loaded_template = json.loads(template_data)
+        except ValueError:
+            raise RuntimeError(
+                'Expected %s to be valid JSON template.' % filepath)
+        return loaded_template
+
+
+class TemplateDeepMerger(object):
+    def merge(self, src, dst):
```
jamesls: FWIW, one of the things that made this hard to read was that the `src` and `dst` seemed backwards to me (in terms of their semantics).  Expanding this into the general case of merging a number of templates someone specifies I was expecting that :

* `src` - The base template that chalice generates
* `dst` (or eventually a list of `dst`) - Overrides/overlays that get merged on top of the values that chalice generates.

This seems to make sense to me because you have a base layer, and you overlay additional resources on top of this base layer.  This has the same behavior it would have now in that the users's template will override things that get generated in chalice.

chalice/package.py
```diff
@@ -594,3 +605,67 @@ def _fixup_deployment_package(self, template, outdir):
...
+        # type: (Dict[str, Any], Dict[str, Any]) -> Dict[str, Any]
+        return self._merge_dict(src, dst)
```
jamesls: It was a little hard to follow that we skip over the general purpose `_merge` and go right to the `_merge_dict`.  I don't know if I'm over simplifying this but, ignoring the behavior of deleting `None`s and also keeping the existing `src` wins behavior (which I think we should change as per my comment above), I would have expected this is as simple as:

```
    def merge(self, src, dst):
        return self._merge(src, dst)

    def _merge(self, src, dst):
        if isinstance(src, dict) and isinstance(dst, dict):
            return self._merge_dict(src, dst)
        return src

    def _merge_dict(self, src, dst):
        merged = dst.copy()
        for key, value in src.items():
            merged[key] = self._merge(value, dst.get(key))
        return merged

```

Caveat: haven't actually checked if this works.

I think if we start implementing different deep merge strategies then we can look at breaking this down into a template method.

chalice/config.py
```diff
@@ -223,6 +223,12 @@ def api_gateway_stage(self):
         return self._chain_lookup('api_gateway_stage',
                                   varies_per_chalice_stage=True)
 
+    @property
+    def package_merge_template(self):
```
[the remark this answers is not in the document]
stealthycoin: I was thinking of the case where I have different files I want to merge depending on my stage.

chalice/package.py
```diff
@@ -532,15 +537,15 @@ def _register_cfn_resource_name(self, name):
 
 class AppPackager(object):
     def __init__(self,
-                 sam_templater,     # type: SAMTemplateGenerator
-                 resource_builder,  # type: ResourceBuilder
-                 post_processor,    # type: TemplatePostProcessor
-                 osutils,           # type: OSUtils
+                 sam_templater,      # type: SAMTemplateGenerator
+                 resource_builder,   # type: ResourceBuilder
+                 post_processors,    # type: List[TemplatePostProcessor]
```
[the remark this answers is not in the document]
stealthycoin: Ah, that was my initial implementation. I thought it was overkill so I switched to this, I can switch it back.

chalice/package.py
```diff
@@ -594,3 +605,67 @@ def _fixup_deployment_package(self, template, outdir):
...
             resource['Properties']['CodeUri'] = './deployment.zip'
+
+
+class TemplateMergePostProcessor(TemplatePostProcessor):
+    def process(self, template, config, outdir, chalice_stage_name):
+        # type: (Dict[str, Any], Config, str, str) -> None
+        if not self._should_merge_template(config):
+            return
+        loaded_template = self._load_template_to_merge(config)
+        merger = TemplateDeepMerger()
```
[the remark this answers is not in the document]
stealthycoin: Yep.

chalice/package.py
stealthycoin: Yea I was immediately thinking of the case where someone wants to remove a key added by Chalice not replace it. I do agree its odd for the default behavior, I also can't think of a great concrete reason for it, it just seemed like an odd gap.

chalice/package.py
stealthycoin: I was thinking of it in terms of copying a source on top of a destination. But I agree its confusing. I can switch it to something like chalice_template, extra_template or something much more explicit so you don't need to map src/dst to real concepts when reading it.

chalice/package.py
stealthycoin: Yea so I had much more complex merging logic initially and I wound up removing a lot of it the more I thought about it, so some of this is because it was being used and I was leaving it there in case we wanted to add more cases back like having special logic for lists etc. But if everyone is happy with the basic starting merge behavior it has now, simplifying implementation is good with me.

PR #1201, opened 2019-08-07
Fix use of null_data_source

kyleknap:
The name of the data provider is `null_data_source` and not `null_data_provider`. I ran into this when trying out after we merged support for terraform. Specifically, I was running into the following error for the general `new-project` app:
```
Error: Invalid data source

  on chalice.tf.json line 85, in data:
  85:     "null_data_provider": {

The provider provider.null does not support data source "null_data_provider".
```
The fix here allows me to success `terraform apply` the template. It is also what the docs say to use for the name: https://www.terraform.io/docs/providers/null/data_source.html

@kapilt @jamesls


kapilt:
lgtm

PR #1204, opened 2019-08-07
Add a wait after sending messages stops in websocket integ test

stealthycoin:
It's possible for a message to be sent over a websocket, and then for
the websocket redeployment message drop integration test to execute a
scan on the DynamoDB table before the websocket message has traveled
through API Gateway -> Lambda far enough to make the DynamoDB PutItem
call. This means consistent read won't wait for it, and the list of
numbers read from the DynamoDB table will be missing the last entry.


jamesls:
I ran this in a loop and I still am seeing failures occassionally:


```

        ws = websocket.WebSocketApp(
            smoke_test_app_ws.websocket_connect_url,
            on_message=on_message,
            on_error=on_error,
            on_open=on_open,
        )
        # This will block until the websocket is closed.
        ws.run_forever()

        echoed = sorted(closure_values['echoed_values'])
        numbers = get_numbers_from_dynamodb(smoke_test_app_ws.app_dir)
>       assert echoed == numbers
E       assert [1, 2, 3, 4, 5, 6, ...] == [1, 2, 3, 4, 5, 6, ...]
E         Right contains more items, first extra item: 514
E         Use -v to get the full diff
```

PR #1206, opened 2019-08-08
Restructure config page

jamesls:
* Make all config options a section.  This lets us
  permalink to specific config options
* Put config options in sorted order.  Makes it easier
  to find the config option you're looking for.

Screen Shot 2019-08-08 at 3 54 32 PM


kyleknap (approved):
Nice!

PR #1212, opened 2019-08-13
fix for _generate_arn in AuthResponse

anthonygrant:
CustomAuthorizer fails when querystring params contain : or /

If the user passes pathParams that contain : or / and the Chalice app uses a CustomAuthorizer then the incoming_arn will be incorrect. Adding .split('?')[0] fixes this issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -1211,7 +1211,7 @@ class AuthResponse(object):
         return allowed_resources
 
     def _generate_arn(self, route, request, method='*'):
-        incoming_arn = request.method_arn
+        incoming_arn = request.method_arn.split('?')[0]
         parts = incoming_arn.rsplit(':', 1)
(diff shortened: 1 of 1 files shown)


stealthycoin:
Thanks for the PR. Can you add test cases for this please? There is also the case where the API Gateway itself has a ? in the name where this code won't work. Can you add a test case for that as well.

anthonygrant:
Thanks @stealthycoin . I look to create some tests soon. Can you be specific about the case where there would be a ? in the API Gateway in the name. I was thinking you might mean a custom domain name but a quick search confirms that that character is illegal so I think I'm looking in the wrong direction. Can you help me here so I can make sure my solution works in all situations and the tests can confirm that outcome.

stealthycoin:
Ah, I don't think its illegal. I made an api gateway API with ? in its name.

Issue #1205, opened 2019-08-07
Return multiple Set-Cookie headers


dmicic:
I am trying to return multiple Set-Cookie statements in my header. However, it doesn't work given that key within the dictionary is the same for both cookies and therefore, only the 2nd Set-Cookie is been returned resp. actually been added to the dict.

```js
"headers": {
  "Set-Cookie": "refreshToken=myToken1; httpOnly; secure; path=/auth/refresh/;",
  "Set-Cookie": "accessToken=myToken;2 httpOnly; secure; path=/;"
 }
```

Is there solution to this problem in chalice?

markalexander:
I'm not sure that API Gateway supports this, even if Chalice could, per https://forums.aws.amazon.com/thread.jspa?threadID=205782

dmicic:
@markalexander it is supported: https://forums.aws.amazon.com/ann.jspa?annID=6185

dmicic:
I am currently using one of those workarounds. But would be good to align chalice with the capabilities of the API Gateway.

stealthycoin:
Ah, ill mark as feature request.

jamesls:
Supported via #1214

PR #1214, opened 2019-08-14
Add support for multi value header responses

stealthycoin:
API Gateway added support for a multi-value header dictionary whcih can
be returned. This commit adds support for it in Chalice.

implements #1205 

API Gateway docs: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ Next Release (TBD)
 
 * Fix mouting blueprints with root routes
   (`#1230 <https://github.com/aws/chalice/pull/1230>`__)
+* Add support for multi-value headers responses
+  (`#1205 <https://github.com/aws/chalice/pull/1205>`__)
 
 
 1.11.0
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -410,11 +410,15 @@ class Request(object):
 
 
 class Response(object):
-    def __init__(self, body, headers=None, status_code=200):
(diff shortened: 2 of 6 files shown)


stealthycoin:
The other option here would be to overload the `headers` dict and just look at if its a string or an list. Would be a better API. The reason I didn't do that is last time we tried to be clever and preserve backwards compatibility we caused problems that needed fixing. 

Thoughts? This case seems simpler than the last one (#1165)

jamesls:
> The reason I didn't do that is last time we tried to be clever and preserve backwards compatibility we caused problems that needed fixing.

Are you referring to https://github.com/aws/chalice/pull/1165?  If so, I think that's a little different in the sense that we're not changing any existing behavior and just supporting a case we previously weren't (a list value in the headers dict).  Given we're not messing around with types or anything, I think this is much safer.

Plus, reading the doc page you linked, the interaction between `headers` and `multiValueHeaders` seems tricky and could be confusing fore people.

My preference is to just make a list-value automatically work.  What do others think?

kyleknap:
>> My preference is to just make a list-value automatically work. What do others think?

Yeah let's just make lists work because:

1) I agree with James it is a safer change than the request one.
2)  It is consistent with what we do for requests and would be more intuitive.

Issue #1021, opened 2018-12-17
Multiple file upload?


montao:
Is it possible to upload multiple files with the same request using chalice? If yes, how can we do it?

stealthycoin:
Not sure what you are asking exactly. Chalice just deploys python code to lambda and does all the configuration wiring for you to other services. Any file uploading in your application is just a python code question unless I am misunderstanding?

stealthycoin:
Ahh got it. Yea that makes sense as something to work toward.

PR #1216, opened 2019-08-19
Add multipart file upload support

dmitry-kiselev-light-it:
*Issue #, if available:*
https://github.com/aws/chalice/issues/1021
*Description of changes:*
Add `files` property to the Request object. These changes allow uploading one or multiple files using `multipart/form-data` request.


stealthycoin:
@dmitry-kiselev-light-it thanks for the PR. I made some changes, mainly switching to use the `email` interface as it is not slated for removal like the `cgi` module. There are some changes coming to the email module in 3.10, mainly the only affected parts are that `email.message.Message` is getting replaced with `email.message.EmailMessage`. From what I can tell though the part of the interface I use on the `Message` object is shared between the two so there should be no breaking changes and we should be forwards compatible

I also added an integration test and moved the differences between python 2/3 up to the `try:except:` where we handle the other differences.

It would be nice to follow this PR with one to add support more generally for `multipart/form-data` since this only supports files. Should be relatively easy to add after this is accepted.

It looks like codecov is just complaining about the new branch that differs between python versions, so I don't think its correct.

chalice/app.py
```diff
@@ -397,6 +414,30 @@ def json_body(self):
...
+        if content_type.startswith('multipart'):
+            if self._files is None:
+                self._files = self._generate_files(
+                    content_type, self.raw_body
+                )
+            return self._files
+
+    def _generate_files(self, content_type, raw_body):
+        full_content = b'Content-Type: ' + \
+            content_type.encode('utf-8') + \
```
jamesls: This doesn't look like the right content type.  We're plumbing through the content type of the form data, not the content type of the file.

On second thought, why are we adding an explicit content type header?  That seems like the wrong thing to do.  If there's one available for the file we should use it.

chalice/app.py
```diff
@@ -397,6 +414,30 @@ def json_body(self):
                     raise BadRequestError('Error Parsing JSON')
             return self._json_body
 
+    @property
+    def files(self):
+        content_type = self.headers.get('content-type')
+        if content_type.startswith('multipart'):
```
jamesls: Why is this only checking multipart?  Are there other content types besides multipart/form-data that we're intending to support here?  I didn't see anything in the tests.

chalice/app.py
```diff
@@ -350,6 +359,13 @@ def __eq__(self, other):
         return False
 
 
+class File(object):
```
jamesls: Is there a rationale for deviating from what flask has?  Seems like it would help if we use the same attribute names.

Also, I think we should at least add the content type and the filename to this object (if available), those are commonly used fields.

chalice/app.py
```diff
@@ -350,6 +359,13 @@ def __eq__(self, other):
...
+    def __init__(self, name, content):
+        self.name = name
+        self.content = content
+        self.size = len(content)
```
jamesls: Can we add a repr to this?  We should at least include the filename (and maybe content type) to help with debugging and test failures.

chalice/app.py
```diff
@@ -397,6 +414,30 @@ def json_body(self):
...
+    def files(self):
+        content_type = self.headers.get('content-type')
+        if content_type.startswith('multipart'):
+            if self._files is None:
+                self._files = self._generate_files(
+                    content_type, self.raw_body
+                )
+            return self._files
+
+    def _generate_files(self, content_type, raw_body):
```
jamesls: Can we rename this to parse files or something similar?  We're not really generating any new files we're just extracting them out from the request body.

tests/aws/test_features.py
```diff
@@ -300,6 +300,17 @@ def test_supports_put(smoke_test_app):
         response.raise_for_status()
 
 
+def test_supports_multipart_form_data(smoke_test_app):
```
jamesls: For completeness can we add a test for ascii text?

chalice/app.py
stealthycoin: What do you mean? Adding this header tells the email parse function how to parse the following content. The individual files/keys are unaffected.

iyogesh:
@jamesls @stealthycoin It'll be very useful feature for many of us. When can we expect this to be merged?

jamesls:
Looking over this PR:

* Travis is failing
* There's still feedback that hasn't been incorporated.

If someone wants to pick up this PR and run with it, I can review it again, otherwise I'm not sure when I'll be able to add this to my backlog, there's several PRs that I'm working through right now.

If other's would find this useful, please feel free to +1 the original tracking issue (#1021), which will help prioritize this request.

Issue #1125, opened 2019-05-24
support jsonschema decorator for injection into swagger gen / validation


kapilt:
i've been using a simple decorator to get nicer ergonomics on view/route function signatures and type validation. I'd like to go ahead and add something like this to chalice, with the notion that it could also be used for swagger generation. Atm chalice swagger gen is effectively wild card method parameters with the empty object definition. Ideally with this decorator it would allow offloading request validation to api gateway, though i'm a bit unclear if thats supported by the underlying SAM transform although it looks like we could just inject a separate non sam request validator resource.


```python
def schema(schema, required=None):
    """Decorator for chalice view functions to enable jsonschema                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                       
    At the moment it allows validation and keyword parameter                                                                                                                                                                                                                                                           
    passing based on extracted values.                                                                                                                                                                                                                                                                                 
    """
    # sugar syntax for skipping the declaration enclosure                                                                                                                                                                                                                                                                          
    if 'properties' not in schema:
        schema = {'properties': schema,
                  'type': 'object',
                  'additionalProperties': False}
    if '$schema' not in schema:
        schema['$schema'] = 'http://json-schema.org/schema#'
    if required and 'required' not in schema:
        schema['required'] = list(required)

    def wrapper(func):

        func.schema = schema

        @functools.wraps(func)
        def validate_invoke(*args, **kw):
            r = app.current_request
            if r.method == 'POST':
                rj = app.current_request.json_body
            elif r.method == 'GET':
                rj = app.current_request.query_params
            jsonschema.validate(rj, schema)
            kw = {}
            for k in schema['properties'].keys():
                if k in rj:
                    kw[k] = rj[k]
            return func(*args, **kw)

        return validate_invoke

    return wrapper
```


in terms of using it

```python
@app.route('/create-foo', authorizer=authorizer, methods=['POST'])
@schema({'Duration': {'type': 'number'},
                  'Budget': {'type': 'number'}}, required=('Duration'))
def create_foo(Duration, Budget=0):
    # direct parameter extraction :-)
    return {}
```

Potentially this would allow for future extension to using python typing information.

tbd on referencing common model definitions across view functions.

hjiawei:
Very interesting. I have been doing similar validation in my functions but this proposal is much better than what I have.

PR #1223, opened 2019-08-24
Modeling

stealthycoin:
Quick proposal done as a proof of concept. Below is an example app you can deploy using this branch.

Proposed interface.
==============

Two additional keys on the route decorator. `input_model` and `output_model` (maybe request/response instead?) The value of these is either `None` (default) meaning no models are applied to this route. Or  an instance of `ModelConfig`. Which is defined as such

```python
class ModelConfig(object):
    def __init__(self, model, validate=False, cls=None):
        self.model = model
        self.validate = validate
        self.cls = cls
    ...
```

Where `model` is a marshmallow schema (currently class, though I plan to change it to instance), `validate` is whether or not the model should be enforced, and `cls` is a concrete data type that can be deserialized from the input body, or serialized to a response body.

The request object has a new property `model_body` which will be an instance of `cls` created from deserializing the `json_body`. If the `input_model` `cls` is not set then it will raise an error.

An instance of the `cls` on the `output_model` can be returned from a view function and will be automatically serialized using the schema.

Validate works slightly differently for in put and output. For input it will create a body validator in API gateway and enforce validation at the API Gateway layer. Invalid requests will not trigger a lambda and will return a 400.
For output API gateway does not provide any validation (understandably), so we provide it ourselves using the marshmallow schema. Validate will be manually called, and if any errors are discovered an error will be raised and the validation errors will be logged. A 500 will be returned to the user.

Sample Project
===========

```
$ tree .
.
├── app.py
├── chalicelib
│   ├── __init__.py
│   └── models.py
└── requirements.txt

1 directory, 4 files
```

app.py
```python
from chalice import Chalice
from chalice import ModelConfig

from chalicelib import models


app = Chalice(app_name='model-test')


@app.route('/hi', methods=['POST'],
           input_model=ModelConfig(
               models.UserSchema, validate=True, cls=models.User))
def say_hi():
    user = app.current_request.model_body
    return user.make_greeting()


@app.route('/get-user', methods=['POST'],
           output_model=ModelConfig(
               models.UserSchema, cls=models.User))
def get_user():
    totally_a_user_i_got_from_a_db = models.User(
        first_name='William',
        middle_name='Henry',
        last_name='Harrison',
    )
    return totally_a_user_i_got_from_a_db

```

chalicelib/models.py
```python
import random

from marshmallow import Schema, fields


class UserSchema(Schema):
    first_name = fields.String(required=True)
    middle_name = fields.String()
    last_name = fields.String(required=True)


class User(object):
    def __init__(self, first_name, last_name, middle_name=None):
        self.first_name = first_name
        self.middle_name = middle_name
        self.last_name = last_name

    def make_greeting(self):
        greeting = random.choice(['Hi', 'Hello', 'Sup'])
        if self.middle_name is None:
            return '%s %s %s' % (greeting, self.first_name, self.last_name)
        return '%s %s %s %s' % (
            greeting, self.first_name, self.middle_name, self.last_name)

```

Sample Calls
=========

```python

### DIFF
--- a/chalice/__init__.py
+++ b/chalice/__init__.py
@@ -4,7 +4,8 @@ from chalice.app import (
     NotFoundError, ConflictError, TooManyRequestsError, Response, CORSConfig,
     CustomAuthorizer, CognitoUserPoolAuthorizer, IAMAuthorizer,
     UnprocessableEntityError, WebsocketDisconnectedError,
-    AuthResponse, AuthRoute, Cron, Rate, __version__ as chalice_version
+    AuthResponse, AuthRoute, Cron, Rate, ModelConfig,
+    __version__ as chalice_version
 )
 # We're reassigning version here to keep mypy happy.
 __version__ = chalice_version
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -899,6 +899,7 @@ class Chalice(_HandlerRegistration, DecoratorAPI):
         # This is marked as internal but is intended to be used by
         # any code within Chalice.
         self._features_used = set()
+        self.features_with_dependencies = set()
 
     def _initialize(self, env):
         if self.configure_logs:
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -354,13 +354,14 @@ class Request(object):
     """The current request from API gateway."""
 
     def __init__(self, query_params, headers, uri_params, method, body,
-                 context, stage_vars, is_base64_encoded):
+                 context, stage_vars, is_base64_encoded, input_model):
         self.query_params = None if query_params is None \
             else MultiDict(query_params)
         self.headers = CaseInsensitiveMapping(headers)
         self.uri_params = uri_params
         self.method = method
         self._is_base64_encoded = is_base64_encoded
+        self._input_model = input_model
         self._body = body
         #: The parsed JSON from the body.  This value should
         #: only be set if the Content-Type header is application/json,
@@ -396,6 +397,10 @@ class Request(object):
                 except ValueError:
                     raise BadRequestError('Error Parsing JSON')
             return self._json_body
+
+    @property
+    def model_body(self):
+        return self._input_model.deserialize(self.json_body)
 
     def to_dict(self):
         # Don't copy internal attributes.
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by jamesls** [comment]
Would you mind moving this over to an issue?  I don't mind having POC code, but I think it's more useful to separate out the proposal from the implementation.  There's a few proposals for similar ideas so I'd like to make sure we're considering all of them.

That being said, in general, I would like something that lets you define more structured inputs/outputs to rest APIs, so I'm in favor of adding something like this to chalice.

Here's some initial feedback, but I'd like to look over this in more depth.

The main sticking point for me is going to be whether or not to pull in a third party library for this.  On the one hand, marshmallow is going to provide more capabilities than what we'd be able to implement for the immediate future, but on the other hand I'm not sure how much of that we really need.

For example, the error messages in marshmallow are really detailed (and tedious to get all the details right), but we're not leveraging that and instead relying on API gateway to validate our model for us, which isn't super helpful:

```
HTTP/1.1 400 Bad Request

{
    "message": "Invalid request body"
}
```

Also, as a side note, it looks like `marshmallow-jsonschema` uses json schema draft v7 and API gateway uses v4.  Not sure if that's an issue for us now, but also seems like that could be problematic in the future.

In terms of the API for specifying models, I have some smaller feedback:

* What do you think about the decorator syntax in #1125?  I feel like that version reads a bit easier, but it also doesn't have the flexibility proposed here.
* Not a fan of the `model_body`.  Not sure if this is possible, but it seems like this would make it hard to work with IDE auto completion or type checking in general without an explicit cast of `model_body`.  I think it's worth considering separating out the schema/modeling API from the actual serialization API.
* While I get that you may want to decouple your serialization from your model (and that should always be possible), the simple case seems verbose in that you have to define your model as well as your schema.  In your examples, the same set of fields are duplicated.  I think we should have a way that consolidates the two things for the simple case where you don't need them decoupled.

kapilt:
this references #1124 but i think it meant #1125 

personally i'd have a preference for just embedding jsonschema as a decorator over having to write alot of class models on the validation. users that want class models can take a dependency on a jsonschema library that does that (https://github.com/cwacek/python-jsonschema-objects for example). using jsonschema as a decorator directly also enables using api gateway builtin validation, and avoids picking up a hard dependency and the cognitive overhead of learning a new programming model for chalice.

jamesls:
> using jsonschema as a decorator directly also enables using api gateway builtin validation,

Just wanted to clarify real quick, this PR supports builtin validation from API gateway using https://github.com/fuhrysteve/marshmallow-jsonschema.  It is nice we can avoid the lambda invocation in that case, but I mentioned in my comment above that the error messages from API gateway aren't as good as what we could generate.

Just FYI, @stealthycoin move'd this over to an issue (https://github.com/aws/chalice/issues/1224) so we don't have to tie the propsal to a specific PR.  Let's continue the discussion over there.

PR #1232, opened 2019-09-05
Remove trailing slash when using url_prefix on register_blueprint.

marcoslima:
The _HandlerRegistration._register_route have a logic that ends with a trailing slash on the route when we use the url_prefix parameter on register_blueprint.
With this PR, the trailing slash is removed and, if it results on an empty string, it will be a simple '/'.

For example, if we have a code as follows:
```python
>users_routes.py
bp_users = Blueprint(__name__)
@bp_users.route('/')
def handle_get_all_users():
    pass

@bp_users.route('/{user_id}')
def handle_get_user_by_id(user_id):
    pass

>app.py
...
from users_routes import bp_users
app = Chalice(app_name='the app')
app.register_blueprint(bp_users, url_prefix='user')
```
When deploying the above code, an error raises because the resulting route for ```/{user_id}``` is ```"/user/{user_id}/"```.
Now this last slash is removed and the deploy can be made.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -847,7 +847,7 @@ class _HandlerRegistration(object):
         url_prefix = kwargs.pop('url_prefix', None)
         if url_prefix is not None:
             path = '/'.join([url_prefix.rstrip('/'),
-                             path.strip('/')])
+                             path.strip('/')]).rstrip('/') or '/'
         methods = actual_kwargs.pop('methods', ['GET'])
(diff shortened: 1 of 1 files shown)


stealthycoin:
Hi, thanks for the PR. I didn't see this before making my own PR to fix this sorry about that. I pulled your commit into my branch (#1242) so I'm going to close this one.

Issue #1245, opened 2019-09-18
[Proposal] Model generate command line option


stealthycoin:
Introduction
------------
Proposal for #36 

Proposed Interface
-----------

A new top level command `chalice generate-models` which will print out a model of the API to stdout.

The format of the models will be OpenAPI, in the future a ``--format`` flag can be added to support multiple formats.

Example
---------

```
(chalice) ➜  tmp chalice new-project testproject
(chalice) ➜  tmp cd testproject
(chalice) ➜  testproject chalice generate-models
{
    "swagger": "2.0",
    "info": {
        "version": "1.0",
        "title": "testproject"
    },
    "schemes": [
        "https"
    ],
    "paths": {
        "/": {
            "get": {
                "consumes": [
                    "application/json"
                ],
                "produces": [
                    "application/json"
                ],
                "responses": {
                    "200": {
                        "description": "200 response",
                        "schema": {
                            "$ref": "#/definitions/Empty"
                        }
                    }
                },
                "x-amazon-apigateway-integration": {
                    "responses": {
                        "default": {
                            "statusCode": "200"
                        }
                    },
                    "uri": "arn:aws:apigateway:{region_name}:lambda:path/2015-03-31/functions/{api_handler_lambda_arn}/invocations",
                    "passthroughBehavior": "when_no_match",
                    "httpMethod": "POST",
                    "contentHandling": "CONVERT_TO_TEXT",
                    "type": "aws_proxy"
                }
            }
        }
    },
    "definitions": {
        "Empty": {
            "type": "object",
            "title": "Empty Schema"
        }
    },
    "x-amazon-apigateway-binary-media-types": [
        "application/octet-stream",
        "application/x-tar",
        "application/zip",
        "audio/basic",
        "audio/ogg",
        "audio/mp4",
        "audio/mpeg",
        "audio/wav",
        "audio/webm",
        "image/png",
        "image/jpg",
        "image/jpeg",
        "image/gif",
        "video/ogg",
        "video/mpeg",
        "video/webm"
    ]
}
```

jamesls:
I think this would be good to add, especially as we add support for modeling inputs/outputs.

Looking over our chalice CLI commands, we're not consistent with what we have now, but we should try to not make it any more inconsistent:

```
Commands:
  delete
  deploy
  gen-policy
  generate-pipeline  Generate a cloudformation template for a...
  generate-sdk
  invoke             Invoke the deployed lambda function NAME.
  local
  logs
  new-project
  package
[...]

Issue #36, opened 2016-07-12
Swagger Integration


JamieCressey:
Swagger has been hugely beneficial when developing RESTful APIs, being able to generate Swagger specs from a Chalice app would be a massive time saving.

Docstring could then be used to define Swagger specific options per endpoint to define Swagger specific config such as path variable types etc.

jamesls:
The console/API has a `GetExport` call that allows you to generate a service definition from an API.  Given that we one of those routes in `@app.route` actually correspond to an API gateway resource, I think we could probably just rely on the default behavior from API gateway and expose something like `chalice export` or `chalice api-export`.

JamieCressey:
That sounds like a reasonable starting point, and something pretty simple to implement as it. Would such a method provide the ability to include API and endpoint descriptions, parameter schemas etc though?

jonmcoe:
if its helpful, here's what my export looked like (picked YAML for readability)

```
---
swagger: "2.0"
info:
  version: "2016-10-28T17:37:52Z"
  title: "helloworld"
host: "MYADDRESS.execute-api.us-east-1.amazonaws.com"
basePath: "/dev"
schemes:
- "https"
paths:
  /:
    get:
      consumes:
      - "application/json"
      produces:
      - "application/json"
      responses:
        200:
          description: "200 response"
          schema:
            $ref: "#/definitions/Empty"
        400:
          description: "400 response"
          schema:
            $ref: "#/definitions/Empty"
        500:
          description: "500 response"
          schema:
            $ref: "#/definitions/Empty"
        401:
          description: "401 response"
          schema:
            $ref: "#/definitions/Empty"
        403:
          description: "403 response"
          schema:
            $ref: "#/definitions/Empty"
        404:
          description: "404 response"
          schema:
            $ref: "#/definitions/Empty"
        429:
          description: "429 response"
          schema:
            $ref: "#/definitions/Empty"
        409:
          description: "409 response"
          schema:
            $ref: "#/definitions/Empty"
  /hook:
    post:
      consumes:
      - "application/json"
      produces:
      - "application/json"
      responses:
        200:
          description: "200 response"
          schema:
            $ref: "#/definitions/Empty"
        400:
          description: "400 response"
          schema:
            $ref: "#/definitions/Empty"
        500:
          description: "500 response"
          schema:
            $ref: "#/definitions/Empty"
        401:
          description: "401 response"
          schema:
            $ref: "#/definitions/Empty"
        403:
          description: "403 response"
[...]

PR #1246, opened 2019-09-18
Add generate-models command to export swagger 2.0 model

stealthycoin:
Add export-models command to export swagger 2.0 model

implements #1245 #36

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -372,11 +372,11 @@ def generate_sdk(ctx, sdk_type, stage, outdir):
         raise click.Abort()
 
 
-@cli.command('export-models')
+@cli.command('generate-models')
 @click.option('--stage', default=DEFAULT_STAGE_NAME,
               help="Chalice Stage for which to generate models.")
 @click.pass_context
-def export_models(ctx, stage):
+def generate_models(ctx, stage):
     # type: (click.Context, str) -> None
     factory = ctx.obj['factory']  # type: CLIFactory
     config = factory.create_config_obj(stage)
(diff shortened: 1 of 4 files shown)


chalice/cli/__init__.py
```diff
@@ -367,6 +372,31 @@ def generate_sdk(ctx, sdk_type, stage, outdir):
...
+    graph = graph_builder.build(config, stage)
+    rest_apis = [m for m in list(graph.resources)
+                 if isinstance(m, models.RestAPI)]
+    if not rest_apis:
+        click.echo('No REST API found to generate model from.')
+        raise click.Abort()
+    swagger_generator = TemplatedSwaggerGenerator()
+    model = swagger_generator.generate_swagger(
+        config.chalice_app,
+        rest_apis[0]
```
jamesls: It looks like this rest api param is only needed for api gateway policies.  Can we simplify this and remove this param?

chalice/cli/__init__.py
```diff
@@ -367,6 +372,30 @@ def generate_sdk(ctx, sdk_type, stage, outdir):
...
 
+@cli.command('generate-models')
+@click.option('--stage', default=DEFAULT_STAGE_NAME,
+              help="Chalice Stage for which to generate models.")
+@click.pass_context
+def generate_models(ctx, stage):
+    # type: (click.Context, str) -> None
+    factory = ctx.obj['factory']  # type: CLIFactory
+    config = factory.create_config_obj(stage)
+    graph_builder = ApplicationGraphBuilder()
```
jamesls: Let's remove all the app graph builder stuff.  To check for restapis we can just check `config.chalice_app.routes` which is what the app graph builder does.

Issue #960, opened 2018-10-06
CORS not working with cognito authorization


selte1:
Hi,

CORS doesn’t seem to work properly when using cognito authorization. When you give cors=True or use your own CORS configuration the headers are added to success responses, but at least 401 and 403 responses are missing the headers completely.

Is this a bug or am I missing some configuration step?

selte1:
Hi!

I solved the issue by manually modifying "Gateway Responses" for my API in AWS console to add CORS headers to 401, 403 and other responses that result in my actual API not being reached.

Would be neat if chalice could do this for me though, or at least maybe add a mention to the docs if this part of CORS setup has to be done manually.

Cheers

dstufft:
Marking this as a feature request.

benlvovsky:
The source code for CognitoUserPoolAuthorizer I see doesn't make sense how it actually works under the hood. How can I see the actual code, extend and add CORS headers into response on failing authorization? I understand there is a feature request but I want to do a simple quick fix extending the class. The problem is I do not see the logic in available python code. Where does all that logic functionality occur?

Issue #787, opened 2018-04-07
Support more CORS options


3dw1np:
Currently, Chalice supports CORS options to inject some headers ([https://github.com/aws/chalice#tutorial-cors-support](https://github.com/aws/chalice#tutorial-cors-support)).

But there are two missing options which are available in the web console when you enable CORS through API Gateway service :

Gateway Responses
DEFAULT 4XX  DEFAULT 5XX

If they are not enabled, the error like "Missing Authentication Token" is sent with CORS header which is very annoying when you make a request from your web application.

stealthycoin:
Seems like it should be a straightforward enhancement.

3dw1np:
@stealthycoin in which release the enhancement can be added? :)

logosity:
Hello! first off, chalice is an amazing tool, thank you! In addition to saying +1 to this feature, I was wondering if anyone has found a workaround? When I try to enable the 4xx headers via the web console, they get applied, but there is a message saying "you must redeploy your API for these settings to take effect"... but then redeploying via chalice, wipes them out. :/ All suggestions appreciated! (and If I figure it out, I'll add a note here)...

logosity:
Turns out you can "redeploy" an API method from the web console - select the "actions" dropdown and then "deploy API" as seen in this example:
![this_one](https://user-images.githubusercontent.com/445960/42011922-61048212-7a5b-11e8-816f-f0356b910352.png)

3dw1np:
It seems to be not possible to set this with the awscli, I will still wait for this implementation in Chalice.

3dw1np:
Any update ?

ima747:
+1 this would be really helpful. Related: being able to specify enabling detailed cloudwatch metrics. I think that may be a limitation of some of the distribution tools, so probably a similar block to this.

crawforde:
+1 On enabling detailed cloudwatch metrics. Every new update with chalice turns off detailed metrics and I have to go manually turn them back on.

ima747:
FWIW I created a post deploy step in my pipeline which uses the CLI to force metrics back on. It's janky, but beats manually having to flip the switch every time.

crawforde:
My solution in the pipeline is to create the template with chalice package, run a python script that modifies the output json, and then deploy the template with cloudformation. The benefit of this is that all the features are there in one deploy, and I can continue to add missing chalice features in the future by modifying the python script.

crawforde:
[...]

PR #1270, opened 2019-10-09
Add api_gateway_response config

felipe-lee:
*Issue #, if available:*
Mainly addresses #787 
Could potentially address #960 if the CORS docs are updated to indicate you need to set `api_gateway_responses` in the `config.json` file (which I'd be happy to do, but not sure if that issue should be addressed separately, or in a different way).

*Description of changes:*
Added a key/value pair mapping to `config.json` that extends the swagger docs that are generated for both packaging and deploying. This mapping is specifically to enable customization of API Gateway responses such as setting the "Access-Control-Allow-Origin" header for Default 4XX/5XX responses, which is needed for CORS to work properly. 

I'm not sure if this is the best way to do this based on how the code is set up. I went through a few iterations (hence the many commits), before settling on modifying the swagger doc since that's the common piece for the deploy command as well as both types of packaging (CFN/Terraform). Let me know if you'd rather this be implemented in a different way. 

I ran all the commands mentioned in the contributing guide to check that everything still works ok. I did get a few errors in pylint, but I get those even just on the master branch so I assume they already exist or might be because of how some things are set up on my computer (e.g. watchdog import error):
image

I also get two test errors, but the same with these, I get them even on the master branch, so again I think it's based on my local setup. I'm hoping that when the CI gets triggered by this PR I can see for sure if there are errors. 

### DIFF
--- a/chalice/cli/__init__.py
+++ b/chalice/cli/__init__.py
@@ -388,6 +388,7 @@ def generate_models(ctx, stage):
     swagger_generator = TemplatedSwaggerGenerator()
     model = swagger_generator.generate_swagger(
         config.chalice_app,
+        api_gateway_responses=config.api_gateway_responses
     )
     ui = UI()
     ui.write(json.dumps(model, indent=4, cls=PlanEncoder))
--- a/chalice/config.py
+++ b/chalice/config.py
@@ -250,7 +250,7 @@ class Config(object):
 
     @property
     def api_gateway_responses(self):
-        # type: () -> Dict[str: Dict[str, Dict[str, Any]]]
+        # type: () -> Dict[str, Dict[str, Dict[str, Any]]]
         return self._chain_lookup('api_gateway_responses',
                                   varies_per_chalice_stage=True)
 
--- a/chalice/config.py
+++ b/chalice/config.py
@@ -249,6 +249,12 @@ class Config(object):
                                   varies_per_chalice_stage=True)
 
     @property
+    def api_gateway_responses(self):
+        # type: () -> Dict[str: Dict[str, Dict[str, Any]]]
+        return self._chain_lookup('api_gateway_responses',
+                                  varies_per_chalice_stage=True)
+
+    @property
     def minimum_compression_size(self):
         # type: () -> int
         return self._chain_lookup('minimum_compression_size',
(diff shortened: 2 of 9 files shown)


stealthycoin:
Thanks for the pull request @felipe-lee this features seems like it would be nice to have.

To your point about not knowing where to put it, the config or somewhere in the app code. The distinction for us is that the config file is where properties that would typically vary between stages live. Also they typically don't affect the behavior of any AWS resources. Examples being the size of your lambda function, timeout etc. Those don't really affect how the resource behaves, just how it does it.

The application code is where we expect most things related to the application itself, this feature in particular changes the default behavior of every route, so I would consider that to be something that is inherent to your application, and not something that would vary between a beta/gamma/prod stage.

I made another pr (#1276) that implements a similar feature that you should take a look at. Its similar in that it is an API gateway global property that affects how the swagger file is generated, so the implementations should be somewhat similar.

I'd be happy to review the PR for you if you'd like to update it, it might also be a good idea to open a tracking issue where we can discuss what the API should look like.

felipe-lee:
@stealthycoin ah that's a good point, I hadn't made that distinction before between the config vs app level settings. I can definitely rewrite this to follow your pr example more. 

I can open a tracking issue if that's preferred. I wasn't sure if #787 was considered a tracking issue, or if there was a more specific format to track new features.

stealthycoin:
The main two interface items I can think if to discuss are:

1) What is the name of the property hanging off of `app.api.???`.
2) What is the content of that property. It can be the raw JSON string which seems unfriendly. One of the points I didn't like about the config one is the escaped JSON string inside another JSON string. Now that it would be in code it could be something like a `Dict[str,Any]` or a class with hard requirements. I'm not super familiar with all the options you can put there, but maybe a dictionary would be the best choice.

What do you think?

felipe-lee:
Yeah, I'll open a new issue to discuss what the API should look like. I'll include a link to the values that are possible (or the ones that AWS shows as examples).

apfromiit:
+1

I use chalice as backend and angular as frontend for my apps. Recently started using chalice authorizer. For unauthorized requests, Chalice authorizer always returns status 0 (CORS error) while expected to get Http 401 or 403 status code. As of now, I can fix the issue by enabling cors with DEFAULT_4xx Gateway Response in AWS API Gateway console and redeploying the apis.
 
 Though fix might exists via **[chalice package](https://github.com/aws/chalice/issues/1282#issuecomment-696761819)**, I would like to continue using **chalice deploy**. 
 
 Having this api config to customize API Gateway response is a must need for all Angular/React/web apps to receive 4xx (Unauthorized) errors when using chalice authorizer in chalice apps.
 
I tried achieving the same via awscli and boto3 without any success.

Like to know when will this feature be available in chalice.

luis199230 (approved):

PR #1275, opened 2019-10-15
Reorganize integration tests

stealthycoin:
Tests under aws/ require credentials and will hit AWS
endpoints. Other integration tests are now run as a part of our prcheck
makefile rule.


tests/integration/test_package.py
```diff
@@ -41,6 +42,7 @@ def _get_random_package_name():
 
 class TestPackage(object):
     def test_can_package_with_dashes_in_name(self, runner, app_skeleton):
+        sys.modules.pop('app', None)
```
jamesls: We can address this for all tests by using an autouse fixture:

```
diff --git a/tests/conftest.py b/tests/conftest.py
index ee758d1..811c089 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import sys
 import zipfile
 from collections import namedtuple

@@ -11,6 +12,11 @@ def pytest_addoption(parser):
                      help='Skip slow tests')


+@fixture(autouse=True)
+def teardown_function():
+    sys.modules.pop('app', None)
+
+


```

Issue #1288, opened 2019-10-30
Fix packager.py name normalization


johnvencky:
After deployment we noticed that our application didn't work anymore:
`[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'sqlalchemy'`

Our _requirements.txt_ file has the following, so no problem there:
```
$ grep -i sqlalchemy requirements.txt 
SQLAlchemy==1.3.9
```

While debugging we noticed that the resulting deployment has sqlalchemy files in the wrong location:
```
$ unzip -l 5a76ade4f39e704f0925422520b454f2-python3.7.zip | grep -i 'sqlalchemy[^/]*/[^/]*$'
     4621  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/__init__.py
     5648  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/processors.py
     2377  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/schema.py
     3377  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/types.py
     2990  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/inspection.py
     6693  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/log.py
    16938  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/exc.py
    12721  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/interfaces.py
    51669  2019-10-29 17:49   SQLAlchemy-1.3.9.data/purelib/sqlalchemy/events.py
       11  2019-10-29 17:49   SQLAlchemy-1.3.9.dist-info/top_level.txt
      104  2019-10-29 17:49   SQLAlchemy-1.3.9.dist-info/WHEEL
     1100  2019-10-29 17:49   SQLAlchemy-1.3.9.dist-info/LICENSE
    23329  2019-10-29 17:49   SQLAlchemy-1.3.9.dist-info/RECORD
     7240  2019-10-29 17:49   SQLAlchemy-1.3.9.dist-info/METADATA
```

It looks like _packager.py_ does in fact have the code to move _purelib_ back to the root directory -- function __install_purelib_and_platlib_ -- however, it is not working as intended.

[...]

PR #1289, opened 2019-10-30
Fix packager.py name normalization.

johnvencky:
Issue #1288.

Fix packager.py name normalization.


jamesls:
Thanks for the PR.  This was fixed in https://github.com/aws/chalice/pull/1365.  Instead of not lower casing the package name, we needed to perform the same normalization to directory names when we're checking for matches.

PR #1306, opened 2019-11-19
Enable name for pure lambda function in chalice package

rajpdus:
Add function name to lambda for output template. This is required If you want to have a static name for your lambda which may be called by other client libraries.

*Issue #, if available:*

*Description of changes:*

`chalice deploy` allows one to define a name for your lambda function by simply passing name=<FN_NAME> as a parameter to your @app.lambda_function decorator. 
You could invoke the same deployed function by --<FN_NAME> from aws cli.

However, if one were to `chalice package` and bundle the template with other resources for deployment it does not retain the Function name. I figured out that during AWS template generation it does not add resource name in the cloud formation template builder.

Screenshot 2019-11-19 at 10 15 22 PM

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -187,6 +187,7 @@ class SAMTemplateGenerator(TemplateGenerator):
         lambdafunction_definition = {
             'Type': 'AWS::Serverless::Function',
             'Properties': {
+                'FunctionName': resource.function_name,
                 'Runtime': resource.runtime,
                 'Handler': resource.handler,
                 'CodeUri': resource.deployment_package.filename,
--- a/tests/unit/test_package.py
+++ b/tests/unit/test_package.py
@@ -684,6 +684,7 @@ class TestSAMTemplate(TemplateTestBase):
         assert cfn_resource == {
             'Type': 'AWS::Serverless::Function',
             'Properties': {
+                'FunctionName': 'app-dev-foo',
                 'CodeUri': 'foo.zip',
                 'Handler': 'app.app',
                 'MemorySize': 128,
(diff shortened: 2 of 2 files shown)


jamesls:
Thanks for the pull request and the nice writeup!

I can see why people would want this change, it can simplify other tools when there's a static name for your resources.

One of the issues with changing this right now is that this will replace customers lambda functions on the next deploy so it's not necessarily a backwards compatible change.  Specifically:

* Customer is using an existing version of chalice with the random suffix'd names.
* They upgrade to a version that includes this change.
* On the next deploy, cloudformation will replace the existing lambda functions with this new one because the physical name changed.

So if a customer was depending on these resources existing (e.g alarms, monitors, dashboards, additional tooling, etc.) they'd run into issues.

There's a couple options to explore:

* We could make this opt-in behavior (maybe somewhere in config.json)
* This will eventually be possible with deployment hooks, which will let people customize the deployment however they like.
* Your client tools can query the cloudformation stack for the name of the resource (e.g `aws cloudformation describe-stack-resources --stack-name Foo --logical-resource-id FN_NAME`).

Happy to discuss this more.

rajpdus:
The way I have been able to accomplish this for now is to use rely on deep merge i.e. --merge-template with an existing template I have for other resources in my application. I have additional parameters for the lambda resources and deep merge produces a template as I need it to be. Works well.

To an extent additional config params makes sense.

jamesls:
Yeah I think if we add a config param or something that allows this to be configurable I would be open to this, but I don't think we can merge this as is.  I'm going to go ahead and close out this PR, but I'm open to accepting a version that makes this opt-in.  Thanks again for the pull request.

PR #1299, opened 2019-11-14
Add Python 3.8 to Travis

gliptak:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1318, opened 2019-12-06
Py38

stealthycoin:
Add support for python 3.8.

closes #1318


stealthycoin:
The +2 misses is from existing code that got un-indented and taken out of an else statement causing the diff checker to get mad, however its not a real change.

gliptak:
I'm also working this at https://github.com/aws/chalice/pull/1299

stealthycoin:
Ah right! Ill take a look.

PR #1358, opened 2020-02-19
changes to make app.current_request thread safe

jtetrault:
- fixes race conditions that can occur when chalice is being run locally

Related Issue: #759

Wraps `app.current_request` access in a `dict` that is automatically keyed by the current thread's id, with each thread getting its own `app.current_request`.

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -16,7 +16,16 @@ from collections import namedtuple
 from six.moves.BaseHTTPServer import HTTPServer
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler
 from six.moves.socketserver import ThreadingMixIn
-from typing import List, Any, Dict, Tuple, Callable, Optional, Union  # noqa
+from typing import (
+    List,
+    Any,
+    Dict,
+    Tuple,
+    Callable,
+    Optional,
+    Union,
+    cast,
+)  # noqa
 
 from chalice.app import Chalice  # noqa
 from chalice.app import CORSConfig  # noqa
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks for the pull request!  Reading over the discussion in https://github.com/aws/chalice/issues/759, I'd prefer to make this change in the local mode code because it only affects local mode.

jtetrault:
> Thanks for the pull request! Reading over the discussion in #759, I'd prefer to make this change in the local mode code because it only affects local mode.

That makes sense. I made some changes so that only code running in local mode will be affected. Unfortunately I had to resort to a bit of a hack because the `Chalice` instance that gets passed around is instantiated in the user's `app.py` module, so there is no chance to subclass it properly.

jamesls:
Thanks for the update.  It looks like this failed the travis build so we'll need to get that resolved before we can merge.

jtetrault:
I've resolved most of the mypy issues, but I think the one remaining is actually a known bug with mypy.
https://github.com/python/mypy/issues/4125

jamesls:
Thanks for updating.  I'd still like to see the travis build passing before merging so I'm open to ideas on how to do that.  What if instead of subclassing we just had `LocalChalice` take a reference to the original `Chalice` object and override the `current_request` property but then delegating everything else via `getattr()`.  Maybe that could work?

jtetrault:
Yeah, I had considered that approach as well. I can give that a try. I am not sure if `mypy` will be alright with it.

jtetrault:
Looks like things are passing now. Thanks for the suggestion! I added a basic unittest as well to satisfy codecov.

jamesls:
Merged in 417e73e

PR #1362, opened 2020-02-25
Put back imports that weren't in module scope

### LINKED ISSUE #1355: watchdog is required
a recent change has caused chalice to fail if watchdog is not installed

jamesls:
Looks like this was added because a new change in pylint will
warn if you import stuff not at the module level.  In these cases,
this was by design.

To help mitigate this in the future I've also added a new pylint check
that will fail if you don't wrap the watchdog import in a conditional
try/except.

I'll also look into updating our CI to run tests against a venv
that doesn't have any conditional deps installed.

Fixes #1355


jamesls:
Note, didn't add a changelog entry because this only affects the unreleased master branch.

Issue #1356, opened 2020-02-17
sqlalchemy cannot be deployed from Linux


sproctor:
Packages with uppercase letters in their name, such as SQLAlchemy are not handled properly on case-sensitive file systems. The directory contained in the package has uppercase letters, but the package name is normalized to make them lowercase. Getting rid of the lower() fixes the issue.

    def _normalize_name(self, name):
        # type: (str) -> str
        # Taken directly from PEP 503
        return re.sub(r"[-_.]+", "-", name).lower()

jamesls:
Can you clarify what specifically isn't being handled properly?  Here's what I tried, which is the behavior I'd expect:

```
$ cat app.py
import sqlalchemy
from chalice import Chalice

app = Chalice(app_name='foo')


@app.route('/')
def index():
    return {'sqlalchemy': sqlalchemy.__file__}

$ cat requirements.txt
SQLAlchemy

# Note I'm on linux.
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.6 LTS
Release:	16.04
Codename:	xenial


$ chalice deploy
Creating deployment package.
Creating IAM role: foo-dev
Creating lambda function: foo-dev
Creating Rest API
Resources deployed:
  - Lambda ARN: arn:aws:lambda:us-west-2:927523630115:function:foo-dev
  - Rest API URL: https://abcd.execute-api.us-west-2.amazonaws.com/api/

$ curl https://abcd.execute-api.us-west-2.amazonaws.com/api/
{"sqlalchemy":"/var/task/sqlalchemy/__init__.py"}

```

Are you seeing different behavior?

marklynch:
I’ll see if I can put a simple replication together tomorrow to demonstrate what is happening.

sproctor:
I'm out of town right now. I have a similar example that doesn't work for me. I'll test this when I'm home.

jamesls:
Oh interesting.  I added an [integration test that packages sqlalchemy](https://github.com/aws/chalice/pull/1365), and it fails on travis: https://travis-ci.org/aws/chalice/jobs/655456628.

I was testing on an ubuntu box through VMWare on my Mac, but let me spin up an EC2 ubuntu instance and see if I can repro it there.

[...]

PR #1365, opened 2020-02-26
Add additional test for sqlalchemy packaging

jamesls:
There's been several issues related to this, so we should make sure
we can package this correctly.

*Issue #, if available:*

#1356


jamesls:
Interesting.  Looks like this is failing on travis with the same error people are running into in #1356.  Digging into this more.

jamesls:
This ended up being a bufix to get the original test I added passing.

Fixes #1356.

PR #1371, opened 2020-03-03
Store current event on Chalice object for later use and debugging

xsaero00:
*Issue #, if available:*

*Description of changes:*
Store current event on Chalice object for later use and debugging


jamesls:
Can you give me an example of how you're using this?  I'd just like to get a little bit more context.

PR #1372, opened 2020-03-04
Make integ test more robust

jamesls:
* Bump up provisioned throughput.  A sleep time of 0.05
  is a TPS of 20, so to be safe I set the throughput to 50.
* Catch and report any errors in the lambda function.  That way
  we have a better idea if anything fails.  Also assert that there
  were no errors in the lambda function.
* Fix bug where the counter would send a value of `None` instead of
  an integer.
* Retry any HTTP requests if we see a 5xx response.

### DIFF
--- a/tests/aws/test_features.py
+++ b/tests/aws/test_features.py
@@ -87,7 +87,14 @@ class SmokeTestApplication(object):
             )
         )
 
+    @retry(max_attempts=10, delay=5)
     def get_json(self, url):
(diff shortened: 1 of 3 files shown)


jamesls:
This is looking good, I've been running these tests for a while now.  I'm going to go ahead and merge this.  There might be some follow up related to the websocket tests but I think this is a good start for now.

PR #1375, opened 2020-03-12
changes to previous LocalChalice approach to fixing chalice local mul…

jtetrault:
…tithreading problems

*Issue #759

*Description of changes:*

- wrapping Chalice with LocalChalice didn't actually change how Chalice instances interacted with current_request internally (i.e. within __call__)
- this approach is much more hacky, but is required to force Chalice instances to interact with current_request in a thread-safe manner
- added a more useful unit test that actually ensures thread isolation

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -15,7 +15,7 @@ from collections import namedtuple
 
 from six.moves.BaseHTTPServer import HTTPServer
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler
-from six.moves.socketserver import ThreadingMixIn
+from six.moves.socketserver import ForkingMixIn
 from typing import (
     List,
     Any,
@@ -24,7 +24,6 @@ from typing import (
     Callable,
     Optional,
     Union,
-    cast,
 )  # noqa
 
 from chalice.app import Chalice  # noqa
@@ -56,9 +55,7 @@ class Clock(object):
 
 def create_local_server(app_obj, config, host, port):
     # type: (Chalice, Config, str, int) -> LocalDevServer
-    local_app_obj = LocalChalice(app_obj)
-    casted_local_app_obj = cast(Chalice, local_app_obj)
-    return LocalDevServer(casted_local_app_obj, config, host, port)
+    return LocalDevServer(app_obj, config, host, port)
 
 
 class LocalARNBuilder(object):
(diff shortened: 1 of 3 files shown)


jamesls:
Thanks for the PR. I agree this approach is starting to get a bit hacky.

Crazy idea, what if we switched from the threadingmixin to the forkingmixin?  It would actually be closer to how lambda executes your function anyways: only a single invocation of your app "instance" can happen a given time and there can be multiple "instances" of your app running.

cc @stealthycoin who originally added the forkingmixin.

jtetrault:
Yeah, I think that makes sense.
I went ahead and changed `LocalDevServer` to subclass `ForkingMixIn` instead of `ThreadingMixIn`, and I added a functional test that just ensures requests are being evaluated in separate processes.

jamesls:
Thanks for updating.  On second thought, I think this change might be a little more tricky because the filewatcher auto-reloader code uses threads as well and my guess is it's assuming it can communicate through threading events when to shutdown and restart.

Let me dig into this a bit and see what's going on.

jamesls:
Yeah I'd say hold off on updating this.  After digging into this code, it's quite complicated and I don't think it needs to be.  I'm going to refactor this to try and make it much simpler.

jtetrault:
Sounds good. I was trying to wrap my head around it a bit when that test was failing, and there's much more going on than I thought at first glance.

jamesls:
Just wanted to give a quick update on this.  I was able to get the POC working (https://github.com/aws/chalice/compare/master...jamesls:localchalice-fixes?expand=1) but it's not as clear of a winner as I thought it'd be.  Debugging gets a little trickier and the stat-based filewatcher needed some reworking since it's now a persistent thread.

I do think long-term the forking server is the way we should go, but given it's not as much of a drop-in replacement as I thought it'd be, I'd like to vet it more before replacing the threaded version.  The branch above also needs the tests updated since they're still using the old threading code.

I would like a fix for this to go out in the next release, so I think what I'll do for now is pull the previous changes to fix the threading code and continue to work on the forking version.

jtetrault:
Ok, sounds good. I gave that a quick once over and it makes sense, and yeah I see that changing from threading to forking causes issues with which process gets to be bound to the HTTP port, as well as the file watching stuff.

Let me know if you need one of the commits I have from the past. They _should_ all still be kicking around in my reflog 🤞.

jamesls:
Updated PR here: https://github.com/aws/chalice/pull/1375.  I believe it's a similar approach.  You have to modify the existing chalice app object and use thread local storage to track the current request.  I also added a functional test to verify thread safety.

Thanks again for working on this.

jtetrault:
Nice, yeah that looks like the same idea, but done properly (somehow I was unaware of `threading.local()` but that's perfect).

Happy to help.

PR #1377, opened 2020-03-13
Add cognito support to local

sproctor:
*Issue #, if available:*
Removed dependencies from #695 

*Description of changes:*

Add cognito support to local mode without bringing additional dependencies

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -12,6 +12,7 @@ import base64
 import functools
 import warnings
 from collections import namedtuple
+import json
 
 from six.moves.BaseHTTPServer import HTTPServer
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler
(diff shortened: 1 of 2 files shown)


chalice/local.py
```diff
@@ -701,3 +715,11 @@ def __getattr__(self, name):
     def __call__(self, *args, **kwargs):
         # type: (Any, Any) -> Any
         return self._chalice(*args, **kwargs)
+
+
+def base64url_decode(value):
```
jamesls: Why aren't we able to use `base64.urlsafe_b64decode()` directly?  I'd add a comment explaining this.  I'd also move this function into the `LocalGatewayAuthorizer` class.  We can always extract it out later if other parts of the code need it.

sproctor: I basically copied this directly from pyjwt, adjusted to work with strings instead of bytes. It looks weird to me to put this in LocalGatewayAuthorizer, but as you wish. I'll rewrite it to be clearer. I'll have a new version ready later if I have time.

sproctor:
I've added a unit test and tried to make the logic a bit clearer. Thanks.

jamesls:
Rebased/merged in ff503a1.

PR #1390, opened 2020-04-06
Make current_request thread safe in local mode

### LINKED ISSUE #759: app.current_request is not thread safe

jamesls:
This builds on https://github.com/aws/chalice/pull/1375
by ensuring we swap out the current request on the instance
of the running app.

I've also added a functional test to check for thread safety.

Long term, we should move to the forkingmixin, but it's too
drastic of a change to swap out for a bug fix so it will need
more vetting before we can switch over.

Fixes #759.

### DIFF
--- a/.changes/next-release/36408038949-bugfix-local-5121.json
+++ b/.changes/next-release/36408038949-bugfix-local-5121.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "local",
+  "description": "Make ``current_request`` thread safe in local mode (#759)"
+}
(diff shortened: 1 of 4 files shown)


jamesls:
Cool, looks like everything's passing now.  Going to go ahead and merge.

Issue #1250, opened 2019-09-23
Make blueprints no longer an experimental API.


jamesls:
Blueprints were added in 1.7.0 (Feb 2019).  In that time, there haven't been any issues for blueprints that have required any API changes.  And while I don't have exact numbers, it does seem to be getting usage out in the wild.  I propose we mark this as a fully supported API.

thulasi-ram:
Just bumping up. Would appreciate some update on this. About to start a project with blueprints which is going to be a used as a template in the org. 
Given that blueprints have been live for more than a year now I may actually go ahead even if it is experimental but it would be helpful to know if it would be considered an accepted feature.

jamesls:
Given the +1s on this issue and the fact there have been no API changes to blueprints, I'm going to go ahead and mark this as an accepted feature.  I'll send a PR in a bit.

PR #1401, opened 2020-04-22
Mark blueprints as an accepted API

jamesls:
See #1250.  There has been no major feedback about the
feature that required changing the API.  It's stable and
hasn't changed from when it was first introduced.  We can
call this an accepted feature and guarantee backwards
compatibility.


jtwolgamott:
Hoping this gets merged and released soon. Blueprints being classified as "experimental" is a major blocker for using Chalice for production systems of anything more than trivial complexity.

PR #1425, opened 2020-06-02
Add support merge yaml templates

nedelkoartem:
Cloudformation supports both types of extras files - YAML and JSON. That's why we suggesting include support of extras.yaml/yml files. Below are given few points of benefit:
- YAML files can contains more complicated configurations than JSON files;
- YAML makes it easier to embed JSON string - ie step functions;
- "!Ref" works with embedded JSON strings in YAML files, but it is annoying with JSON.

Expanding functionality config templates will helps developers use more convenient file type for them. It will also add flexibility to framework for work with Cloudformation.


chalice/package.py
```diff
@@ -30,6 +34,8 @@ def create_app_packager(
     if package_format == 'cloudformation':
         build_stage = create_build_stage(
             osutils, ui, CFNSwaggerGenerator())
+        if merge_template and is_yml_template(merge_template):
```
jamesls, line 37: I think it's worth extracting out the serializer to a separate class.  That way to don't have to add conditional `is_yml_template()` logic in all the places we need to branch based on json vs. yaml.  Then we could just pass in the serializer to the template generator.

jamesls:
I rebased on top of master and added the changes I suggested (refactor the yaml/json logic to separate classes to avoid duplicating the yaml check).

I also added an additional `--template-format` argument so you can generate yaml output without having to use the `--merge-template` feature, which has been requested by customers in the past.  I figure if we're already going to pull in a yaml library as part of this change we might as well add this arg.

/cc @kyleknap @stealthycoin

chalice/cli/__init__.py
```diff
@@ -499,22 +499,30 @@ def generate_models(ctx, stage):
                     "this argument is specified, a single "
                     "zip file will be created instead. CloudFormation Only."))
 @click.option('--merge-template',
-              help=('Specify a JSON template to be merged '
+              help=('Specify a JSON or YAML template to be merged '
                     'into the generated template. This is useful '
                     'for adding resources to a Chalice template or '
                     'modify values in the template. CloudFormation Only.'))
+@click.option('--template-format', default='json',
```
stealthycoin, line 506: default is `json`, but the doc string only mentions `JSON` and `YAML`. Later on in the code it explicitly checks for `yaml` lower case. I think if you use it as documented, you will silently get JSON instead. Given that yaml and json are hardcoded later on in different modules, it should probably be normalized to lower case here.

jamesls: Feedback incorporated.

PR #1347, opened 2020-02-04
Added option to follow logs

onlyanegg:
*Issue #, if available:*
Fixes #4

*Description of changes:*
Added option in `chalice logs` to follow logs until cancelling

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1430, opened 2020-06-05
Add option to follow logs

jamesls:
This pulls in https://github.com/aws/chalice/pull/1347 and adds a few changes:

* Fixes a couple edge cases based on the code from the aws cli v2's `aws logs tail --follow` command.  Ensuring we follow next tokens when needed.
* Extracted out the following logic to the `logs` module.  It builds on top of the low level API calls and the logic was getting more involved so extracting it out to a separate class made it easier to manage.
* Added the `--since` option that matches what the AWS CLI does in order to avoid pulling in all logs for a function.

I couldn't change the default for `chalice logs`, but for the `chalice logs -f` version I defaulted to only the most recent 10 minutes worth of logs before polling.


chalice/logs.py
```diff
@@ -107,3 +127,105 @@ def retrieve_logs(self, include_lambda_messages=True, max_entries=None):
...
+    def __init__(self, client, sleep=time.sleep, poll_time=_POLL_TIME):
+        # type: (TypedAWSClient, Callable[[int], None], int) -> None
+        self._client = client
+        self._sleep = sleep
+        self._event_id_cache = defaultdict(
+            set)  # type: Dict[datetime, Set[str]]
+        self._poll_time = poll_time
+
+    def iter_log_events(self, log_group_name, options):
+        # type: (str, LogRetrieveOptions) -> Iterator[CWLogEvent]
```
kyleknap, line 164: I wonder if this makes sense to hoist this specialized logic to the options class? The option takes in `follow` as a parameter so we should be able to set 10 minutes in the past there if it is `--follow` and no `--since` was specified.

chalice/utils.py
```diff
@@ -8,13 +8,15 @@
...
+from datetime import datetime, timedelta
 import subprocess
 
 import click
 from typing import IO, Dict, List, Any, Tuple, Iterator, BinaryIO, Text  # noqa
 from typing import Optional, Union  # noqa
-from typing import MutableMapping  # noqa
+from typing import MutableMapping, Callable  # noqa
 from typing import cast  # noqa
+from botocore.utils import parse_timestamp
```
kyleknap, line 19: I'd probably say we should vendor this function or write a subset of it chalice. There is definitely no guarantees we break this in the future.

chalice/logs.py
jamesls: Yeah I can do that.  Personally I'd like to just update the default to be 10 minutes in the past for both cases.  Once your app has been deployed and running for a while showing all the logs for `chalice logs` is almost not what you want.  I worry it's going to be confusing to have different defaults between `--follow` vs. `--no-follow`.

chalice/utils.py
jamesls: Yeah initially I was going to just vendor it, but then I noticed you all added some recent fixes for windows (which also makes it a little more tedious to backport).  I'd probably miss those updates, but I suppose that's better than breaking.

chalice/logs.py
kyleknap: >> Personally I'd like to just update the default to be 10 minutes in the past for both cases.

I think I would lean to that opinion too. I always thought it was a bit jarring that the `logs` command showed all of the logs, as made it harder to figure out if any new logs were produced.

Once this flag is merged, I'd suspect most people will use the `--follow` flag for active debugging. Then, I suspect the `logs` command without `--follow` will be more for postmortems where you will likely be wanting all logs or all logs from a particular date so I can see the default of all logs making sense there. So I'm not sure if will be too big of a deal. I think if the command was named `tail` instead of `logs` it would definitely be a bit more confusing, especially if you compare it to a  Unix `tail` command.

jamesls:
@kyleknap feedback incorporated.  The backport of `parse_timestamp()` wasn't as bad as I thought.  The recent bug fixes are related to epoch timestamp parsing and I only wanted to support iso8601 so it ended up being a call to `dateutils.parser.parse()`.

Should be ready to look at again.

PR #1444, opened 2020-06-24
Add support for OAuth scopes on routes

### LINKED ISSUE #1326: Support requiring OAuth scopes for routes
When using the `CognitoUserPoolAuthorizer` authorizer, there doesn't seem to be any way to require specific OAuth scopes for routes, as is possible through the API Gateway console.  This would be a useful feature to add.  One possible API would be to add a `scopes` keyword argument to `Chalice.route()`, which allows specifying a list of OAuth scopes that are required for that endpoint.

```python
authorizer = CognitoUserPoolAuthorizer('MyAuth', ['pool_arn'])

@app.route('/admin_stuff', authorizer=authorizer, scopes=['myapp/admin'])
def admin_stuff():
    do_some_privileged_task()

@app.route('/normal_stuff', authorizer=authorizer, scopes=['myapp/user'])
def normal_stuff():
    do_normal_stuff()
```

The would result in the outputted swagger having security sections like:

```yaml
paths:
  /admin_stuff:
    get:
      ...
      security:
      - MyAuth: ['myapp/admin']
  /normal_stuff:
    get:
      ...
      security:
      - MyAuth: ['myapp/user']
```

Currently, from what I can tell from looking at the source, (I believe this is the relevant code in [swagger.py](https://github.com/aws/chalice/blob/master/chalice/deploy/swagger.py#L152)), it seems to be hardcoded to always emit an empty scopes list for the security section.

jrbeilke:
This change adds support for a new `scopes` kwarg on routes which is used to add OAuth 2.0 and OpenID Connect style scopes for access control to methods in API Gateway.

Resolves #1326 

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -479,7 +479,7 @@ class RouteEntry(object):
 
     def __init__(self, view_function, view_name, path, method,
                  api_key_required=None, content_types=None,
-                 cors=False, authorizer=None):
+                 cors=False, authorizer=None, scopes=None):
         self.view_function = view_function
         self.view_name = view_name
         self.uri_pattern = path
@@ -500,6 +500,7 @@ class RouteEntry(object):
             cors = None
         self.cors = cors
         self.authorizer = authorizer
+        self.scopes = scopes
 
     def _parse_view_args(self):
         if '{' not in self.uri_pattern:
@@ -896,6 +897,7 @@ class _HandlerRegistration(object):
             'content_types': actual_kwargs.pop('content_types',
                                                ['application/json']),
             'cors': actual_kwargs.pop('cors', self.api.cors),
+            'scopes': actual_kwargs.pop('scopes', None),
         }
         if route_kwargs['cors'] is None:
             route_kwargs['cors'] = self.api.cors
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -8,6 +8,7 @@ import json
 import traceback
 import decimal
 import base64
+import copy
 from collections import defaultdict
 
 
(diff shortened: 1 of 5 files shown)


jamesls:
I'd like to look at this PR in more depth, but quick question just in looking at the API, I wonder if `scopes` should be a separate kwarg.  It seems like scopes does nothing if you don't also pass in an `authorizer` so unless we're going to error out, it seems like someone could think they're adding scopes to a route when they're not.  Not only that, but it would only apply to a specific type of authorizer, so using something like an `IAMAuthorizer` and also specifying `scopes` wouldn't make sense.

Because scopes is always part of an authorizer, what if we added a method or something to the authorizer to also specify scopes?

Something like:

```
@app.route('/user-pools', methods=['GET'],
           authorizer=authorizer.with_scopes("https://mychaliceapp.example.com/todos.read"]), ...)
```

Open to other ideas as well, what do you think?

jrbeilke:
Good question. Since this is all being handled through Swagger for security every route with an authorizer will technically have scopes, but some authorizers such as IAM will just have an empty list.

So maybe we add a new `scopes` attribute to all authorizers with the default being an empty list, and some sort of `with_scopes` method as you've described? This would allow users to set a common list of scopes at the authorizer level for all protected routes if desired, or have more specific control to set scopes per route as needed.

Either way it does seem like we'll need some better error handling for authorizers that don't support adding scopes (ie. IAM).

jamesls:
Yeah I think that idea's growing on me.  Add `scopes` to the authorizers, along with a `with_scopes` (name TBD) method.  For certain authorizers, `with_scopes()` can raise an error saying it's not supported, but for the cognito auth, it would essentially clone itself with an updated `scopes` attribute.  I think that works for me.

Ph4rell:
Thank you for this commit !

PR #1446, opened 2020-06-30
Break up quickstart/landing page

jamesls:
Split up quickstart into separate tutorials

The original "quickstart" was basically our README
and for a while was the only documentation we had.

We've also added a lot of features since the original
README was created, and as a result that page wasn't
really a "quickstart" anymore.

So I've split out the quickstart to just show how to deploy
a basic REST API, and moved the remaining content out into
separate tutorials.

I'd also split up the websocket tutorial into two separate
tutorials, the echo server and the chat server.

For the landing page, I added two more section that include
a quickstart vid along with the a section that outlines
additional Chalice features.


jamesls:
Merged in https://github.com/aws/chalice/commit/6093e222dc4cfe44c1e0fac9d58b272093e875a9

Issue #1468, opened 2020-07-31
[proposal] Chalice Test Client


jamesls:
# Test Client Proposal

This proposal outlines improvements to Chalice that enable
users to easily test their applications.

## Goals

* Add new Chalice APIs for easily testing a Chalice application.
* Must support not only REST APIs, but all other Chalice event handlers.
  Even if all event handlers are not implemented initially, the design
  of the APIs should be finished.
* We want the API to feel familiar to python developers.  While we will
  need to adapt to Lambda, we don't want to stray too far from existing
  testing APIs that developers use in other frameworks.  This API should not
  have a steep learning curve.

## Non Goals

Part of successfully testing an app requires the ability to swap out
objects with mock/fake versions.  This is out of scope for this proposal, but
is addressed in a separate follow up proposal.  This proposal focuses on new
testing APIs through clients that allow you to invoke your lambda handlers
and rest/websocket APIs.

## Prior Work

* Test client for Chalice rest APIs: https://github.com/aws/chalice/issues/1120
  Corresponding PR: https://github.com/aws/chalice/pull/1193

* Tracking github issue for testing: https://github.com/aws/chalice/issues/289
  "Sample for writing tests for Chalice app"

The proposal mostly builds off that work and generalizes it to include
testing for all of Chalice'event handlers.

### Flask

Docs for testing flask apps: https://flask.palletsprojects.com/en/1.1.x/testing/

Flask lets you test through a context manager, the basic test looks like:

```python

def test_foo():
    with mymodule.app.test_client() as client:
        rv = client.get('/')
        assert b'my test' in rv.data

```

They also have some documentation on how to use this with pytest.
There's also a more low level context manager that lets you set
the context:

```
import flask

app = flask.Flask(__name__)

with app.test_request_context('/?name=Peter'):
    assert flask.request.path == '/'
    assert flask.request.args['name'] == 'Peter'
```

### Django

Django also has a [test client](https://docs.djangoproject.com/en/3.0/topics/testing/tools/):

```
>>> from django.test import Client
>>> c = Client()
>>> response = c.post('/login/', {'username': 'john', 'password': 'smith'})
>>> response.status_code
200
>>> response = c.get('/customer/details/')
>>> response.content
b'<!DOCTYPE html...'
```

## Specification

Most of the frameworks I looked at have a dedicated module for testing:

* Flask and Sanic includes the `test_client` directly on the app, along
  with a corresponding test module.
* Django has a `django.test` client.
[...]

PR #1469, opened 2020-07-31
Initial commit of chalice test client

jamesls:
Allow you to test all the chalice decorators
we support with the exception of websockets.

Proposal: #1468.

### DIFF
--- a/chalice/test.py
+++ b/chalice/test.py
@@ -1,0 +1,326 @@
+import os
+import json
+import base64
+import contextlib
+from types import TracebackType
+
+from typing import Optional, Type, Generator, Dict, Any, List  # noqa
+
+from chalice import Chalice  # noqa
+from chalice.config import Config
+from chalice.local import LocalGateway, LambdaContext, LocalGatewayException
+from chalice.cli.factory import CLIFactory
+
+
+class FunctionNotFoundError(Exception):
+    pass
+
+
+class Client(object):
+    def __init__(self, app, stage_name='dev', project_dir='.'):
+        # type: (Chalice, str, str) -> None
+        self._app = app
+        self._project_dir = project_dir
+        self._stage_name = stage_name
+        self._http_client = None  # type: Optional[TestHTTPClient]
+        self._events_client = None  # type: Optional[TestEventsClient]
+        self._lambda_client = None  # type: Optional[TestLambdaClient]
+        self._chalice_config_obj = None  # type: Optional[Config]
+        # We have to be careful about not passing in the CLIFactory
+        # because this is a public interface we're exposing and we don't
+        # want the CLIFactory to be part of that.
+        self._cli_factory = CLIFactory(project_dir)
+
+    @property
+    def _chalice_config(self):
+        # type: () -> Config
+        if self._chalice_config_obj is None:
+            try:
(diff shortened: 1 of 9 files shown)


docs/source/api.rst
```diff
@@ -1237,3 +1237,167 @@ Websockets
...
+
+      Invoke a Lambda function by name.  The name should match
+      the resource name of the function.  This is typically the
+      name of the python function unless an explicit ``name=``
+      kwarg is provided when registering the function.
+
+      Returns an :class:`InvokeResponse` instance.
+
+
+.. class:: TestHTTPCLient(import_name)
```
kyleknap, line 1306: `TestHTTPCLient` -> `TestHTTPClient`

chalice/test.py
```diff
@@ -0,0 +1,326 @@
...
+            'eventSourceARN': 'arn:aws:sqs:us-west-2:12345:%s' % queue_name,
+            'md5OfBody': '754ac2f7a12df38320e0c5eafd060145',
+            'messageAttributes': {},
+            'messageId': 'message-id',
+            'receiptHandle': 'receipt-handle'
+        } for body in message_bodies]
+        sqs_event = {'Records': records}
+        return sqs_event
+
+    def generate_cw_event(self, source, detail_type,
```
kyleknap, line 285: For scheduled events, are you supposed to use this method to get a sample event? I wonder if it makes sense to have a separate method for it.

docs/source/topics/testing.rst
```diff
@@ -0,0 +1,310 @@
...
+   @app.lambda_function()
+   def foo(event, context):
+       return {'hello': 'world'}
+
+   @app.lambda_function()
+   def bar(event, context):
+       return {'event': event}
+
+
+Here's how you can test these functions with the test client:
```
kyleknap, line 32: What is the file path for these tests? Is it under a `test_app.py` file? I'm wondering if this guide should help suggest where to write tests as well. I also wonder if we help give direction on how to write tests in a separate directory from the `app.py` file in the case they want a `tests` directory.

chalice/test.py
jamesls: Yeah the intent is to use this method.  It's the same thing as a cloudwatch event without any additional detail (https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html#schedule_event_type):

```
{
  "id": "53dc4d37-cffa-4f76-80c9-8b7d4a4d2eaa",
  "detail-type": "Scheduled Event",
  "source": "aws.events",
  "account": "123456789012",
  "time": "2019-10-08T16:53:06Z",
  "region": "us-east-1",
  "resources": [ "arn:aws:events:us-east-1:123456789012:rule/MyScheduledRule" ],
  "detail": {}
}
```

docs/source/topics/testing.rst
jamesls: I can update this.  It doesn't really matter where you put them, either a top level `test_app.py` or a `tests/test_app.py`.  I'll update the docs to recommend a `tests/test_app.py` along with instructions on how you might run these tests with pytest.

jamesls:
@kyleknap Feedback incorporated.

PR #1476, opened 2020-08-05
Add pytest fixtures docs for new test client

jamesls:
Follow up from original PR/blog post feedback.

cc @kyleknap @stealthycoin.


studio3104:
@jamesls Maybe as you know, there’s [a pytest plugin for Chalice ](https://pypi.org/project/pytest-chalice/) that has been published, and I implemented. Though I’m willing to add an implementation that applies the new test API, is it better to take over the authority of the library to you guys?

Issue #1494, opened 2020-08-21
Issue using dataclasses.asdict in response body


rickbhardwaj:
Hi, 

Btw, I love Chalice. I ran into an issue where I tried to have a body variable in my response as follows:

```
@app.route('/private/users', cors=CORS_CONFIG)
def get_user():
    user_dicts = [dataclasses.asdict(user) for user in users.list_users()]

    return Response(
        status_code=201,
        headers=CORS_ENABLED_HEADERS,
        body=user_dicts
    )
```

User is a `@dataclass` file. It's pretty simple: 

```
from dataclasses import dataclass
from typing import List

from chalicelib.models.Message import Message


@dataclass(frozen=True)
class User:
    user_id: str
    sms_number: str
    email_address: str
    trusted_devices: List[str]
    conversation_history: List[Message]

```

This works when I run `chalice local`. However, when I run `chalice deploy` and try to build, I get this message. 

```
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/symtable.py", line 95, in lookup
    flags = self._table.symbols[name]
KeyError: 'user_dicts'
```

Full error log here: [chalice_deploy_err.txt](https://github.com/aws/chalice/files/5106624/chalice_deploy_err.txt)

Replacing the body variable with an empty dict fixes the issue and I can run `chalice deploy`.

rickbhardwaj:
FYI I'm running `chalice 1.16.0, python 3.7.3, darwin 19.4.0`

jamesls:
I think I figured out what was causing this.  If I add a function using the same name, then Chalice will get confused because it doesn't realize the scope for the function is overriden.  So I can get it to fail if I add a `def get_user(): pass` before your sample code:

```
import dataclasses
from chalice import Chalice, Response

app = Chalice(app_name='policybug')


def get_user():
    pass


@app.route('/private/users')
def get_user():
    user_dicts = [dataclasses.asdict(user) for user in users.list_users()]
```

Working on a fix, but a quick workaround is to same the second function something different.

subssn21:
This is really weird because I use dataclasses and asdict extensively throughout my project and I have never seen this before.

I have seen similar errors if I inadvertently name two routes with the same function name, which clearly won't work.

Is the dataclass in your app.py or a separate file in chalielib?

jamesls:
[...]

PR #1498, opened 2020-08-24
Handle symbols with multiple (shadowed) namespaces

jamesls:
In the case where a function is defined twice, e.g:

```
def foo():
    pass

def foo():
    x = 'bar'
```

We were incorrectly looking up the corresponding symbol
table for the function.  This is because we were doing
a name lookup on the children of the parent scope and returning
as soon as we found a match.  In the example above that would
always return the symbol table for the first function, even
if we were analyzing the second `foo` function.  This would
result in errors because we aren't able to find symbols we expect
to be defined in the current scope.

To fix this, I modified the child table lookup to also match up line
numbers if provided, so that way we can match scope by verifying that
the name matches in addition to matching the lineno of the FunctionDef
AST node with the lineno of the associated symtable.  As far as I can
tell, this is the best way to do this (there's a get_id() but that seems
to be the address of an internal struct type representing an AST and not
of the corresponding AST node we get from the ast module).

Fixes https://github.com/aws/chalice/issues/1494

### DIFF
--- a/chalice/analyzer.py
+++ b/chalice/analyzer.py
@@ -70,8 +70,9 @@ def get_client_calls_for_app(source_code):
 
     """
     parsed = parse_code(source_code)
-    parsed.parsed_ast = AppViewTransformer().visit(parsed.parsed_ast)
-    ast.fix_missing_locations(parsed.parsed_ast)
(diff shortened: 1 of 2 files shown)


rickbhardwaj:
Thanks so much guys!

PR #1512, opened 2020-09-09
Fix type hint

candypoplatte:
*Description of changes:*

I fixed the type hint for generate_sqs_event function.

* The "message_bodies" argument of "generate_sqs_event" function is written to receive a list of string.
* However, the type hint is a "str", not a "List[str]". So IDE displays a warning sign.
* The reason why these miswritten type hints passed the "make typecheck" command is that the string type also has a "\_\_iter\__" method.


jamesls (approved):
Thanks!

Issue #1509, opened 2020-09-04
[proposal] Chalice middleware


jamesls:
This doc proposes supporting middleware in Chalice.  Middleware allows
you to write request and response handlers that can alter
how Chalice handles inputs/outputs.  You can create general purpose handlers
that are automatically connected to every event handler in an application.

# Motivation

Modern Python web framework contains some way of writing
middleware, and support for this feature has been requested repeatedly
in Chalice.

See:

* https://github.com/aws/chalice/issues/553
* https://github.com/aws/chalice/issues/315
* https://github.com/aws/chalice/issues/1072
* https://github.com/aws/chalice/issues/1258

Middleware also falls under the theme of getting out of the user's way.
By adding extensibility into the request/response lifecycle, users
can add additional functionality that either isn't supported in Chalice
yet or will never be supported in Chalice.  See Appendix B for a list of
existing middleware in other frameworks.  There is potential for those
same type of middleware to be used in Chalice.

## Goals

When considering the design for middleware in Chalice, we want to
achieve these goals:

1. Middleware must be supported for all event handlers, not just
   for REST APIs.  These can be different APIs but all our event
   types should support some type of middleware.
2. Must be able to register middleware for specific handler types,
   e.g. only for S3 events, only for websocket APIs, REST APIs etc.
   Especially for REST APIs, there may be certain middleware that
   only make sense for a specific type of event.
3. Must be able to register a middleware for all event handlers.  This
   is especially useful for features such as tracing, logging,
   error handling etc.
4. Must support the ability to modify the request flow to short
   circuit a request handler, as well as alter the request and
   response.

## Non-Goals

To keep the initial scope small, we're not considering how we can have
a shared namespace/context where middleware can add arbitrary attributes.
This would help write shared middleware such as caching or DB management,
where the middleware could create a connection of some type and then store
it in some shared context that can be accessed within an event handler,
but is out of scope for this initial version.  I think we can use
the proposed components API (detailed elsewhere) for this functionality.

# Specification

A new `@app.middleware` decorator is added that lets you control
the request/response lifecycle.

Decorated functions take two args: an input `event` object and
a `get_response()` function that gets the event handler's response.

```python
def middleware(event_types='all'):
    ...
```

Usage:

```python
# Basic no-op handler

@app.middleware()
def my_middleware(event, get_response):
    return get_response(event)
```

For an `http` event type, the `event` object will be of type
`chalice.Request`.  For all other event handlers, the `event` object
[...]

PR #1514, opened 2020-09-10
Add support for Chalice middleware

jamesls:
Implements #1509.

A few notes on implementation specific details:

The middleware proposes a unified interface regardless of the
type of lambda function you're working with (event handlers, REST APIs,
Websockets, pure lambda, etc).  However, the underlying implementation
of these handlers were not consistent, so a large chunk of this change
was refactoring the code to be more consistent.  Notably:

* Created a new base class BaseLambdaHandler.  The idea is that all
  event handlers from now on will subclass from this class if they
  can't use an existing handler class.
* I moved all the REST API handling code from `Chalice.__call__` to
  a new RestAPIEventHandler, which implements BaseLambdaHandler.
  Because of backwards compat/legacy reason I can't actually change
  the entry point of the REST APIs to be `Chalice.__call__` so the
  best compromise we have for now is to have `Chalice.__call__`
  call out to the `RestAPIEventHandler`.
* I created a new `LambdaFunctionEvent` that matches our existing
  event handlers.  For backwards compat reasons I can't change
  the interface of `@app.lambda_function()`, but we require a
  consistent interface for middleware where your handler takes
  a single `event` arg.  Perhaps in Chalice 2.0 we can unify this
  and have everything take a single event arg.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -9,6 +9,7 @@ import traceback
 import decimal
 import base64
 import copy
+import functools
 from collections import defaultdict
 
 
@@ -1793,3 +1794,36 @@ class Blueprint(DecoratorAPI):
         raise RuntimeError("Blueprints are unable to retrieve middleware "
                            "handlers, must be registered to an app using "
                            "register_blueprint()")
+
+
+# This class is used to convert any existing/3rd party decorators
+# that work directly on lambda functions with the original signature
+# of (event, context).  By using ConvertToMiddleware you can automatically
+# apply this decorator to every lambda function in a Chalice app.
+# Example:
+#
+# Before:
+#
+# @third_part.decorator
+# def some_lambda_function(event, context): pass
+#
+# Now:
+#
+# app.register_middleware(ConvertToMiddleware(third_party.decorator))
+#
+#
+class ConvertToMiddleware(object):
+    def __init__(self, lambda_wrapper):
+        self._wrapper = lambda_wrapper
+
+    def __call__(self, event, get_response):
+        original_event, context = self._extract_original_param(event)
+        @functools.wraps(self._wrapper)
+        def wrapped(original_event, context):
+            return get_response(event)
+        return self._wrapper(wrapped)(original_event, context)
+
+    def _extract_original_param(self, event):
+        if isinstance(event, Request):
+            return event.to_original_event(), event.lambda_context
+        return event.to_dict(), event.context
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -68,7 +68,7 @@ def error_response(message, error_code, http_status_code, headers=None):
     response = Response(body=body, status_code=http_status_code,
                         headers=headers)
 
-    return response.to_dict()
+    return response
 
 
 def _matches_content_type(content_type, valid_content_types):
@@ -617,6 +617,12 @@ class WebsocketAPI(object):
 
 
 class DecoratorAPI(object):
+    def middleware(self, event_type):
+        def _middleware_wrapper(func):
+            self._register_middleware(func, event_type)
+            return func
+        return _middleware_wrapper
+
     def authorizer(self, ttl_seconds=None, execution_role=None, name=None):
         return self._create_registration_function(
             handler_type='authorizer',
@@ -725,10 +731,31 @@ class DecoratorAPI(object):
             'on_sqs_message': SQSEvent,
             'on_cw_event': CloudWatchEvent,
             'schedule': CloudWatchEvent,
+            'lambda_function': LambdaFunctionEvent,
+        }
+        middleware_mapping = {
+            'on_s3_event': 's3',
+            'on_sns_message': 'sns',
+            'on_sqs_message': 'sqs',
+            'on_cw_event': 'cloudwatch',
+            'schedule': 'scheduled',
+            'lambda_function': 'pure_lambda',
         }
         if handler_type in event_classes:
+            if handler_type == 'lambda_function':
+                # We have to wrap existing @app.lambda_function()
+                # handlers for backwards compat reasons so we can
+                # preserve the `def handler(event, context): ...`
+                # interface.  However we need a consistent interface
+                # for middleware so we have to wrap the event
+                # here.
+                user_handler = PureLambdaWrapper(user_handler)
             return EventSourceHandler(
-                user_handler, event_classes[handler_type])
+                user_handler, event_classes[handler_type],
+                middleware_handlers=self._get_middleware_handlers(
+                    event_type=middleware_mapping[handler_type],
+                )
+            )
 
         websocket_event_classes = [
             'on_ws_connect',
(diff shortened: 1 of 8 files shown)


jamesls:
I'll update the docs shortly with more examples, but the latest commit adds a converter that lets you take things that wrap/decorate a lambda function that uses the low level signature `event, context` and adapt them to Chalice middleware.  Here's an example using [AWS Lambda Powertools](https://awslabs.github.io/aws-lambda-powertools-python/) where we use this `ConvertToMiddleware` to automatically apply the tracer and logger to every lambda function in our app.  Another really cool feature is that we can bring in other people's blueprints (e.g. https://pypi.org/project/chalice-cognito-auth/) and we'll automatically apply this middleware to the lambda functions from the blueprint as well.


```python

from chalice import Chalice
from chalice.app import ConvertToMiddleware

# First, instead of using Chalice's built-in logger, we'll instead use
# the structured logger from powertools.  In addition to automatically
# injecting lambda context, let's say we also want to inject which
# route is being invoked by writing additional Chalice middleware.
from aws_lambda_powertools import Logger
from aws_lambda_powertools import Tracer

app = Chalice(app_name='chalice-powertools')


logger = Logger()
tracer = Tracer()
# This will automatically convert any decorator on a lambda function
# into middleware that will be connected to every lambda function
# in our app.  This lets us avoid decoratoring every lambda function
# with this behavior, but it also works in cases where we don't control
# the code (e.g. registering blueprints).
app.register_middleware(ConvertToMiddleware(logger.inject_lambda_context))
app.register_middleware(ConvertToMiddleware(tracer.capture_lambda_handler))

# Here we're writing Chalice specific middleware where for any HTTP
# APIs, we want to add the request uri to our structured log message.
# This shows how we can combine both Chalice-style middleware with
# other existing tools.
@app.middleware('http')
def inject_route_info(event, get_response):
    logger.structure_logs(append=True, request_uri=event.uri)
    return get_response(event)


@app.route('/')
def index():
    logger.info("In index() function, this will have a 'request_uri' key.")
    return {'hello': 'world'}


@app.route('/foo/bar')
def foobar():
    logger.info("In foobar() function")
    return {'foo': 'bar'}


@app.lambda_function()
def myfunction(event, context):
    logger.info("In myfunction().")
    tracer.put_annotation(key="PaymentStatus", value="SUCCESS")
    return {}
```

This will generate a log message like this when we make a request to our root url `/`:

```json
{
  "level": "INFO",
  "location": "index:42",
  "message": "In index() function, this will have a request_uri key.",
  "timestamp": "2020-09-11 21:20:00,579",
  "service": "service_undefined",
  "sampling_rate": 0,
  "cold_start": true,
  "function_name": "chalice-powertools-dev",
  "function_memory_size": "128",
  "function_arn": "arn:aws:lambda:us-west-2:12345:function:chalice-powertools-dev",
  "function_request_id": "f2eac779-7af3-4165-8a48-12345",
  "request_uri": "/",
  "xray_trace_id": "1-12345-e711a51ab6f0dd8d76d42d08"
}
```

This logging/tracing is automatically applied when we invoke the `myfunction` Lambda function as well, but it doesn't have the `request_uri` key because we only registered `inject_route_info` as an `http` middleware:

```
$ chalice invoke -n myfunction
...

{
  "level": "INFO",
  "location": "myfunction:54",
  "message": "In myfunction().",
  "timestamp": "2020-09-11 21:27:29,482",
  "service": "service_undefined",
  "sampling_rate": 0,
  "cold_start": true,
  "function_name": "chalice-powertools-dev-myfunction",
  "function_memory_size": "128",
  "function_arn": "arn:aws:lambda:us-west-2:12345:function:chalice-powertools-dev-myfunction",
  "function_request_id": "e8e624a1-58df-4301-a8ff-aa2611d3a1f1",
  "xray_trace_id": "1-12345-c42c88568bc6f16efaeca172"
}
```

Note: if you want to try out this example, there's a few small quirks I'm working on smoothing out first.
(cc @heitorlessa)

heitorlessa:
Looks great @jamesiri - **Does that allow a middleware to pass keyword arguments too?** or does it have to be a partial in this case?

**Example**: 

```python
import functools

# Middleware with arguments
app.register_middleware(ConvertToMiddleware(tracer.capture_lambda_handler(capture_response=False)))

# Middleware with arguments as a partial
trace_handler = functools.partial(tracer.capture_lambda_handler, capture_response=False)
app.register_middleware(ConvertToMiddleware(trace_handler))
```

---

**A note on Powertools integration** 

We expect customers to pass a `service` parameter, or an environment variable (POWERTOOLS_SERVICE_NAME) - **What's the best way to get the `app_name` from chalice?**

**Example**:

```python
from aws_lambda_powertools import Logger, Tracer, Metrics

app = Chalice(app_name='chalice-powertools')

# Payment is the name of the service - It can be one or multiple functions
# Logger will use `service` value within the `service` key in its structured logging
# Tracer will use `service` value within tracing metadata
# Metrics will use `service` value for metrics dimension
logger = Logger(service="payment")
tracer = Tracer(service="payment")
metrics = Metrics(service="payment", namespace="E-Commerce")
```

Perhaps there's a way to access Chalice's app name attribute and that can be used instead - I couldn't find it in the [API docs](https://aws.github.io/chalice/api.html). 

Worst case, customers can define a service/app name value and reuse as a variable - If they don't it's also no big deal, `POWERTOOLS_SERVICE_NAME` env var is the most commonly used amongst our customers.

jamesls:
Thanks for looking at this!

> Does that allow a middleware to pass keyword arguments too? 

You can pass kwargs as well, so this snippet will work:

```python
app.register_middleware(
    ConvertToMiddleware(tracer.capture_lambda_handler(capture_response=False))
)
```


> What's the best way to get the `app_name` from chalice?

The best way would be through the `app.app_name` attribute, but you're correct that it's not in our public API docs.  Given this attribute hasn't changed in years I don't mind making this part of our public API.

chalice/app.py
```diff
@@ -1599,7 +1785,45 @@ def _register_handler(self, handler_type, name, user_handler,
...
+#
+# @third_part.decorator
+# def some_lambda_function(event, context): pass
+#
+# Now:
+#
+# app.register_middleware(ConvertToMiddleware(third_party.decorator))
+#
+#
+class ConvertToMiddleware(object):
```
kyleknap, line 1815: Is there a particular reason this is not available in `chalice/__init__.py`? I noticed in the example that gets imported from `chalice.app` instead of `chalice` and I was more used to seeing classes being imported from there.

kyleknap (commented):
Changes look good to me. I liked the updates that you made to the original proposal with regards to adding the `register_middleware()` method.  I just had one really small comment/question.

chalice/app.py
jamesls: Good catch, this should be in `chalice/__init__.py`.

kyleknap (approved):
Nice! :ship:

PR #1527, opened 2020-09-21
Check for acceptable action when multiple actions are provided.

stevejarvis:
*Issue #, if available:*
I didn't file an Issue, just opened this PR.

*Description of changes:*
In an IAM policy, the `Action` field may be a list of actions. But when running Chalice locally, the comparison for permitted action only compares directly against the string `'execute-api:Invoke'`, not allowing for it to be included in a list, like `['execute-api:Invoke']`.

This change supports Chalice running locally to accept `['execute-api:Invoke']`.

Reference: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html

### DIFF
--- a/chalice/local.py
+++ b/chalice/local.py
@@ -382,7 +382,8 @@ class LocalGatewayAuthorizer(object):
         allow_resource_statements = []
         for statement in statements:
             if statement.get('Effect') == 'Allow' and \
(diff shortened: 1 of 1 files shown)


jamesls (approved):
Thanks, I'm going to add a test for this before merging.  Otherwise looks good.

Issue #987, opened 2018-11-16
[proposal] Add support for Kinesis and DynamoDB stream events


kgutwin:
Similar to the S3, SNS and SQS event sources implemented earlier, it would be awesome (especially for my current use case) to also support DynamoDB streams - and while we're at it, supporting Kinesis streams should also be easy.

Fortunately, the implementation in #886 (proposed in #884) should be relatively easy to adapt to support these other sources, since the Lambda event source mapping function is [the same](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.create_event_source_mapping). 

### Public API

The DynamoDB streams and Kinesis streams would be implemented nearly identically to the existing event sources - a parameterized decorator.

```python
@app.on_dynamodb_stream_event(table='mytable', batch_size=30, starting_position='TRIM_HORIZON')
def handler(event):
    for record in event:
        print(record.body)

@app.on_kinesis_stream_event(stream='mystream', batch_size=100, starting_position='LATEST')
def handler(event):
    for record in event:
        print(record.body)
```

### Backend

While we could maintain the existing SQS event source as-is, because a significant amount of functionality would be shared between the three stream sources (SQS, DynamoDB and Kinesis), it may make more sense to rename the backend "sqs_event_source" methods as "stream_event_source" methods, and then map each of the three stream decorators to the same core event source with different configs. Then you could build up the event source arn using the source as another parameter.

kyleknap:
@kgutwin Thanks for the proposal! We will get to looking at it and give you feedback on it.

Sytten:
I think this would be very useful if chalice is to become the standard for AWS lambda in python

benlvovsky:
I really need this for my project. I architected with that in mind but Chalice cannot do it at the moment. And I found it out just now. Any workaround currently available?

henkesde92:
I'm also looking forward to see this soon. Do you already have any news regarding this?

I want to implement a fairly easy REST API with AWS chalice, but this is something that I'm missing right now. So Chalice might not be the right choice when this is not on the timeline yet.

benlvovsky:
I have setup manually via AWS console trigger my lambda (easy in Chalice
*'@app.lambda_function(name='MyFunction')*) from DynamoDB stream and it
works. I think the ease of this approach is the main reason for Chalice
team is so slow on this to implement and not too many people complain about
lack of this functionality. I'd prefer to have it in code of course. When
[...]

PR #1532, opened 2020-09-28
Add support for kinesis/dynamodb stream handlers

jamesls:
Issue #987.

This implements the proposal in #987 with a few modifications.

* This handlers were renamed to `on_kinesis_message` and `on_dynamodb_message`.  I'm not tied to the names.  I was trying to stay consistent with `on_sqs_message`, `on_sns_message`, but we could also call this `on_kinesis_stream_event` as originally suggested.
* The DynamoDB stream must accept a full stream ARN, and not the table name.  There's autogenerated parts in the stream ARN that we can't calculate (explained [here](https://github.com/aws/chalice/issues/987#issuecomment-700183881)).  We could support only a table name for the `chalice deploy` case, but that's also an additive change we can add later.

Other caveats, follow up actions after this PR:

* Not all the underlying Lambda params are supported when subscribing to an event source.  These were added after the original proposal, but I'm fine with starting with this initial set of params.
* Need to update the test client to support generating these new events.

### DIFF
--- a/chalice/app.pyi
+++ b/chalice/app.pyi
@@ -167,11 +167,11 @@ class DecoratorAPI(object):
                           startition_position: str='LATEST',
                           name: Optional[str]=None) -> Callable[..., Any]: ...
 
-    def on_dynamodb_message(self,
-                            stream_arn: str,
-                            batch_size: int=100,
-                            startition_position: str='LATEST',
-                            name: Optional[str]=None) -> Callable[..., Any]: ...
+    def on_dynamodb_record(self,
+                           stream_arn: str,
+                           batch_size: int=100,
+                           startition_position: str='LATEST',
+                           name: Optional[str]=None) -> Callable[..., Any]: ...
 
     def schedule(self,
                  expression: str,
--- a/chalice/app.pyi
+++ b/chalice/app.pyi
@@ -161,11 +161,11 @@ class DecoratorAPI(object):
                        batch_size: int=1,
                        name: Optional[str]=None) -> Callable[..., Any]: ...
 
-    def on_kinesis_message(self,
-                           stream: str,
-                           batch_size: int=100,
-                           startition_position: str='LATEST',
-                           name: Optional[str]=None) -> Callable[..., Any]: ...
+    def on_kinesis_record(self,
+                          stream: str,
+                          batch_size: int=100,
+                          startition_position: str='LATEST',
+                          name: Optional[str]=None) -> Callable[..., Any]: ...
 
     def on_dynamodb_message(self,
                             stream_arn: str,
--- a/chalice/app.pyi
+++ b/chalice/app.pyi
@@ -167,6 +167,12 @@ class DecoratorAPI(object):
                            startition_position: str='LATEST',
                            name: Optional[str]=None) -> Callable[..., Any]: ...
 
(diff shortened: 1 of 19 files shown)


chalice/app.pyi
```diff
@@ -161,6 +161,18 @@ class DecoratorAPI(object):
                        batch_size: int=1,
                        name: Optional[str]=None) -> Callable[..., Any]: ...
 
+    def on_kinesis_message(self,
```
kyleknap, line 164: How about on `on_kinesis_record` and `on_dynamodb_record` instead? In the documentation that seems to be the terminology:
* Kinesis: https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html
* DynamoDB: https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html

docs/source/api.rst
```diff
@@ -326,6 +326,100 @@ Chalice
...
+          app.debug = True
+
+          @app.on_kinesis_message(stream='mystream')
+          def handler(event):
+              app.log.info("Event: %s", event.to_dict())
+              for record in event:
+                  app.log.info("Message body: %s", record.data)
+
+      :param stream: The name of the Kinesis stream you want to subscribe to.
+        This is the name of the queue, not the ARN.
```
kyleknap, line 356: Do you mean `stream` instead of `queue` here?

docs/source/api.rst
```diff
@@ -326,6 +326,100 @@ Chalice
...
+
+        * ``LATEST`` - Process new records that are added to the stream.
+        * ``TRIM_HORIZON`` - Process all records in the stream.
+
+      :param name: The name of the function to use.  This name is combined
+        with the chalice app name as well as the stage name to create the
+        entire lambda function name.  This parameter is optional.  If it is
+        not provided, the name of the python function will be used.
+
+   .. method:: on_dynamodb_message(stream, batch_size=100,
```
kyleknap, line 375: I think the parameter needs to be updated to `stream_arn` here

kyleknap (commented):
Yeah looks fine. Approach looks straightforward. I think `stream_arn` for DynamoDB makes sense for this initial launch as it covers all of the deployment backends. Like you said, we can always add to the signature in the future to simplify it. Otherwise, just had a few comments.

chalice/app.pyi
jamesls: Yeah I'm not opposed to that.  The event payload uses `Records` for all the event types (including SNS and SQS) but at least for kinesis, the API call is `PutRecord` so I think `on_kinesis_record` makes sense.  For DynamoDB, unless we're going to break completely from the existing naming schemes and do something like `on_dynamodb_change` or `on_dynamodb_activity`, etc.  I think it makes sense to stick with either `message` or `record`.

jamesls:
@kyleknap Ready to look at again.

chalice/app.pyi
kyleknap: Yeah it definitely makes sense for Kinesis. I only suggested it also for DynamoDB because that is what the docs seem to favor instead of `message` (which is used in the docs for SQS/SNS).

kyleknap (approved):
Nice! :ship:

Issue #761, opened 2018-03-22
500 error running locally and hitting a non-CORS endpoint from a different localhost port


BenMusch:
## Context

I am running `chalice local` on my machine and hitting the local chalice app (running on port 8000) from a front-end running on `localhost:8080`. I do not get this error when hitting the API from Postman or other http clients.

Here is the request made via my front-end (doesn't work):

```
curl 'http://localhost:8000/translate' -X OPTIONS \
  -H 'Access-Control-Request-Method: POST' \
  -H 'Origin: http://localhost:8080' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36' \
  -H 'Access-Control-Request-Headers: content-type' --compressed
```

Here is a request made via Postman (works):

```
curl -X POST \
  http://localhost:8000/translate \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -H 'Origin: http://localhost:8080' \
  -H 'Postman-Token: 66936da1-a61c-adb8-11c7-7e4ee0e27fd7' \
  -d '{
	"text": "My name is Ben",
	"numTranslations": "1"
}'
```

## Actual response

500 error with no content

#### Traceback

```
  File "/Users/bmuschol/.pyenv/versions/2.7.13/lib/python2.7/SocketServer.py", line 596, in process_request_thread
    self.finish_request(request, client_address)
  File "/Users/bmuschol/.pyenv/versions/2.7.13/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/Users/bmuschol/.virtualenvs/translate/lib/python2.7/site-packages/chalice/local.py", line 529, in __init__
    self, request, client_address, server)  # type: ignore
  File "/Users/bmuschol/.pyenv/versions/2.7.13/lib/python2.7/SocketServer.py", line 652, in __init__
    self.handle()
  File "/Users/bmuschol/.pyenv/versions/2.7.13/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/Users/bmuschol/.pyenv/versions/2.7.13/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
[...]

PR #1534, opened 2020-10-01
Fix local server crash on CORS request

vladimirze:
*Issue #, if available:*
#761

*Description of changes:*
Should resolve couple of situations when incoming requests causes local server to crash:
- incoming request to a view that does not have CORS enabled
- incoming request to a view with CORS enabled, but which was defined after another view on the same route but without CORS


jamesls (approved):
Thanks for the PR!

PR #1549, opened 2020-10-14
Add support for HTTP middleware catching exceptions

### LINKED ISSUE #1541: REST API exception handling through middleware
I'm using Chalice for both Lambda events and as a REST API, with a consistent error response structure used for both of those. The error response structure is different from the default one provided through ChaliceViewError as we have extra requirements to be passed down to the caller.

Right now I'm using decorators on functions to try/catch handlers and convert to a response structure, but this is becoming unwieldy as the amount of endpoints grows. Middleware seemed like a good solution for this as a "register and forget" approach, and works perfectly for Lambda events, just not REST APIs.

The main issue comes from `chalice.app.RestAPIEventHandler._get_view_function_response` , which swallows any unknown `Exception` and converts it into a `chalice.app.Response` to be returned to the caller, which happens to be my error handling middleware. With the exception information gone, there isn't really any easy way to create and return a custom error response without doing overly complicated parsing of the response body.

One solution I have found was to have my custom exceptions extend `BaseException` instead of `Exception`, bypassing the `except Exception` in  `_get_view_function_response` and allowing for it to be propagated. Extending that type doesn't seem to be the recommended way to create an exception, and my code would have to deal with two root types of exceptions, `Exception` and the custom one which extends `BaseException`.

Based on what I've seen of the code, an easy fix would be to have a custom `ChaliceUnhandledError` which is caught by `chalice.app.RestAPIEventHandler._get_view_function_response`  and re-thrown without any extra operations. This should still be backwards compatible for REST APIs and would work just fine for Lambda events, but I'm not sure about others as I haven't looked too deeply into it.

Any thoughts on the above proposal or an alternative approach to building custom error responses without having to add decorators on every function?

jamesls:
Originally proposed in #1541.

This adds a new ``ChaliceUnhandledException`` error that
can be raised from a view function.  Chalice explicitly
allows this error to propagate so that middleware for
REST APIs can catch and process this exception as needed.
This also allows you to create an app/middleware with
consistent error handling (you can create your own errors
that subclass from ``ChaliceUnhandledException``).

As part of this change, there were a few inconsistencies I fixed:

* Update the error log message to use the path instead of the view
  function.  In the context of middleware, the error may not have
  originated from the view function so the path is a more consistent
  name to use.
* Remove error handling about unknown routes.  This was old code
  from the early versions of Chalice that was dead code.  API Gateway
  will never send a request for an unknown URL (it will short circuit
  and send back a 403 forbidden).  Additionally, if this were to
  actually happen the exception we raised would propagate out of
  Lambda resulting in a 502 error response which isn't what we
  would want anyways.

Closes #1541.


kyleknap (approved):
The interface looks fine to me! :ship:

Issue #1502, opened 2020-08-28
Load custom chalice class local


muriloviana:
The feature for running chalice local is pretty nice. However, I've had a issue to run a custom chalice class local. Let's get the following case:

I need a custom method in my chalice app, so I did:
```python
from chalice import Chalice

class CustomChalice(Chalice):
    def custom_method(self):
        # do something

app = CustomChalice(app_name="custom")
```

Cool, it works perfectly in the prod stage, but when I run it local the app does not have the `custom_method` because `LocalChalice` is a subclass of `Chalice` instead of `CustomChalice`.

My suggestion is to use the current chalice app class in local mode.

jamesls:
This was originally how the local mode worked, but we had to update it in order to fix the race conditions from using a threaded server (see #759 and #1375).  I'm open to changing this to support custom subclasses as long as we're still able to preserve the thread safety from those patches for local mode.

aalvrz:
Issue can be closed it seems.

PR #1556, opened 2020-10-16
Load local custom chalice class

muriloviana:
*Issue #1502 *

*Description of changes:* Load a custom Chalice class in local environment.

Let's suppose we have the following case:
```python
from chalice import Chalice

class CustomChalice(Chalice):
    def custom_method(self):
        pass

app = CustomChalice(app_name="custom")
```

Locally the `custom_method` was not available.


jamesls (approved):
Thanks for the pull request, looks good to me!

Issue #1551, opened 2020-10-15
Support manylinux2014 wheels on newer runtimes


danni:
Newer runtimes on Amazon Linux 2 can support manylinux2014 wheels. 

Hacking this into the builder lets one use newer multilinux wheels under Python 3.8.

In `chalice.deploy.packager`:

```python
    def download_manylinux_wheels(self, abi, packages, directory):
        # type: (str, List[str], str) -> None
        """Download wheel files for manylinux for all the given packages."""
        for package in packages:
            arguments = ['--only-binary=:all:', '--no-deps', '--platform',
                         'manylinux2014_x86_64', '--implementation', 'cp',
                         '--abi', abi, '--dest', directory, package]
            self._execute('download', arguments)
```

```python
    _MANYLINUX_COMPATIBLE_PLATFORM = {'any', 'linux_x86_64',
                                      'manylinux1_x86_64',
                                      'manylinux2010_x86_64',
                                      'manylinux2014_x86_64'}
```

jamesls:
Thanks!  I'll get a few tests added for this and send a PR.

PR #1575, opened 2020-11-12
Support manylinux2014 packages

danni:
*Issue #, if available:* #1551 

*Description of changes:* Will use manylinux2014 packages. Probably needs to check what runtime is in use, but works on Python 3.8 fine.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -437,7 +437,9 @@ class DependencyBuilder(object):
     packager.
     """
     _MANYLINUX_COMPATIBLE_PLATFORM = {'any', 'linux_x86_64',
-                                      'manylinux1_x86_64'}
+                                      'manylinux1_x86_64',
+                                      'manylinux2010_x86_64',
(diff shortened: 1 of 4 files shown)


jamesls (commented):
The change looks good to me, we just need to get the tests updated to pass now that we've changed the platform to `manylinux2014_x86_64`.  I haven't experimented much with `manylinux2014` but from reading the [PEP](https://www.python.org/dev/peps/pep-0599/#backwards-compatibility-with-manylinux2010-wheels) it seems like this should be ok.

>  As a result, manylinux1 and manylinux2010 wheels are considered manylinux2014 wheels. A pip that recognizes the manylinux2014 platform tag will thus install manylinux2010 wheels for manylinux2014 platforms -- even when explicitly set -- when no manylinux2014 wheels are available.

jamesls (approved):
I've added some additional tests and verified everything's working.  I also tested on py27/py36 and everything's passing for me.  Looks like we're good to merge this change.  Thanks for the  PR!

Issue #1449, opened 2020-07-02
Chalice did not configure vpc endpoint correctly for private API


choonming:
I'm trying to create a private api while specifying the VPC endpoint. It created the private API but I do not see the vpce configuration when I try to verify it via AWS CLI and console

cli output:
```
{
    "id": "xxxxx",
    "name": "app",
    "createdDate": 1591091485,
    "version": "1.0",
    "apiKeySource": "HEADER",
    "endpointConfiguration": {
        "types": [
            "PRIVATE"
        ]
    },
    "tags": {}
}

```

config.json
```{
  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "api_gateway_endpoint_type": "PRIVATE",
      "api_gateway_endpoint_vpce": ["vpce-xxxxxx"],
      "autogen_policy": false, 
    }
  },
  "version": "2.0",
  "app_name": "app"
}
```

texlan:
I wanted to second this;  my config contains:

`  "stages": {
    "dev": {
      "api_gateway_stage": "api",
      "api_gateway_endpoint_type": "PRIVATE",
      "api_gateway_endpoint_vpce": ["vpce-067219e4a057bd528"]
    }
  }`

However upon deploying, it's deployed as PRIVATE but does not have the VPCE selected.  I'm having to use the console to manually add the VPCE to the API gateway config after every deployment.  (I'm sure I could script it but..)

choonming:
@jamesls could you look into this?

jamesls:
Sorry for the delay, taking a look now.

jamesls:
Looking into this more, how are you invoking your private API? (https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-api-test-invoke-url.html).

If you want to use the Route 53 DNS alias record, then we'd need to associate the VPC endpoint with the REST API (https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html#associate-private-api-with-vpc-endpoint), which looking over the current code, doesn't look like we do that.  Can you confirm that's what you're saying is missing?


[...]

PR #1579, opened 2020-11-13
1449 propagate VPC Endpoint registration for PRIVATE Api Gateway's

ystarikovich:
https://github.com/aws/chalice/issues/1449

*Description of changes:*

Made VPCE available for the CLI usage and for the CloudFormation / Terraform code generation

### DIFF
--- a/chalice/deploy/appgraph.py
+++ b/chalice/deploy/appgraph.py
@@ -180,7 +180,7 @@ class ApplicationGraphBuilder(object):
         return models.RestAPI(
             resource_name='rest_api',
             swagger_doc=models.Placeholder.BUILD_STAGE,
-            endpoint_type=config.api_gateway_endpoint_type,
+            endpoint=models.Endpoint(endpoint_type=config.api_gateway_endpoint_type, endpoint_vpce=config.api_gateway_endpoint_vpce),
             minimum_compression=minimum_compression,
             api_gateway_stage=config.api_gateway_stage,
             lambda_function=lambda_function,
--- a/chalice/deploy/appgraph.py
+++ b/chalice/deploy/appgraph.py
@@ -180,7 +180,10 @@ class ApplicationGraphBuilder(object):
         return models.RestAPI(
             resource_name='rest_api',
             swagger_doc=models.Placeholder.BUILD_STAGE,
-            endpoint=models.Endpoint(endpoint_type=config.api_gateway_endpoint_type, endpoint_vpce=config.api_gateway_endpoint_vpce),
+            endpoint=models.Endpoint(
+                endpoint_type=config.api_gateway_endpoint_type,
+                endpoint_vpce=config.api_gateway_endpoint_vpce
+            ),
             minimum_compression=minimum_compression,
             api_gateway_stage=config.api_gateway_stage,
             lambda_function=lambda_function,
--- a/chalice/deploy/appgraph.py
+++ b/chalice/deploy/appgraph.py
@@ -181,8 +181,8 @@ class ApplicationGraphBuilder(object):
             resource_name='rest_api',
             swagger_doc=models.Placeholder.BUILD_STAGE,
             endpoint=models.Endpoint(
-                endpoint_type=config.api_gateway_endpoint_type,
-                endpoint_vpce=config.api_gateway_endpoint_vpce
+                type=config.api_gateway_endpoint_type,
+                vpce_ids=config.api_gateway_endpoint_vpce
             ),
             minimum_compression=minimum_compression,
             api_gateway_stage=config.api_gateway_stage,
--- a/chalice/deploy/models.py
+++ b/chalice/deploy/models.py
@@ -291,8 +291,8 @@ class RestAPI(ManagedModel):
 
 @attrs
 class Endpoint(Model):
-    endpoint_type = attrib()                     # type: str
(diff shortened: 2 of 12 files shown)


ystarikovich:
This is my first PR for the Python codebase. Happy to discuss and adapt code to make this feature available.

chalice/deploy/appgraph.py
```diff
@@ -180,7 +180,10 @@ def _create_rest_api_model(self,
         return models.RestAPI(
             resource_name='rest_api',
             swagger_doc=models.Placeholder.BUILD_STAGE,
-            endpoint_type=config.api_gateway_endpoint_type,
+            endpoint=models.Endpoint(
```
jamesls, line 183: I think it might simplify things to just normalize the `vpce_ids` to always be a list, and if a string is provided in the config to just convert it `[config.api_gateway_endpoint_vpce]`.

chalice/deploy/planner.py
```diff
@@ -1227,7 +1227,7 @@ def _plan_restapi(self, resource):
                 (models.APICall(
                     method_name='import_rest_api',
                     params={'swagger_document': resource.swagger_doc,
-                            'endpoint_type': resource.endpoint_type},
+                            'endpoint': resource.endpoint},
```
jamesls, line 1230: We should convert this down to the native python types.  `resource.endpoint` is a `model.Endpoint`, and right now the `awsclient` module knows nothing about `models`.

As a result this fails for me when trying to import the rest API:

```
$ chalice deploy
...
Creating Rest API
  File "chalice/chalice/cli/__init__.py", line 656, in main
    return cli(obj={})
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/core.py", line 829, in __call__
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/core.py", line 610, in invoke
  File ".virtualenvs/chalice-37/lib/python3.7/site-packages/click/decorators.py", line 21, in new_func
  File "chalice/chalice/cli/__init__.py", line 216, in deploy
    deployed_values = d.deploy(config, chalice_stage_name=stage)
  File "chalice/chalice/deploy/deployer.py", line 376, in deploy
    return self._deploy(config, chalice_stage_name)
  File "chalice/chalice/deploy/deployer.py", line 392, in _deploy
    self._executor.execute(plan)
  File "chalice/chalice/deploy/executor.py", line 43, in execute
    self._default_handler)(instruction)
  File "chalice/chalice/deploy/executor.py", line 55, in _do_apicall
    result = method(**final_kwargs)
  File "chalice/chalice/awsclient.py", line 1069, in import_rest_api
    'endpointConfigurationTypes': endpoint['type']
TypeError: 'Endpoint' object is not subscriptable
```

chalice/deploy/planner.py
```diff
@@ -1252,8 +1252,14 @@ def _plan_restapi(self, resource):
                     '/endpointConfiguration/types/%s' % (
                         '{rest_api[endpointConfiguration][types][0]}'),
                     ['rest_api']),
-                'value': resource.endpoint_type}
-            )
+                'value': resource.endpoint.type
+            })
+            if resource.endpoint.vpce_ids:
```
jamesls, line 1257: You'll have to create a new patch operation for each VPCE, otherwise you'll get an error in botocore that the `value` has to be a string type but a list of strings was received instead.  This happens when you specify this config value as a list, e.g. `      "api_gateway_vpce_endpoint": ["vpce-abcdefg"],`

chalice/deploy/models.py
```diff
@@ -289,6 +289,12 @@ def dependencies(self):
         return cast(List[Model], resources)
 
 
+@attrs
+class Endpoint(Model):
+    type = attrib()                         # type: str
+    vpce_ids = attrib(default=None)         # type: Union[str, List[str]]
```
jamesls, line 295: (in reference to my comment in `appgraph.py`) I think we can simplify the code and types if we just require this to be `List[str]` and if a user provides a string in their config file just convert it to a list of a single string.

jamesls (changes requested):
Thanks for the PR, sorry for the delay on this.  I think this is pretty close, it just looks like there's a few issues I ran into when I tried it out which I've commented below.

chalice/deploy/appgraph.py
ystarikovich: Agree, I was biased by our production scenario and forgot that API Gateway in general allows you to attach multiple VPC endpoints to the private gateway.

Will change, thanks.

chalice/deploy/planner.py
ystarikovich: Yes, completely related to the missing fact of possibility to attach to multiple VPC endpoints.

Far3t:
@ystarikovich Hi, is this something you are working on?
Or is there a workaround as this is an issue I'm having right now?

I'm using the cli `update-rest-api` to "solve" this but a way to do it in Chalice could be really cool.

PR #1584, opened 2020-11-18
Add Allow header with list of allowed methods when returning 405 error

### LINKED ISSUE #1583: Chalice should return an Allow header with the list of allowed methods in a 405 response
Looks like the initial work was done in #160 to return an error with a 405 status code when an unsupported method is used, but Chalice should also include an `Allow` header in the response with a list of the supported methods.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow

```
The Allow header lists the set of methods supported by a resource.

This header must be sent if the server responds with a 405 Method Not Allowed status code to indicate which request methods can be used. An empty Allow header indicates that the resource allows no request methods, which might occur temporarily for a given resource, for example.
```

jrbeilke:
This change adds the `Allow` header with a list of the allowed methods in the response when returning a 405 error due to an unsupported method in the request.

Resolves #1583 


jamesls (approved):
Looks good to me.

jamesls:
Merged in 1ea01fc0

PR #1586, opened 2020-11-22
Fix #1516 (VCS package handling in the packager)

### LINKED ISSUE #1516: chalice package doesn't handle git repos referenced in requirements.txt properly
### Description
When using `chalice package` to package my app I expect it to respect the git repos referenced in the `requirements.txt` file. Instead it uses the version of the installed package to fetch a wheel. This makes it impossible to reference commits in unreleased versions or forks of published python packages.  

### Steps To Reproduce
1. Create a `requirements.txt` file that includes a reference to a git repo (see end of this bug report for an example)
1. Run `chalice package`
1. Extract the zip file
1. Check the contents of `github3/repos/commit.py`
1. Notice that the `associated_pull_requests` method isn't there
1. Swear

This was tested with Chalice 1.18.1 on Python 3.7 with the build running on CodeBuild

### Expected Result
Instead of pulling the wheel, chalice will pull the code from the git repo as instructed by the `requirements.txt` file. This would also result in the `associated_pull_requests` method being present in `github3/repos/commit.py`

### requirements.txt
```txt
-e git+ssh://git@github.com/sigmavirus24/github3.py@197173b5669b189f94c433261497cb652f1db1a9#egg=github3
semantic-version==2.8.5
```

moriyoshi:
This is a fix for #1516 .

[x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -586,8 +586,10 @@ class DependencyBuilder(object):
         # dependencies that do not have a compatible wheel file downloaded.
         # For these packages we need to explicitly try to download a
         # compatible wheel file.
-        missing_wheels = sdists.union(incompatible_wheels)
-        self._download_binary_wheels(abi, missing_wheels, directory)
+        # Just try to download incompatible wheels here, as
+        # sdist-only packages (including VCS packages) should have been
+        # downloaded in the previous step.
+        self._download_binary_wheels(abi, incompatible_wheels, directory)
 
         # Re-count the wheel files after the second download pass. Anything
         # that has an sdist but not a valid wheel file is still not going to


jamesls:
Hi, thanks for the pull request.  Before we can merge this, we need to have tests added that demonstrate that this PR fixes the underlying bug.

But at a higher level, I think this fix might cause regressions for other cases.  The sdists that's downlaoded from the `download_all_dependencies()` is the result of a platform specific `pip download` in order to capture the entire closure of dependencies.  So it's possible that there's an `sdist` only for your non-manylinux1 platform, but there's a compatible wheel file for `manylinux1`.  So in that case we still need to include the `sdists` in the list of packages we need to iterate through.

I think the best option is to try and detect if we've downloaded an "editable" package (vs. an sdist) and special case those to not try and download manylinux1 wheels for.  I'm not sure offhand if there's a reliable way to do this, so it just need to be a set best-effort heuristics.

PR #1596, opened 2020-12-09
terraform - increase upper bounds of aws provider to include 3.x

marcoceppi:
*Issue #, if available:*

*Description of changes:*

Terraform AWS 3.x provider brings in a lot of great new features and bug fixes especially when running terraform in AWS environments and loading credentials from SDK/Config. However, 3.x provider does not support 0.11 of terraform. In order to avoid pushing all chalice terraform users to 0.12+ of TF I've opted to increase the upper bounds to `< 4.x`. This will result in aws provider 2.x for 0.11 users and 3.x for 0.12+.

All changes to the [AWS provider for 3.x](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.0.0) are compatible with the configurations produced by chalice in my testing.

If the chalice maintainers feel strongly about this, I'd be happy to update the pull request to move to 3.x aws provider and increase the lowest supported version of terraform to 0.12

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -795,7 +795,7 @@ class TerraformGenerator(TemplateGenerator):
             },
             'provider': {
                 'template': {'version': '~> 2'},
-                'aws': {'version': '~> 2'},
+                'aws': {'version': '>= 2, < 4'},
                 'null': {'version': '~> 2'},
             },
             'data': {


jamesls:
Thanks for the pull request!  In general, I'm not opposed to this, and looking over the [upgrade guide](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-3-upgrade) it doesn't seem to affect anything that Chalice itself generates.

What would happen to users with existing Chalice apps that are including additional terraform modules that only work with the aws provider v2?  If they pulled in this update (and as a result switched to the v3 aws provider) would their app deployments no longer work?  If so, perhaps there's something we should do for them.

marcoceppi:
Great question, thanks for taking a look. Because we're not moving the lower bound at all. Unless a user explicitly wipes their .terraform directory it'll continue to favor the existing aws ~> 2 version. Also, since this is an open range if another module explicitly depends on aws 2.x that's is satisfied by these bounds.

Also, if users are still on 0.11 of terraform, they will only get 2.x of the aws provider as 3.x has 0.12 as minimum terraform version.

For newly generated chalice applications, and those that don't meet the previous criteria, they will move to 3.x of the provider.

jamesls (approved):
Ok, that sounds reasonable to me, thanks for clarifying.

PR #1601, opened 2020-12-10
feature - output cloudfront_zone_id and cloudfront_domain_name in terraform

marcoceppi:
*Issue #, if available:*

*Description of changes:* When using https://aws.github.io/chalice/topics/domainname.html, and terraform packaing, if the api_gateway_custom_domain `cloudfont_zone_id` and `cloudfront_domain_name` aren't provided as outputs, using the generated chalice output as a terraform module won't allow configuring route53 for the domain in question. This exposes those as outputs, similar to `EndpointURL` and `RestAPIId`

I'm uncertain how to generate a new changelog entry, there don't appear to be any documentation on the process or I'd include that as well with this as it is a feature

### DIFF
--- a/.changes/next-release/28657849715-enhancement-Terraform-39810.json
+++ b/.changes/next-release/28657849715-enhancement-Terraform-39810.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Terraform",
+  "description": "Map custom domain outputs in Terraform packaging (#1601)"
+}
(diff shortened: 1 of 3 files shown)


jamesls (commented):
Thanks for the PR.  I think we'd need to add logic for mapping the regional outputs in the case that they're using a regional endpoint, similar to what we do in the `chalice deploy` case.  Then we just need to add tests and this should be good to merge.

marcoceppi:
I'll make those updates and add some tests, thanks!

jamesls:
Oh no worries, it was pretty quick to add so I went ahead and added it.  Once CI passes we should be good to merge.

jamesls (approved):
CI's good, merging.

marcoceppi:
Thank you! Appreciate that

Issue #1602, opened 2020-12-14
Environment Variables not applied Properly


loesak:
New to Chalice so forgive me if I am mistaken.

I am using Chalice to generate Terraform configuration.

It seems that environment variables may not be applied properly in certain circumstances. Or they are and it's not intuitive and leads to misconfiguration.

I have two lambdas; one that is triggered via a CloudWatch event (cron scheduled)
```python
_CRON = os.environ.get('CRON')
app.schedule(_CRON)
def kick_off_batch_job(event):
```

and one that is triggered off an S3 event
```python
_BUCKET = os.environ.get('BUCKET')
@app.on_s3_event(bucket=_BUCKET, events=['s3:ObjectCreated:*'])
def persist_data(event: S3Event):
```

Using the following configuration:
```json
{
  "version": "2.0",
  "app_name": "my-app",
  "tags": {
    "Terraform": "yes"
  },
  "autogen_policy": false,
  "stages": {
    "dev": {
      "environment_variables": {
        "ENV": "dev"
      },
      "tags": {
        "Environment": "dev"
      },
      "lambda_functions": {
        "kick_off_batch_job": {
          "iam_policy_file": "policy/dev/kick_off_batch_job.json",
          "environment_variables": {
            "CRON": "cron(0, 6, *, *, ?, *)"
          }
        },
        "persist_data": {
          "iam_policy_file": "policy/dev/persist_data.json",
          "environment_variables": {
            "BUCKET": "my-bucket"
          }
        }
      }
    }
  }
}
```

and running the command
```shell
AWS_DEFAULT_REGION=us-east-1 chalice package --stage dev --pkg-format terraform terraform/
```

results in the following stack
```shell
Creating deployment package.
Traceback (most recent call last):
  File "/home/aloes/Development/git/chalice/.venv/lib/python3.8/site-packages/chalice/cli/__init__.py", line 656, in main
    return cli(obj={})
  File "/home/aloes/Development/git/chalice/.venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/aloes/Development/git/chalice/.venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/aloes/Development/git/chalice/.venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #1603, opened 2020-12-14
Add note about env vars being set

jamesls:
Addresses #1602


loesak (approved):

aalvrz:
Should this close #1602?

PR #1613, opened 2021-01-03
Allow custom headers in built-in authorizers

smarzola:
*Description of changes:*
Right now it is possible to set the header where the auth token will be specified only with `CustomAuthorizer`s, this PR makes it possible to change the default header also wit `Built-in` authorizers. The default is `Authorization`, preserving backward compatibility.


jamesls (approved):
Thanks for the PR, looks good to me.

Issue #692, opened 2018-01-24
Bug: `chalice deploy` fails to discover required permissions for Lambda IAM role


mileserickson:
Ordinarily, `chalice deploy` finds IAM permissions required for the application to work, builds an IAM role and prompts the user for authorization to add necessary permissions to the role:

```
$ chalice deploy
Regen deployment package.

The following actions will be added to the execution policy:

s3:GetObject

Would you like to continue?  [Y/n]:
```

However, in some situations, the required permissions are not discovered and the app is deployed without the required permissions, resulting in botocore.exceptions.ClientError (AccessDenied) when a request is received.

In the example below, the app is deployed with a role that is missing the permissions that it requires.

### MVCE
**app.py:**
```
import boto3
from chalice import Chalice

app = Chalice(app_name='bug')
s3 = boto3.client('s3')
app.debug = True


@app.route('/')
def index():
    data = get_content('foo')
    return {'data': data.decode()}


def get_content(key):
    content = s3.get_object(Bucket='this-chalice-resembles-a-grail',
                            Key=key)['Body'].read()
    return content
```

kyleknap:
I can confirm as well. I was even just doing this:
```python
from chalice import Chalice
import boto3

app = Chalice(app_name='pycascades')
ec2 = boto3.client('ec2')


@app.route('/')
def index():
    return get_regions()


def get_regions():
    return ec2.describe_regions()
```
I get prompted correctly for policy generation for this though:
```python
from chalice import Chalice
import boto3

app = Chalice(app_name='pycascades')
ec2 = boto3.client('ec2')


@app.route('/')
def index():
    return ec2.describe_regions()
```

weilu:
It looks like `s3.upload_file` is not detected at all even if it's in `index()`. Is there a workaround in the meantime? Because as it is now, even when I edit the policy manually, the next deployment will prompt me to overwrite it and there doesn't seem to be a way to skip the overwrite.

mileserickson:
@weilu You can specify a role manually and tell Chalice to stop overwriting it.

Here's an example `.chalice/config.json` file:
```json
{
  "version": "2.0",
  "app_name": "example-app",
  "manage_iam_role": false,
  "iam_role_arn": "arn:aws:iam::424242424242:role/service-role/S3_read_access",
  "stages": {
    "dev": {
      "api_gateway_stage": "api"
    }
  }
}
```

GergelyM:
[...]

PR #1628, opened 2021-01-27
Improve client call detection

jonathancary1:
Fixes Issue #692

Currently, AppViewTransformer enables detection of client calls by inserting invocations of Chalice views immediately following their definitions, e.g.

```
from chalice import Chalice
import boto3

app = Chalice(app_name='post-declaration')

@app.route('/')
def index():
    return helper()

def helper():
    return boto3.client('s3').list_buckets()
```
becomes
```
from chalice import Chalice
import boto3

app = Chalice(app_name='post-declaration')

@app.route('/')
def index():
    return helper()

index()

def helper():
    return boto3.client('s3').list_buckets()
```
which is then analyzed by SymbolTableTypeInfer for any client calls triggered from the module level.

Unfortunately, this fails to handle name resolution correctly. In the example above, `helper` will fail to be resolved.

This pull request solves this problem by removing the initial NodeTransformer pass and instead adding a second NodeVisitor pass. The second pass, which is done by AppViewTypeInfer, reuses the SymbolTable information produced by SymbolTableTypeInfer in order to properly handle name resolution and triggers a `SymbolTableTypeInfer.bind_types` call for each Chalice view.

### DIFF
--- a/chalice/analyzer.py
+++ b/chalice/analyzer.py
@@ -70,10 +70,7 @@ def get_client_calls_for_app(source_code):
 
     """
     parsed = parse_code(source_code)
-    parsed.parsed_ast = ast.fix_missing_locations(
-        AppViewTransformer().visit(parsed.parsed_ast)
-    )
-    t = SymbolTableTypeInfer(parsed)
+    t = AppViewTypeInfer(parsed)
     binder = t.bind_types()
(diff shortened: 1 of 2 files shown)


jamesls (approved):
Fantastic.  Thanks for fixing this!

Issue #1339, opened 2020-01-12
Error Messages, "Message" field question.


johntmyers:
When / Why was the class name of a Chalice Error added to the actual HTTP response sent by API GW. 

IMO it adds an unnecessary and redundant piece of information to strip out of the JSON response.  The  "Code" field already contains the class name so now we have it again the Message field.

Relevant snippet:

```
class ChaliceViewError(ChaliceError):
    STATUS_CODE = 500

    def __init__(self, msg=''):
        super(ChaliceViewError, self).__init__(
            self.__class__.__name__ + ': %s' % msg)
```

Proposal, revert back to only containing the actual user provided message and drop the redundant class name. FWIW the docs still have the "old" style of this: 

```
$ chalice deploy
$ http https://endpoint/api/cities/vancouver
HTTP/1.1 400 Bad Request

{
    "Code": "BadRequestError",
    "Message": "Unknown city 'vancouver', valid choices are: portland, seattle"
}
```

jamesls:
I had to go back quite a ways, and it looks like this changed when we switched from the old VTL based "AWS" integration method to "AWS_PROXY" back in 2017: https://github.com/aws/chalice/pull/240 

This is the first time I've heard it brought up in the past 3 years, but I do agree it's redundant.  I don't think the message is subject to any backwards compatibility issues, so I'd be fine removing it if someone wants to send a PR.

PR #1637, opened 2021-02-03
Remove exception class name from HTTP error response

aalvrz:
*Issue #, if available:* #1339 

- [x] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/docs/source/tutorials/basicrestapi.rst
+++ b/docs/source/tutorials/basicrestapi.rst
@@ -111,7 +111,7 @@ Notice what happens if we try to request a city that's not in our
 
     {
         "Code": "ChaliceViewError",
-        "Message": "ChaliceViewError: An internal server error occurred."
+        "Message": "An internal server error occurred."
     }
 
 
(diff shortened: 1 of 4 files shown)


jamesls:
Thanks for the pull request.  I think it's worth doing a quick pass through the docs for any `Message:` strings and update the error messages.

docs/source/tutorials/basicrestapi.rst
```diff
@@ -111,7 +111,7 @@ Notice what happens if we try to request a city that's not in our
 
     {
         "Code": "ChaliceViewError",
-        "Message": "ChaliceViewError: An internal server error occurred."
```
aalvrz, line 114: @jamesls This is the only instance I found in the docs that reflect the current behavior. All others interestingly do not include the class name at the beginning :thinking:

aalvrz:
@jamesls Is this still being considered for merging?

Issue #1650, opened 2021-02-10
Add support for terraform 0.14


marcoceppi:
Terraform 0.14 has been available for a little over 2 months now. Validating and raising the upper bounds for terraform packages generated by chalice would be great.

jamesls:
Closing, #1651 has been merged and released.

PR #1651, opened 2021-02-10
terraform - bump upper bounds for terraform to include 0.14

marcoceppi:
*Issue #, if available:* #1650

*Description of changes:*
This allows terraform modules generated by chalice package to be consumed and deployed with the latest release of Terraform.

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -791,7 +791,7 @@ class TerraformGenerator(TemplateGenerator):
         template = {
             'resource': {},
             'terraform': {
-                'required_version': '> 0.11.0, < 0.14.0'
+                'required_version': '> 0.11.0, < 0.15.0'
             },
             'provider': {
                 'template': {'version': '~> 2'},


kapilt:
@jamesls this feels kinda of bonkers re not supporting more liberal versions here, chalice doesn't release that often and prs are often overlooked for long swathes. 

this isn't a library dependency, its a iaac tool/app dep. chalice doesn't version against the cloudformation and sam apis which would be the equivalent semantic. terraform advances quickly and has built in support for upgrading terraform files. it seems like we should be significantly more liberal on the version spec here to avoid chalice being consistently months/years behind.

jamesls:
Sorry for the delay in this.  I was double checking there's no issues with the latest version but had a few things come up over this past month.  

@kapilt  As for the versions here, the main difference is that terraform does occasionally include changes that can change existing behavior (https://www.terraform.io/upgrade-guides/0-14.html) vs. cfn/sam which are more conservative in that regard, but I understand where you're coming from.

jamesls (approved):

kapilt:
@jamesls afaics the only issue that's come up has been chalice's trying to enforce constraints on terraform, the actual resources produced by chalice haven't had an issue with terraform changes since their inception. if a *framework* user aka application team wants to pin they can do that trivially themselves by dropping a requirements into their. They will commonly be creating additional resources themselves (dynamodb, sqs, databases, etc), they are in a much better position to define this requirement. Doing this in chalice is a constant source of issues. Ironically the one warning i have emitted by terraform right now is due to this chalice constraint being defined using a now deprecated syntax.

kapilt:
as additional commentary from a hashicorp founder ...


  |   | mitchellh 4 hours ago  [–]                                                I'm one of the HashiCorp founders.Terraform  0.11 to 0.12 is by far the most difficult of the versions to upgrade  between. I am really sorry about that. The other upgrades should be  relatively minor as long as you read and follow the upgrade guides and  upgrade one minor version at a time (0.11 => .12 => .13 etc.).  There are rough edges for very specific cases but most of our customers  were able to upgrade from 0.12 to subsequent versions same day without  issue.Breaking changes and difficult upgrades is not something we  want to do with Terraform (0.12 being a big exception as that was a  very core "reset" so to speak). The reason there have been these changes  in these recent releases is that we've been ensuring Terraform is in a  place for 1.0 that we don't have to have difficult upgrades.You can see this path happening in each release:- Terraform 0.15: state file format stability- Terraform 0.14: provider dependency lock file- Terraform 0.13: required provider block (for more deterministic providers)-  Terraform 0.12: stable JSON formats for plans/config parsing/etc.  (particularly useful for interop with things like Terraform Cloud and  3rd party tooling)This was all done to lead up to a stable 1.0 release.As  noted in the blog post, 0.15 is effectively a 1.0 pre-release so if  everything goes well, we've MADE IT. For 1.0, we'll be outlining a  number of very detailed compatibility promises which will make upgrades  much easier going forward. :)
-- | -- | --

kapilt:
and to note again the only change affecting chalice is where it attempts to limit compatibility :man_shrugging:

PR #1667, opened 2021-02-28
Remove enum-compat dependency

kylelaker:
`enum-compat` is nothing but a virtual package to install `enum34` when
the Python version is less than 3.4. See the
[project README](https://github.com/jstasiak/enum-compat).

*Issue #, if available:* N/A

*Description of changes:*

Changes the dependency to do the same thing `enum-compat` is doing.


jamesls (approved):
Thanks for updating.  IIRC, we originally had to support old versions of pip that still didn't support env markers, but we're already using them now so this should be ok to change.

PR #1686, opened 2021-03-26
expose HTTP headers and cookies in AuthRequest

vfilimonov:
*Description of changes:*

Class `AuthRequest` now exposes also HTTP headers (via an attribute) and cookies (via a property). 

This will allow to pass authentication tokens via cookies, and not only as a "Authorization" field in headers. This could be helpful when building websites on Chalice.

Sample usage (modified from the `@app.authorizer` example:

``` python
@app.authorizer()
def demo_auth(auth_request):
    token = auth_request.cookies['auth_token']

    if token == 'allow':
        return AuthResponse(routes=['/'], principal_id='user')
    else:
        return AuthResponse(routes=[], principal_id='user')
```

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -26,6 +26,7 @@ try:
     from urllib.parse import unquote_plus
     from collections.abc import Mapping
     from collections.abc import MutableMapping
+    from http.cookies import SimpleCookie
 
     unquote_str = unquote_plus
 
@@ -36,6 +37,7 @@ except ImportError:
     from urllib import unquote_plus
     from collections import Mapping
     from collections import MutableMapping
+    from Cookie import SimpleCookie
 
     # This is borrowed from botocore/compat.py
     def unquote_str(value, encoding='utf-8'):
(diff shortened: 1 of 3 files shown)


vfilimonov:
Hello @jamesls 

Tests are failing on https://github.com/aws/chalice/blob/aa3837209d86fc7de91f3c57742bd5a0eef2f31b/tests/unit/test_test.py#L128
because they expect 401 to be raised on absence of `authorization` header, and not 403 that it would throw.


I think that this could be a useful feature - to be able to use not only `authorization`, but also `cookies` for auth tokens. It would be much more elegant to have it in the same framework as the current build-in authorizer (rather than introducing a new arument of `app.route` - e.g. `cookies_authorizer=myauth`).

But this would require the authorizer to decide on where to get the token - and thus which error to throw on absence of it.

Would you be up to modifying the test suite?

vfilimonov:
It looks like it will not be that straightforward - and on AWS side things are different than in local mockup. I was confused by "Built-in Authorizers" - and assumed that they are implemented within the same lambda function. But apparently `app.authorizer` is deployed as a stand-alone lambda function and cookies will be lost when it will be called anyway. 

Or is there any way to preserve cookies on call to authorizer in AWS?

Issue #1640, opened 2021-02-05
chalice.cdk Chalice object with custom domain generates broken sam.json file


gentili:
### Steps to reproduce

1. chalice new-project
2. test
3. [CDK] Rest API with a DynamoDB table
4. edit test/infrastructure/stacks/chaliceapp.py
```python
        self.chalice = Chalice(
            self, 'ChaliceApp', source_dir=RUNTIME_SOURCE_DIR,
            stage_config={
                "api_gateway_custom_domain": {
                    "domain_name": "api.example.com",
                    "certificate_arn": "arn:aws:acm:us-east-1:1234567:certificate/00000-00000-000000",
                    },
                'environment_variables': {
                    'APP_TABLE_NAME': self.dynamodb_table.table_name
                }
            }
        )
```
5. cdk synth
### output
```
jsii.errors.JavaScriptError:
  Error: Element used in Ref expression with logical ID: 'RestAPI.Stage' not found
      at CfnParser.parseIfCfnIntrinsic (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/core/lib/cfn-parse.js:409:31)
      at CfnParser.parseValue (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/core/lib/cfn-parse.js:379:39)
      at CfnParser.parseValue (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/core/lib/cfn-parse.js:385:33)
      at Function._fromCloudFormation (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/aws-apigateway/lib/apigateway.generated.js:554:51)
      at CfnInclude.getOrCreateResource (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/cloudformation-include/lib/cfn-include.js:557:48)
      at new CfnInclude (/tmp/jsii-kernel-5AgLxw/node_modules/@aws-cdk/cloudformation-include/lib/cfn-include.js:66:18)
      at /tmp/tmpq4500vx5/lib/program.js:2700:58
      at Kernel._wrapSandboxCode (/tmp/tmpq4500vx5/lib/program.js:3128:24)
      at Kernel._create (/tmp/tmpq4500vx5/lib/program.js:2700:34)
      at Kernel.create (/tmp/tmpq4500vx5/lib/program.js:2441:29)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #1711, opened 2021-04-22
Fix to #1640 custom domain generation

mrmikemon:
Fix to issue #1640 - Custom Domain Generation breaks CDK chalice application synth and deploy.

Problem identified in issue is generating a Ref in SAM template when using CDK. Reference appears to be invalid.

The Stage element of the APIGateway in the template appears to be a simple string. The code has a reference to RestAPI.Stage. The RestAPI only appears to contain a StageName.

Thought is that the Stage provided in API gateway needs to be the same as the application stage being deployed. Instead of a reference, changed to the 'stage' from the resource.

Updated tests to verify the generated SAM templates contain the value from the test application configuration's stage.

Summary of changes:
1. Tweaked SAM template generation to output application stage and remove the reference to the RestAPI.
2. Updated tests to verify application's stage in tests.

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -733,7 +733,8 @@ class SAMTemplateGenerator(TemplateGenerator):
                 'DomainName': {'Ref': 'ApiGatewayCustomDomain'},
                 'RestApiId': {'Ref': 'RestAPI'},
                 'BasePath': domain_name.api_mapping.mount_path,
-                'Stage': {'Ref': 'RestAPI.Stage'},
+                # Workaround for domainname in cdk deployment issue #1640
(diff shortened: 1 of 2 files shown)


jamesls:
Thanks for looking into this.  It's a little weird with SAM for this property where the property name doesn't match the name you use as a ref (more details in this comment [here](https://github.com/aws/chalice/issues/1640#issuecomment-775510617)), but looking over the code, I'm not seeing (nor do I recall) a reason we have to use a ref for the stage name.  Let me double check, but otherwise I'm inclined to merge this PR.

jamesls (approved):
Looks good to me, thanks again for the fix.

PR #1720, opened 2021-05-03
Add optional queue_arn to on_sqs_message

### LINKED ISSUE #1681: Using on_sqs_message with a CDK created queue
I'm playing with the "[CDK] Rest API with a DynamoDB table" demo project and am trying to add an SQS queue+handler to it.

I have added the queue like so:
   ```
 def _create_queue(self) -> sqs.Queue:
        the_queue = sqs.Queue(
            self, 'the_queue',
            delivery_delay=cdk.Duration.seconds(15),
        )
        cdk.CfnOutput(self, 'QueueName', value=the_queue.queue_name)
        return the_queue
```

And added the handler like so:
```
@app.on_sqs_message(queue=os.environ.get('THE_QUEUE', ''))
def the_queue_handler(event):
    app.log.info("Event: %s", event.to_dict())
    for record in event:
        app.log.info("Message body: %s", record.body)
```

The queue appears to be created correctly, but registering the EventSourceMapping fails:
```
12:19:23 pm | CREATE_FAILED        | AWS::Lambda::EventSourceMapping | HandlerHandlerSqsEventSource
Invalid request provided: Invalid event source arn: The Relative Id must be non-empty: arn:aws:sqs:ap-southeast-2:341354347128: (Service: Lambda, Status Co
de: 400, Request ID: aca69971-9c5f-4529-8182-738f1e4d044a, Extended Request ID: null)
```

Have I done something wrong?

jamesls:
This allows sqs handlers to be configured with the
Chalice/CDK integration.

More context on why this is needed [here](https://github.com/aws/chalice/issues/1681#issuecomment-813509116), but this is because CDK and it's token parsing doesn't play well with the `Fn::Sub` in CFN templates when using `CfnInclude` which we use for Chalice/CDK integration.

Couple notes:

* The public facing API now has a mutually exclusive `queue`/`queue_arn` to simplify the API (instead of having a union type on the `queue` param).
* The internal API (`models.py`) uses `Union[str, QueueARN]` as I think it's a little cleaner.

Fixes #1681

### DIFF
--- a/.changes/next-release/41104421952-feature-SQS-66900.json
+++ b/.changes/next-release/41104421952-feature-SQS-66900.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "category": "SQS",
+  "description": "Add queue_arn parameter to enable CDK integration with SQS event handler (#1681)"
+}
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -720,11 +720,16 @@ class DecoratorAPI(object):
             registration_kwargs={'topic': topic}
         )
 
-    def on_sqs_message(self, queue, batch_size=1, name=None):
+    def on_sqs_message(self, queue=None, batch_size=1,
+                       name=None, queue_arn=None):
         return self._create_registration_function(
             handler_type='on_sqs_message',
             name=name,
-            registration_kwargs={'queue': queue, 'batch_size': batch_size}
+            registration_kwargs={
+                'queue': queue,
+                'queue_arn': queue_arn,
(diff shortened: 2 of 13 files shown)


stealthycoin (approved):

alisensibill:
Hi @jamesls Im facing the same issue you raised here but with the SNS topics used as arguments for lambda decorators. Would there be a similar fix for SNS topics? I see that unlike SQS we cant provide topic arns to the on_sns_message decorators, and I'm facing the same error with Fn::Sub when I use the topic name.

PR #1731, opened 2021-05-12
Support new manylinux tags and fix tests on Python 3.8.10

jamesls:
This PR contains two fixes, one for Python 3.8.10 and one for manylinux tags.
I'm combining them into a single PR because you need both to get the tests passing
again.

## Python 3.8.10 fix

Due to Chalice using relative paths when importing an app,
in Python 3.8.10, we'll use a cached version of a previous test case's
sys.path location which will result in import errors, for example see:

https://github.com/jamesls/chalice/runs/2529906754

The GitHub action for 3.8 upgraded from 3.8.9 to 3.8.10 which triggered
this issue.  It looks like this issue has been resolved upstream
(https://bugs.python.org/issue44070) but we still need a fix in Chalice
until a new version of 3.8 is released.  Probably not a bad idea to
clear the cache anyways if we're removing stuff from sys.path.

## Manylinux tag fix

The packaging tests started failing because we couldn't package numpy.
This is because they started publishing wheels using the new perennial tag
for manylinux defined in [PEP 600](https://www.python.org/dev/peps/pep-0600/)
as well as using the compressed tag sets defined in [PEP 425](https://www.python.org/dev/peps/pep-0425/).

We technically only need the compressed tag set fix which requires us to iterate
through all the possible combinations of compatibility triples, but I'm guessing
the new manylinux tags will become more common so I've also added support
for the new tags without the older aliases.

See also:

https://github.com/aws/aws-lambda-builders/issues/233
https://github.com/numpy/numpy/issues/18988

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -436,10 +436,25 @@ class DependencyBuilder(object):
     into a site-packages directory, to be included in the bundle by the
     packager.
     """
-    _MANYLINUX_COMPATIBLE_PLATFORM = {'any', 'linux_x86_64',
-                                      'manylinux1_x86_64',
-                                      'manylinux2010_x86_64',
-                                      'manylinux2014_x86_64'}
+    _ADDITIONAL_COMPATIBLE_PLATFORM = {'any', 'linux_x86_64'}
+    _MANYLINUX_LEGACY_MAP = {
+        'manylinux1_x86_64': 'manylinux_2_5_x86_64',
+        'manylinux2010_x86_64': 'manylinux_2_12_x86_64',
+        'manylinux2014_x86_64': 'manylinux_2_17_x86_64',
+    }
+
+    # Mapping of abi to glibc version in Lambda runtime.
+    _RUNTIME_GLIBC = {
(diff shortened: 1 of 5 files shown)


chalice/deploy/packager.py
```diff
@@ -456,23 +471,74 @@ def __init__(self, osutils, pip_runner=None):
...
+            perennial_tag = self._MANYLINUX_LEGACY_MAP.get(platform, platform)
+            m = re.match("manylinux_([0-9]+)_([0-9]+)_(.*)", perennial_tag)
+            if m is None:
+                return False
+            tag_major, tag_minor = [int(x) for x in m.groups()[:2]]
+            runtime_major, runtime_minor = self._RUNTIME_GLIBC.get(
+                expected_abi, self._DEFAULT_GLIBC)
+            if (tag_major, tag_minor) <= (runtime_major, runtime_minor):
+                logger.debug(
+                    "Manylinux glibc (%s, %s) is compatible with "
```
stealthycoin, line 519: Tag glibc makes more sense here.

stealthycoin (approved):
Looks good

PR #1732, opened 2021-05-13
Add support to wait for function to become Active after create and update operations

jamesls:
For more background, see: https://aws.amazon.com/blogs/compute/tracking-the-state-of-lambda-functions/
The botocore version bump was needed because of the waiter we're using to wait on function state.

NOTE: This will fail CI due to unrelated changes.  The fixes for this are in https://github.com/aws/chalice/pull/1731, once that's merged, I'll rebase again.


kyleknap (approved):
:ship:

Issue #1735, opened 2021-05-15
Race condition in CF template for custom domain mapping


chkothe:
There appears to be a race condition in the CF templates generated by `chalice package` when the `api_gateway_custom_domain` feature is used, which can cause the deployment to fail spuriously. 

**Issue**
The cause is that the `ApiCustomDomainMapping` resource can be created before the respective API stage is created, _even though_ it has a reference to `RestAPI.Stage` and therefore waits at least until the API has been created. 

However, it turns out that the stage can be created long after `RestAPI` has been marked "CREATE_COMPLETE" as the below screenshot shows (the stage is `RestAPIv1Stage`). And since the domain mapping has no explicit dependency on the `RestAPIv1Stage`, it can (and sometimes will) be created immediately after `RestAPI`, which we managed to reproduce several times in a row in our API stack. 

![CF-dependencies](https://user-images.githubusercontent.com/5318120/118346876-9c768e00-b4f3-11eb-8cfc-147ec4c265d3.png)

The stack in the above screenshot was generated with a patched version of chalice, and therefore the domain mapping is created only after the `RestAPIv1Stage`. Without the patch, it is created right after the `RestAPI`, with the usual consequences:

![reproduction](https://user-images.githubusercontent.com/5318120/118348195-84a40780-b4fd-11eb-9696-5ee78f143d13.png)

**Fix**
This appears to be a long-standing ('17) issue with the SAM transforms, and I'm adding a companion PR that implements a fairly simple fix using DependsOn from [this](https://github.com/aws/serverless-application-model/issues/192#issuecomment-406834837)  post (in the thread about the issue). The most [liked](https://github.com/aws/serverless-application-model/issues/192#issuecomment-520893111) solution in that same thread obviously doesn't always work, as the screenshot shows. The problem is acknowledged also [here](https://github.com/aws/serverless-application-model/issues/783#issuecomment-467562151) in the feature issue for custom domain name support in SAM. Note that this is may also be the root cause for Chalice #1557 (with terraform in the loop).

The name of the stage resource that's created by the Serverless transform based on this SAM [documentation](https://github.com/aws/serverless-application-model/blob/master/docs/internals/generated_resources.rst#awsserverlessapi), so one would hope that the naming convention will remain stable. In theory there could be a lower-level fix in the SAM transform itself (perhaps merging that `DependsOn` clause into anything that references `RestAPI.Stage`), but we'd rather not be blocked waiting on that.

chkothe:
[...]

PR #1737, opened 2021-05-15
Add RestAPI.Stage ref to ApiGatewayCustomDomainMapping

chkothe:
This is intended to ensure that the mapping is created only after the stage exists.

*Issue #, if available:*

The problem is explained in #1735 (the issue originally proposed another solution using `DependsOn`, but that's less clean since it refers to a logical ID that's only created by the serverless transform (albeit documented), and this fix achieves the same without doing that.

*Description of changes:*

This implements the fix proposed in this [comment](https://github.com/aws/serverless-application-model/issues/192#issuecomment-520893111), which replaces the literal stage name by a reference to `RestAPI.Stage`. It also updates 2 unit tests.

It requires some documentation hunting to figure out why this should work, but basically the serverless transform substitutes 'RestAPI.Stage' by the logical id of the stage (e.g., `RestAPImystageStage`) as described [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-generated-resources-api.html), where it's documented as a Referenceable property (also confirmed by inspecting `sam validate --debug`). That results in delaying the creation of the custom domain mapping until after the stage resource is created (fixing the race condition). Then the `Ref RestAPImystageStage` resolves to the stage name as documented [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html). 

We've tested the PR in a Chalice API stack config that otherwise will pretty much always fail and roll back, and it resolves the issue, and generates a healthy API mapping (if the resolution chain was wrong it would fail with an "invalid stage identifier" error).

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -735,7 +735,7 @@ class SAMTemplateGenerator(TemplateGenerator):
                 'DomainName': {'Ref': 'ApiGatewayCustomDomain'},
                 'RestApiId': {'Ref': 'RestAPI'},
                 'BasePath': domain_name.api_mapping.mount_path,
-                'Stage': resource.api_gateway_stage,
+                'Stage': {'Ref': 'RestAPI.Stage'},
             }
         }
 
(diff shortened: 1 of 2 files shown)


ogrodnek:
I'm running into this, too! Wondering if we can get this merged somehow? Thanks!

datashaman:
I get the following error when I run your forked version:
```
jsii.errors.JavaScriptError:
  Error: Element used in Ref expression with logical ID: 'RestAPI.Stage' not found
```

PR #1744, opened 2021-05-26
fix: add x-amazon-apigateway-endpoint-configuration if an api-gateway-endpoint-vpce is configured

marcoceppi:
*Description of changes:*

When deploying a private API Gateway and defining an api-gateway-endpoint-vpce the swagger/terraform configuration never gets updated to include those values. This branch adds the [`x-amazon-apigateway-endpoint-configuration`](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-endpoint-configuration.html)
 and configures it for the list of VPC endpoints passed in.

### DIFF
--- a/chalice/deploy/appgraph.py
+++ b/chalice/deploy/appgraph.py
@@ -168,6 +168,11 @@ class ApplicationGraphBuilder(object):
                 filename=os.path.join(
                     config.project_dir, '.chalice', policy_path))
 
+        vpce_ids = None
+        if config.api_gateway_endpoint_vpce:
+            vpce = config.api_gateway_endpoint_vpce
+            vpce_ids = [vpce] if isinstance(vpce, str) else vpce
+
         custom_domain_name = None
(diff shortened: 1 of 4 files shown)


marcoceppi:
No idea about this test failure, but I think it's related to the CI system...

marcoceppi:
I just read #1579 - that PR's model is a lot more interesting than what I've done. I'm happy to drive this PR towards that model if it's more acceptable. However, I think having it in the Swagger is a simpler way to support this change rather than adding to each TF / CF template builder.

kapilt:
yeah.. this seems like a much simpler way than what's in #1579 to achieve the same result, afaics ci is breaking because of a link checker thats getting rate limited :/

jamesls:
I agree, this is a much simpler and cleaner approach.  Rebased on latest master and merged via https://github.com/aws/chalice/pull/1864

Issue #1746, opened 2021-05-27
Interface not defined for WebsocketEvent


daka83:
Pyright error: `error: "WebsocketEvent" is unknown import symbol (reportGeneralTypeIssues)`

jamesls:
#1749 has been merged.

PR #1749, opened 2021-05-28
Add interface for WebsocketEvent

daka83:
Issue #1746 

Added missing interface for WebsocketEvent


jamesls (approved):
Looks good, thanks for adding this.  Also added a few additional attributes on the WebsocketEvent.  Will merge once CI passes.

PR #1766, opened 2021-07-15
Remove support for py27

jamesls:
This removes package dependencies for py27 as well as updates the
setup.py to not support py27.  We also no longer test on py27, and the typechecker
is set to check against py3.

See https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-python-2-7-in-aws-chalice/ for more info.


Makefile
```diff
@@ -25,7 +25,7 @@ test:
 	py.test -v $(TESTS)
 
 typecheck:
-	mypy --py2 --ignore-missing-imports --follow-imports=skip -p chalice --disallow-untyped-defs --strict-optional --warn-no-return
```
stealthycoin, line 28: Should remove the py2 rule at the bottom of the makefile as well.

jamesls:
@stealthycoin feedback incorporated.

jamesls:
Looks like the test is failing due to a timeout error, not related to this change:  /Users/runner/work/chalice/chalice/docs/source/upgrading.rst.rst:209:broken link: https://github.com/aws/chalice/issues/348 (429 Client Error: too many requests for url: https://github.com/aws/chalice/issues/348)

PR #1768, opened 2021-07-20
Return responce from websocket event source handler.

ego:
*WebSocket responce:*

Currently there is no ability to return response from user handler to client via WebSocket.
Especially for `Sec-WebSocket-Protocol` header for `$connect` route.

https://aws.amazon.com/about-aws/whats-new/2020/06/amazon-api-gateway-allows-subprotocols-on-websocket-api-connection/

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -1596,7 +1596,7 @@ class EventSourceHandler(BaseLambdaHandler):
 
 
 class WebsocketEventSourceHandler(EventSourceHandler):
-    WEBSOCKET_API_RESPONCE = {'statusCode': 200}
+    WEBSOCKET_API_RESPONSE = {'statusCode': 200}
 
     def __init__(self, func, event_class, websocket_api,
                  middleware_handlers=None):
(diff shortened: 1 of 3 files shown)


jamesls:
Thanks for the pull request.  I'd like to get some tests added before merging, but otherwise looks good.

ego:
> Thanks for the pull request. I'd like to get some tests added before merging, but otherwise looks good.

Added some tests.

jamesls:
Whoops, looks like another PR referenced the wrong PR.  Reopening, thanks for adding tests!

jamesls:
The change looks good, but I noticed that API gateway doesn't forward the body or any additional headers besides the websocket related headers back to the client, so I wonder if people might be expecting those headers to be sent back.  Although the websocket spec is pretty specific about the response it's expecting so I'm not sure what use case someone would have to send back headers besides `Sec-Websocket-Protocol`.

I think I'll just update the docs to state you can return a response but only the `Sec-Websocket-Protocol` header is sent back to the client.

jamesls (approved):
Everything looks good, looks like all the tests are passing, I'll go ahead and merge.  Thanks for the PR!

PR #1780, opened 2021-08-04
Avoid duplicate jobs on push/pull_request triggers

jamesls:


jamesls:
Looks like it worked, not sure why this syntax avoids the duplicate triggers vs. the other way, but regardless looks good to merge.

Issue #1786, opened 2021-08-17
Expand attrs requirement range to 21.* version for compatibility with latest AWS CDK versions


alexpulver:
The latest AWS CDK versions require `attrs~=21.2`, whereas Chalice latest version requires `attrs<20.4.0`.
Can the support for `attrs~=21.*` be added please?

Below is the output from pip-compile for requirements file containing both AWS CDK and Chalice packages:
```
There are incompatible versions in the resolved dependencies:
  attrs~=21.2 (from jsii==1.33.0->aws-cdk.core==1.118.0->cdk-chalice (setup.py))
  attrs<20.4.0,>=19.3.0 (from chalice==1.24.1->cdk-chalice (setup.py))
```

jamesls:
Addressed in https://github.com/aws/chalice/pull/1788.

I've set up dependabot for attrs so we should pick up new attrs releases once they're available.  I'm also looking into setting up a github action to run the CDK tests.  This ideally should have been caught by our CI.

jamesls:
I've also added https://github.com/aws/chalice/pull/1790 which will run the CDK functional tests as part of our CI process so we should be able to prevent this happening in the future.

PR #1790, opened 2021-08-18
Run CDK tests as part of the run-tests GitHub action

jamesls:
I've set this up as a separate job under this action to isolate
setting up node and the CDK to only the tests that need it.

Should prevent issues like #1786 from happening in the future.


jamesls:
Looks like the CDK specific tests have passed.  I've also updated the job to only run on ubuntu so we can keep the execution time to a minimum and not lock up the macos jobs.

PR #1822, opened 2021-10-17
Add HTTP Route Metadata Feature

weaverjess:
*Issue #, if available:* 1818 kind of? I realized after posting the feature didn't exist so I closed/removed it to put up this PR.

*Description of changes:*
This commit/PR adds the ability to add arbitrary metadata to an HTTP route which is then passed down the call-chain to middleware(s) and the handler function. This allows more complex middleware to be created that needs
route-specific data!

This allows middleware to receive route-specific information and can be used in several ways such as:
- Defining "roles" on a per-route basis which is then used within a middleware.
- Conditionally skip middleware on specific HTTP endpoints using metadata to filter within the middleware

To do this the following was done:
-  Added some conditional code on HTTP requests in `RestAPIEventHandler` for looking up a route in order to get metadata
-  Added the ability to pass "metadata" parameter to route.
- `RestAPIEventHandler`'s call function now also looks at RouteEntry and
  also injects the metadata object into the request.
- Two unit tests were added that confirm that if no parameter is handed down, None metadata attribute exists. The second case is just to confirm if something is appended, it gets added and can be seen in middlewares & end response.

The data itself can be whatever you want since it is arbitrary and is default set to None.

An example: `@app.route("/", metadata={"some":"metadata"})`

PR Check was successful and API docs have been altered in the Chalice & Request Object sections. This should be ready to roll and with some review + feedback I'd really like to get this going since!

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -14,7 +14,7 @@ import datetime
 from collections import defaultdict
 
 
-__version__ = '1.26.1'
+__version__ = '1.26.0'
 _PARAMS = re.compile(r'{\w+}')
 
 # Implementation note:  This file is intended to be a standalone file
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -14,7 +14,7 @@ import datetime
 from collections import defaultdict
 
 
-__version__ = '1.26.0'
+__version__ = '1.26.1'
 _PARAMS = re.compile(r'{\w+}')
 
 # Implementation note:  This file is intended to be a standalone file
@@ -411,6 +411,7 @@ class Request(object):
         self.path = event_dict['requestContext']['resourcePath']
         self.lambda_context = lambda_context
         self._event_dict = event_dict
+        self.metadata = None
 
     def _base64decode(self, encoded):
         if not isinstance(encoded, bytes):
@@ -533,7 +534,7 @@ class RouteEntry(object):
 
     def __init__(self, view_function, view_name, path, method,
                  api_key_required=None, content_types=None,
-                 cors=False, authorizer=None):
+                 cors=False, authorizer=None, metadata=None):
         self.view_function = view_function
         self.view_name = view_name
         self.uri_pattern = path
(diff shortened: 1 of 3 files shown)


jamesls:
Hi, thanks for the pull request.  Several of your examples mentioned filtering/processing of middleware based on the HTTP endpoint.  For the REST API, was there an issue with accessing those attributes directly from the [request](https://aws.github.io/chalice/api.html#request)?  e.g. `event.path`:

```python
from chalice import Chalice

app = Chalice(app_name='demo-middleware')
app.debug = True

@app.middleware('all')
def my_middleware(event, get_response):
    app.log.info("Before calling the view function.")
    if event.path == '/':
        app.log.info("Skipping middleware processing.")
        return get_response(event)
    app.log.info("Doing middleware logic.")
    # Custom code here.
    return get_response(event)


@app.route('/')
def index():
    return {'hello': 'world'}


@app.route('/foo')
def foo():
    return {'foo': 'bar'}
```

Are there other concrete examples you can share that demonstrate specific type of metadata you might pass through that aren't attributes you can access in a request?  What kind of "roles" did you have in mind?  I like the idea of additional metadata for middleware, but I think specific examples of how you would use this would help sell the idea.

weaverjess:
Thanks for taking a second to have a look! And ah, thats very true; I'll try to provide some examples.

Re: event/request object
It works, but none of them solve this problem. The "context" object is most similar to the goal of "metadata" maybe, but you can not inject it from route declaration/via the decorator. My gut said it was better to not touch/alter that as it could have impact on existing APIs as it would alter the interface for context a little.

Essentially the goal of this feature is to enhance a **route_entry** with specific metadata to be used at any point of the lifecycle. This allows static rules or information that is always true to be pinned to a route at its very start so anywhere down the line, middleware, handlers etc, it can be used as a source of truth.

A simplified example for something like "role" based allow/deny middleware.

```python

app = Chalice(app_name='demo-middleware')
app.debug = True

@app.middleware('http')
def check_role(event, get_response):
    # imagine this comes from a JWT token or some part of the request, but for the sake of being fast I'll hardcode it
    role = "user"
    
    if "user" in event.get("metadata"):["allow_role"]:
        return get_response(event)
    return Response(status_code=401, body="{"error":"You are unauthorized to make this request"}


@app.route('/', metadata={"allow_role": ["user"] })
def index():
    return {'hello': 'world'}
```
 
The above example assumes there is a middleware or authorizer that injects the role claim somewhere. When declaring the route, we pass the desired metadata. This value is accessible at any point in the request. The middleware looks at this on all HTTP requests in order to allow or short circuit.

weaverjess:
I might be missing something on the event object though that would allow this to be defined dynamically at the route level, but as far as I can tell the example above isn't possible right now.

PR #1853, opened 2021-12-12
Fix: repeating Code in Message for ChaliceViewError and all it's subclasses

dhensen:
*Description of changes:*

[According to docs](https://aws.github.io/chalice/topics/views.html#error-handling) the http response for a BadRequestError should look like:
```
{
    "Code": "BadRequestError",
    "Message": "This is a bad request"
}
```
> (Merging this PR will make this expectation match with actuality) 

But actually this is produced:
```
{
    "Code": "BadRequestError",
    "Message": "BadRequestError: This is a bad request"
}
```
> The Code is repeated in the Message string. This holds for all of ChaliceViewError it's subclasses

This PR fixes that. So basically when you merge this the docs are telling the truth.
Currently I am using a subclassed BadRequestError where I "fixed" this. I would like to use the chalice exceptions, instead of subclassing each time I want to use a different subclass, because I really need to have the  `Message` to not contain the `Code`.
I also think the way the docs put it, is the better way to go about this.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -106,9 +106,6 @@ class WebsocketDisconnectedError(ChaliceError):
 class ChaliceViewError(ChaliceError):
     STATUS_CODE = 500
 
-    def __init__(self, msg=''):
-        super(ChaliceViewError, self).__init__(msg)
-
 
 class ChaliceUnhandledError(ChaliceError):
     """This error is not caught from a Chalice view function.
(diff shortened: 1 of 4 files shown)


jamesls:
Thanks for the PR!

Issue: https://github.com/aws/chalice/issues/1339

Would you mind updating the docs as well?  Looks like there's one place that needs to be updated:

```
$ rg --type rst '"Message": "\w+: .*'
tutorials/basicrestapi.rst
114:        "Message": "ChaliceViewError: An internal server error occurred."
```

dhensen:
@jamesls you're welcome. Thanks for giving it attention. I've done the change :+1:

jamesls (approved):
Looks good, thanks for updating.

jamesls:
Merged in adcc8dbd

PR #1869, opened 2022-01-06
[pkg][terraform]remove template provider

soloradish:
*Issue #, if available:*

*Description of changes:*
terraform template provider cannot work on m1 chip & discontinued. 
Remove this provider, use local variable instead.

### DIFF
--- a/.changes/next-release/23936568646-enhancement-Terraform-50628.json
+++ b/.changes/next-release/23936568646-enhancement-Terraform-50628.json
@@ -1,0 +1,5 @@
+{
+  "type": "enhancement",
+  "category": "Terraform",
+  "description": "Remove template provider in favor of locals (#1869)"
+}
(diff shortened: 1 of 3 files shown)


soloradish:
https://registry.terraform.io/providers/hashicorp/template/latest/docs

```
This provider is deprecated.
We now recommend that you use one of the approaches in the Deprecation section below.

In Terraform 0.12 and later, the templatefile function offers a built-in mechanism for rendering a template from a file. Use that function instead, unless you are using Terraform 0.11 or earlier.
```

jamesls (approved):
Thanks for updating, looks good to me.

As a separate PR it might be good to add the app/stage_name over to locals as well (though we'll need to keep the `null_data_source` for now in case users are using these values in their existing chalice apps).

PR #1881, opened 2022-01-30
Fixes #1879: Serializes Sets in Response

### LINKED ISSUE #1879: Object of type set is not JSON serializable
I'm getting an item from a DynamoDB Table that has a set attribute. Simply returning the item at the end of the request through Chalice raises this error.

## Chalice

```python
@app.route("/items/{item_id}", methods=["GET"])
def get_item(item_id):
   # ...
   item = dynamodb_table.get_item(Key=key).get("Item", key)
   return item
```
## Item in DynamoDB

```jsonc
{
  // ...
  "Set": {
    "SS": [
      "A",
      "B"
    ]
  }
}
```

## Stack trace

```
[ERROR] TypeError: Object of type set is not JSON serializable
  File "/var/task/chalice/app.py", line 1239, in __call__
    return handler(event, context)
  File "/var/task/chalice/app.py", line 1672, in __call__
    return response.to_dict(self.api.binary_types)
  File "/var/task/chalice/app.py", line 473, in to_dict
    body = json.dumps(body, separators=(',', ':'),
  File "/var/lang/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/var/lang/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/var/lang/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/var/task/chalice/app.py", line 64, in handle_extra_types
    raise TypeError('Object of type %s is not JSON serializable'
```

## Solution

Add:

```python
    if isinstance(obj, set):
        return list(obj)
```

To:

https://github.com/aws/chalice/blob/61f43459a078524b4d80cbb816bb4d5d67884a47/chalice/app.py#L55-L65

That should work, but before writing a PR I want to make sure my solution is sound and maybe have some guidance on how to write a test for it. Currently it seems the only test where `handle_extra_types` is used is `test_http_request_to_dict_is_json_serializable`, but there is no test for the response.

https://github.com/aws/chalice/blob/adcc8dbdb0ef86b7aa9238bf90e718c6cd3413f3/tests/unit/test_app.py#L1805

leonelgalan:
I'm getting an item from a DynamoDB Table that has a set attribute.
Simply returning the item at the end of the request through Chalice
raises this error:

> TypeError: Object of type set is not JSON serializable

Issue #1879

It handles set as an "extra" type during JSON Serialization.

___

### DIFF
--- a/tests/unit/test_app.py
+++ b/tests/unit/test_app.py
@@ -1816,6 +1816,17 @@ def test_http_response_to_dict(body, headers, status_code):
     assert isinstance(serialized['body'], six.string_types)
 
 
+@given(headers=STR_MAP,
+       status_code=st.integers(min_value=200, max_value=599))
+def test_http_response_with_set_to_dict(headers, status_code):
+    r = Response(body={'example': set([1, 2, 3])}, headers=headers, status_code=status_code)
+    serialized = r.to_dict()
+    assert 'headers' in serialized
+    assert 'statusCode' in serialized
+    assert 'body' in serialized
+    assert isinstance(serialized['body'], six.string_types)
+
+
 @given(body=st.binary(), content_type=st.sampled_from(BINARY_TYPES))
 def test_handles_binary_responses(body, content_type):
     r = Response(body=body, headers={'Content-Type': content_type})
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -57,6 +57,9 @@ def handle_extra_types(obj):
     # to support that as well.
     if isinstance(obj, decimal.Decimal):
         return float(obj)
+    # DynamoDB responses might contain sets
+    if isinstance(obj, set):
+        return list(obj)
     # This is added for backwards compatibility.
     # It will keep only the last value for every key as it used to.
     if isinstance(obj, MultiDict):


tests/unit/test_app.py
```diff
@@ -1816,6 +1816,17 @@ def test_http_response_to_dict(body, headers, status_code):
     assert isinstance(serialized['body'], six.string_types)
 
 
+@given(headers=STR_MAP,
```
jamesls, line 1819: Not sure we need these property based tests for this change.  We're passing in varying headers and status code values but our test is just verifying the values of sets.  You can probably just create various `Response` instances and verify the `to_dict()` serializes the types properly.  Something like `test_can_encode_binary_body_as_base64()`. 

Probably worth adding an extra test for nested sets as well.

jamesls (commented):
Thanks for the PR.  Not opposed to the change.  I think the original reason for not supporting sets was to match what the default json serializer did in Lambda, but this seems reasonable to me, especially given DynamoDB can return sets.

Had a comment about the test you added, but once that's updated we should be good to merge.

PR #1886, opened 2022-02-04
Convert any ChaliceViewError to a response

vfrico:
*Description of changes:*
Using a method on the class ChaliceViewError would allow any subclass to implement its own custom response for an exception.


jamesls (commented):
Hi, thanks for the pull request, seems reasonable to me.  I'd like to get a test added to `tests/unit/test_app.py` that verifies this behavior with a custom subclass to ensure we don't regress on this behavior in the future.

vfrico:
I've just added tests and docs.

On my local machine, the command `make prcheck` runs ok, but on the last commit the tests didn't ran successfully.

PR #1892, opened 2022-02-19
fix: use correct key for cognito username

kapilt:
*Description of changes:*

local cognito auth is broken, due to using an incorrect claim key for username, its not clear where the previous value comes from all cognito jwt payload docs point to username.   


jamesls:
I think this might be the difference between the ID token, which appears to use the `cognito:username` key (https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-id-token.html) vs. the access token, which uses `username` (https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-access-token.html).

Maybe we can just check for both, but I think we still need to support `cognito:username`.

PR #1896, opened 2022-03-03
Set a default timeout in local

b4stien:
Downstream code (namely `LambdaContext.get_remaining_time_in_millis()` [see here](https://github.com/aws/chalice/blob/91bcfa20ae423c92a94f2043a28828287281b7ee/chalice/local.py#L293-L296)) does not expect a `None` value for `timeout`.

In particular `LambdaContext.get_remaining_time_in_millis()` is called by the Sentry integration [here](https://github.com/getsentry/sentry-python/blob/4c09f3203d6d19789c6fa729a2e46557ad4ea913/sentry_sdk/integrations/chalice.py#L40) and [here](https://github.com/getsentry/sentry-python/blob/4c09f3203d6d19789c6fa729a2e46557ad4ea913/sentry_sdk/integrations/chalice.py#L67) and it provokes a completely silent failure.

As the AWS Lambda max timeout is 15min I think the change is pretty safe.

### DIFF
--- a/.changes/next-release/30122944134-bugfix-Local-93796.json
+++ b/.changes/next-release/30122944134-bugfix-Local-93796.json
@@ -1,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "Local",
+  "description": "Set a default timeout when creating the local LambdaContext instance (#1896)"
+}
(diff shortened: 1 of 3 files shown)


jamesls:
Thanks for the fix.  I'd like to get a test added to ensure we don't regress on this behavior in the future before this gets merged.

jamesls (approved):
Shipit, assuming CI passes.

Issue #1670, opened 2021-03-03
Websocket Terraform Support


seanieb:
Can you provide clarity around Websockets Terraform support? 

It would be  helpful to reference that Terraform wont work under the warning on the Websockets documentation:
https://aws.github.io/chalice/topics/websockets.html

PR #1922, opened 2022-05-11
Add support for WebSockets API Terraform packaging

adityamohta:
*Issue https://github.com/aws/chalice/issues/1670

*Description of changes:*

Adds terraform packaging support for WebSockets API with its unit tests.

### DIFF
--- a/chalice/package.py
+++ b/chalice/package.py
@@ -18,7 +18,7 @@ from chalice.deploy.swagger import (
 from chalice.utils import (
     OSUtils, UI, serialize_to_json, to_cfn_resource_name
 )
-from chalice.awsclient import TypedAWSClient # noqa
+from chalice.awsclient import TypedAWSClient  # noqa
 from chalice.config import Config  # noqa
 from chalice.deploy import models
 from chalice.deploy.appgraph import ApplicationGraphBuilder, DependencyBuilder
@@ -101,8 +101,8 @@ class PackageOptions(object):
 class ResourceBuilder(object):
     def __init__(self,
                  application_builder,  # type: ApplicationGraphBuilder
-                 deps_builder,         # type: DependencyBuilder
-                 build_stage,          # type: BuildStage
+                 deps_builder,  # type: DependencyBuilder
+                 build_stage,  # type: BuildStage
                  ):
         # type: (...) -> None
         self._application_builder = application_builder
@@ -122,7 +122,6 @@ class ResourceBuilder(object):
 
 
 class TemplateGenerator(object):
-
     template_file = None  # type: str
 
     def __init__(self, config, options):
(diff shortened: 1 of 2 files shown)


adityamohta:
We love chalice and actively use it with Terraform deployment in some of our services. 

We found support for websockets_api deployment with `chalice deploy` and `SAM`. But, `terraform` support (which we really love) was missing. 

Therefore, we decided to add support for the same. 

---

Since this is my first contribution to Chalice, it would be great if a maintainer can take out time to run a workflow/review this PR. 

Perhaps @jamesls ?

chalice/package.py
```diff
@@ -858,17 +857,198 @@ def _generate_managediamrole(self, resource, template):
...
+        if 'websocket_connect' in aws_lambda_functions:
+            output['WebsocketConnectHandlerArn'] = {
+                "value": "${aws_lambda_function.websocket_connect.arn}"}
+            output['WebsocketConnectHandlerName'] = {
+                "value": "${aws_lambda_function.websocket_connect}"}
+        if 'websocket_message' in aws_lambda_functions:
+            output['WebsocketMessageHandlerArn'] = {
+                "value": "${aws_lambda_function.websocket_message.arn}"}
+            output['WebsocketMessageHandlerName'] = {
+                "value": "${aws_lambda_function.websocket_message}"}
```
jamesls, line 978: I noticed this prints out all the properties of the lambda function: 

```
WebsocketMessageHandlerName = {
  "architectures" = tolist([
    "x86_64",
  ])
  "arn" = "arn:aws:lambda:us-west-2:12345:function:testws-dev-websocket_message"
  "code_signing_config_arn" = ""
  "dead_letter_config" = tolist([])
  "description" = ""
  "environment" = tolist([])
```

The SAM/cfn version uses `{'Ref': 'WebsocketMessage'}` which just returns the [resource name](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#aws-resource-lambda-function-return-values).

I believe we just need to add a `.function_name` to this output (same to all the other ws function handlers).

chalice/package.py
```diff
@@ -101,8 +101,8 @@ def service_principal(self, service):
 class ResourceBuilder(object):
     def __init__(self,
                  application_builder,  # type: ApplicationGraphBuilder
-                 deps_builder,         # type: DependencyBuilder
-                 build_stage,          # type: BuildStage
+                 deps_builder,  # type: DependencyBuilder
```
jamesls, line 104: From what I can tell, `prcheck` still passes without these formatting changes.  I'm going to try reverting these changes and see if the CI passes.

jamesls (commented):
Changes look good.  I had some small feedback, should be pretty straightforward so I'll go ahead and incorporate it real quick.

jamesls:
I wasn't able to push to the branch for this PR for some reason so I incorporated the feedback and opened up a new PR: https://github.com/aws/chalice/pull/1928.  Let's track the remaining work over in that PR.

chalice/package.py
adityamohta: Nice catch! Thanks for fixing it @jamesls

Issue #1802, opened 2021-09-13
Remove app.pyi file


jamesls:
This tracking issues proposes removing the [app.pyi](https://github.com/aws/chalice/blob/master/chalice/app.pyi) from Chalice as we no longer need it.

This file was originally added because we had to support Python 2.7.  This meant not only could we not use annotations for type info, but we couldn't import `typing` because we wanted to avoid the extra runtime cost of having that module available for python2.

This `app.pyi` file has always been out of sync.  We made a best effort to keep things updated, but this was also before tools such as `mypy.stubtest` were around to help us verify they match.  There have been several PRs from contributors to try and bring things into sync, but this will always be a moving target, and even checking on this now, there's still inconsistencies between the two files.

Now that we no longer support Python 2.7, I'd like to drop the app.pyi file and add the type information directly into app.py.  We'll need to make sure we only use type info from Python 3.6, but this should fix the inconsistencies we have, and make writing Chalice apps easier as we see more people using PyCharm, VSCode, pyright, mypy, etc.

PR #1932, opened 2022-05-26
Remove app.pyi file 

adityamohta:
Issue https://github.com/aws/chalice/issues/1802


chalice/app.py
```diff
@@ -227,19 +236,19 @@ def __repr__(self):
...
-    scopes = []
+    name: str = ''
+    scopes: List[str] = []
 
-    def to_swagger(self):
+    def to_swagger(self) -> Dict[str, Any]:
         raise NotImplementedError("to_swagger")
 
-    def with_scopes(self, scopes):
+    def with_scopes(self, scopes: List[str]) -> 'Authorizer':
```
adityamohta, line 245: Since `from __future__ import annotations` is only supported for python 3.7+ 
To support v3.6, We are using quotes for annotations referring to classes declared after usage. Example - `'Authorizer'`

tests/functional/test_local.py
```diff
@@ -107,7 +107,7 @@ def http_session():
         connect=10,
         # A backoff factor to apply between attempts after the second try.
         backoff_factor=2,
-        method_whitelist=['GET', 'POST', 'PUT'],
+        allowed_methods=['GET', 'POST', 'PUT'],
```
adityamohta, line 110: Fixes method_whitelist deprecation warning in the next version from urllib3 library

jamesls: Would you mind sending a separate PR for this?  I'd like to keep this one just focused on removing the app.pyi file.

jamesls:
Thanks for starting the work on this!  I reran the failed jobs (looks like there's an unrelated intermittent test failure I need to track down).  Looks like there's some type errors that are now failing.

adityamohta:
> Thanks for starting the work on this! I reran the failed jobs (looks like there's an unrelated intermittent test failure I need to track down). Looks like there's some type errors that are now failing.

@jamesls Thanks for taking a look! Yes, I did shift all the type declarations to `app.py` from the `app.pyi` file, but the `app.py` was still missing many. I have defined those and resolved all the test failures.

jamesls:
Wow, thanks for doing this!  Taking a look now.

jamesls (approved):
Looks good to me.  I'm going to go ahead and squash/merge.  I'll also spend some time seeing if we can get even more specific with some of these types and replace some of these `Any`s.

adityamohta:
> Looks good to me. I'm going to go ahead and squash/merge. I'll also spend some time seeing if we can get even more specific with some of these types and replace some of these `Any`s.

Excellent, Thanks for taking out time @jamesls !

> I'll also spend some time seeing if we can get even more specific with some of these types and replace some of these `Any`s.

Awesome! `Any`s are frustrating and I did try to avoid it in this PR but for some vars, it was a rabbit hole, so I skipped some. :/

PR #1945, opened 2022-06-25
Add test helper to generate DynamoDB event

IdrisMiles:
*Description of changes:*

Add `generate_dynamodb_event` to the `TestEventsClient` class to allow testing `on_dynamodb_record` handler lambdas more easily.

### DIFF
--- a/chalice/test.py
+++ b/chalice/test.py
@@ -4,7 +4,7 @@ import base64
 import contextlib
 from types import TracebackType
 
-from typing import Optional, Type, Generator, Dict, Any, List  # noqa
+from typing import Optional, Type, Generator, Dict, Any, List, Tuple  # noqa
 
 from chalice import Chalice  # noqa
 from chalice.config import Config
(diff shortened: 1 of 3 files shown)


chalice/test.py
```diff
@@ -323,6 +323,29 @@ def generate_kinesis_event(self, message_bodies,
         } for body in message_bodies]
         return {'Records': records}
 
+    def generate_dynamodb_event(self, images):
+        # type: (List[Tuple[Dict, Dict]]) -> Dict[str, Any]
+        records = [{
+            "dynamodb": {
+                "ApproximateCreationDateTime": 1545084650.987,
+                "Keys": list(new_image.keys()),
+                "NewImage": new_image,
```
jamesls, line 332: Are the `NewImage`/`OldImage` keys `null` or omitted if your view type doesn't include them, e.g. `KEYS_ONLY`?  We should try and match that behavior.

jamesls (commented):
Looks good, just had a question about the presence of certain keys based on your configured view type for your stream.

chalice/test.py
IdrisMiles: It's not currently respecting the view type properly, will update the behaviour so it omits the relevant keys for the `KEYS_ONLY`/`NEW_IMAGE`/`OLD_IMAGE`  stream view types

PR #2012, opened 2023-01-04
Add support for greedy catch-all path variables

jacobe:
This was attempted/discussed in #140, but here is a more complete and also tested implementation.

API Gateway has support for ["greedy" path parameters](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html#http-api-develop-routes.evaluation) in APIs routes, but that is not fully supported in Chalice. We have run into several use cases for such a feature in our company:
* Varying callback URLs from 3rd party providers that all need to go to one handler
* Static asset mapping

Currently, if you specify a `{proxy+}` path variable in an `@app.route`, it will actually work, but there are a few things lacking:
1. It's not possible to map the value to an actual parameter in the function. I'd expect to be able to get the proxied portion of the path through a parameter named "proxy".
2. The catch-all parameter does not work for local development (`chalice local`)

Those gaps are implemented in this PR. Usage would be as follows:

```python
@app.route('/foo/{proxy+}', methods=['GET'])
def example(proxy):
    return proxy
```

This will match any url under `/foo` (although not `/foo` itself), returning the proxied portion of the path:

```
GET /foo/test
=> test

GET /foo/test/bar
=> test/bar
```

The PR also adds support local development (`chalice local`) so that the matching also works locally just as it does in API Gateway.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -30,7 +30,9 @@ from typing import List, Dict, Any, Optional, Sequence, Union, Callable, Set, \
 if TYPE_CHECKING:
     from chalice.local import LambdaContext
 
-_PARAMS = re.compile(r'{\w+}')
+# the optional + at the end is for supporting the special greedy parameter in
+# API Gateway (ie. "{proxy+}")
+_PARAMS = re.compile(r'{(\w+)\+?}')
 MiddlewareFuncType = Callable[[Any, Callable[[Any], Any]], Any]
 UserHandlerFuncType = Callable[..., Any]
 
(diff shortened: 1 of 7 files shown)


annjawn:
Hey @jamesls any chance we can get this PR merged?

bt:
Just came across this issue whilst building locally.

Any chance we can get this merged?

mangled-data:
Bumping this. I think greedy catch-all path is still a big problem. Atleast I haven't been able to get it working.

Issue #2015, opened 2023-01-24
Question: Is Chalice being supported?


bondgeek:
I notice that no PRs have been closed since Sept 2022 -- can we count on Chalice receiving continued support?  Or should we create our own fork if we want to continue to use it?

FelipeRuizGarcia:
@bondgeek , I saw there is not new branches or updates...  @jamesls @stealthycoin  could you please grant us access to maintainers in this repo or should we fork it?  I would like  continue with this official repo ( PRs, bugs ... )

hitrust:
I also think this question should be considered. 
I think maybe boto team may focus on boto and other things that only the team can reach, maybe just make outline for this repo, and review and approve the PRs, so we can contribute continually.

jamesls:
Hi, sorry for the delay here.

We are reallocating more time for Chalice support for triaging issues, bug
fixes, and maintenance PRs.  Long term, we'd like to work towards a Chalice 2.0
that will allow Chalice to be more easily extended without requiring changes to
the core code.  This requires removing a number of old features that would otherwise
make this work unrealistic.  I don't have a specific timeline for Chalice 2.0 but
will be adding proposals updates as work proceeds (they'll have the label `v2`).

Going through the issues/PRs now, but if there's anything in particular we should look at,
let us know.

bondgeek:
Thanks @jamesls That is good to hear.  I think Chalice is fabulous and would like to lean into it more -- we're not even fully using its current capabilities.  If there is a path to a Chalice 2.0, then that gives the comfort to have more people on the team learn to use it.

In terms of specifics for v2
> Going through the issues/PRs now, but if there's anything in particular we should look at,
let us know.

I would just beg you to start by thinking about the upgrade path for developers.  If upgrading to v2 is equivalent to moving to a new framework, then I will re-evaluate all alternative frameworks before upgrading.

PR #2036, opened 2023-04-21
python 3.10

ogrodnek:
*Issue #, if available:* #2037

*Description of changes:*

- Attempt at adding python 3.10 support (https://aws.amazon.com/blogs/compute/python-3-10-runtime-now-available-in-aws-lambda/)

### DIFF
--- a/chalice/deploy/packager.py
+++ b/chalice/deploy/packager.py
@@ -81,6 +81,7 @@ class BaseLambdaDeploymentPackager(object):
         'python3.7': 'cp37m',
         'python3.8': 'cp38',
         'python3.9': 'cp39',
+        'python3.10': 'cp310',
     }
 
     def __init__(
@@ -497,6 +498,7 @@ class DependencyBuilder(object):
         'cp36m': (2, 17),
         'cp37m': (2, 17),
         'cp38': (2, 26),
+        'cp310': (2, 26)
     }
     # Fallback version if we're on an unknown python version
     # not in _RUNTIME_GLIBC.
--- a/tests/integration/test_package.py
+++ b/tests/integration/test_package.py
@@ -52,13 +52,13 @@ def _get_random_package_name():
                     reason='Extended packaging tests only run on py3.')
 @pytest.mark.parametrize(
     'package,contents', [
-        ('pandas==1.1.5', [
+        ('pandas==1.5.3' if sys.version_info[1] >= 10 else 'pandas==1.1.5', [
             'pandas/_libs/__init__.py',
             'pandas/io/sas/_sas.cpython-*-x86_64-linux-gnu.so']),
-        ('SQLAlchemy==1.3.20', [
+        ('SQLAlchemy==1.4.47', [
             'sqlalchemy/__init__.py',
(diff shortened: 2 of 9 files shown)


VitalyAssurance:
Thank you for providing the fix. I hope AWS team will merge it soon.

tests/integration/test_package.py
```diff
@@ -52,13 +52,13 @@ def _get_random_package_name():
                     reason='Extended packaging tests only run on py3.')
 @pytest.mark.parametrize(
     'package,contents', [
-        ('pandas==1.1.5', [
+        ('pandas==1.5.3' if sys.version_info[1] >= 10 else 'pandas==1.1.5', [
```
ogrodnek, line 55: I had to update a lot of the package versions to ones that would work with 3.10.  I tried to find releases that would work for 3.7 up to 3.10, but wasn't able to for pandas, so split it out to use a different version for 3.10 and up.

chalice/deploy/packager.py
```diff
@@ -497,6 +498,7 @@ class DependencyBuilder(object):
         'cp36m': (2, 17),
         'cp37m': (2, 17),
         'cp38': (2, 26),
+        'cp310': (2, 26)
```
ogrodnek, line 501: I think this is right based on the AWS provided docker image

```
$ docker run -it --entrypoint "ldd" public.ecr.aws/lambda/python:3.10 --version
ldd (GNU libc) 2.26

```

sorendaugaard:
@jamesls in https://github.com/aws/chalice/issues/2015 you mentioned that Challice is still being supported. 

Any chance we can get this Python 3.10 upgrade reviewed and merged?

dlchet:
I successfully used this branch to deploy a 3.10 lambda function.

jamesls:
Thanks for the PR, looks like all the CI is passing for 3.10, going to pull it down locally and try it out.

devinmatte:
Any update on this @jamesls? This is the last blocker for updating my whole stack to 3.10

tests/integration/test_package.py
jamesls: Sounds good.  Given this is already run as part of the slower CI test runs, I might split these out into different versions per python version just to ensure we don't regress on supporting these packages for older versions of python.

jamesls (approved):
Looks good, sorry for the delay.  Just want to make a few tweaks to the functional tests, but otherwise I think this is good to merge.  Thanks again!

tests/integration/test_package.py
jamesls: Pushed https://github.com/aws/chalice/pull/2036/commits/c7906bfc34126a3cf6599601ade299ce6ec1afe5 for what I had in mind.  I'm being cautious here, but I'm preserving the existing package version for python < 3.10, and then all python versions >= 3.10 will use the versions you have in your PR.

Assuming the CI passes, I'll go ahead and merge, unless there's any concerns with this approach.  Thanks again!

ogrodnek: thanks James!

jamesls:
Version [1.29.0](https://pypi.org/project/chalice/1.29.0/) is now released, which includes this PR.

PR #2050, opened 2023-06-25
Fix type of Response.headers

cdroege:
The type for `Response.headers` was previously typed as `Dict[str, str]`, but that is not completely correct. Header values can also be a list of strings, e.g. for multiple `Set-Cookie` headers. See #1214 when this feature was introduced. 

The wrong type definition was introduced in afd23d7d729f8383033fe3851bc1de5effd1f22f.


jamesls (approved):
Thanks!

PR #2052, opened 2023-07-05
(Easy merge ⭐️) Upgrade Sample Todo App to Use Latest 3rd Party Python Dependencies

codyfletcher:
Upgrading the dependencies doesn't work by itself as there was an issue with jwt token. This resolves that problem.

*Description of changes:*

1. Upgrading the 3rd-party Python dependencies 
2. Step 1 above doesn't work by itself, as there was an issue with jwt token code in chalicelib/auth.py. Two simple adjustments to that file has everything working with the latest versions.


jamesls (approved):
Thanks!

PR #2087, opened 2023-12-15
Add support for Python 3.12

### LINKED ISSUE #2086: Support Python 3.12
AWS announced today support for Python 3.12 in Lambda

https://aws.amazon.com/about-aws/whats-new/2023/12/aws-lambda-support-python-3-12/

It would be great to see Chalice support this new version as well 😄

Thanks!

jamesls:
In addition to the mechanical changes needed to support a new Python version, I also made a few additional changes to get everything passing:

* Removed testing cdkv1.  CDK v1 is deprecated now, and while we still have code that will try and import cdkv1, we're no longer testing support for CDK v1.
* Refresh the test/dev dependencies via `pip-compile`.  New package versions were needed for Python 3.12 support.
* Update the pylintrc config to match the previous behavior of earlier versions
* Only use `typing_extensions` on Python 3.7, which is the only version that needs this.  `TypedDict` was added in 3.8, and it's the only thing from that package that we use.

Closes #2086


jamesls:
Looks like there's some issues with the pylint specific to python3.12 that I'll need to track down.  Otherwise the feature itself seems to be working (I can deploy python3.12 apps to Lambda).

.github/workflows/run-tests.yml
```diff
@@ -53,8 +53,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10', 3.11]
-        cdk-version: [cdk, cdkv2]
+        python-version: [3.7, 3.8, 3.9, '3.10', 3.11, 3.12]
```
devinmatte, line 56: Just curious, any reason to still support Python 3.7 since it's EOL? https://devguide.python.org/versions/

jamesls: I have a branch that removes 3.7 support, but we're following the python support policy for AWS SDKs and Tools (https://aws.amazon.com/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/), which would drop 3.7 support at the end of this month.

I was hoping to get 3.12 support released quickly and then drop support for 3.7 next month, but depending on how long it takes to work out the remaining 3.12 issues, we might end having 3.12 support + dropping 3.7 in the same release if it ends up getting pushed to next month.

devinmatte: @jamesls It's the new year, any chance you could take a look at this again?

werlex:
@jamesls @devinmatte seems like 3.7 support was dropped. So you are planning to remove 3.7 support and add 3.12 with this same PR?

chalice/config.py
```diff
@@ -155,9 +155,9 @@ def lambda_python_version(self) -> str:
         # 3.11 and higher will use 3.11
```
MRigal, line 155: isn't this whole property needing to be removed? AWS supports only Python 3.8+ anyway...

tests/unit/test_config.py
```diff
@@ -379,8 +379,10 @@ def test_can_load_python_version():
         expected_runtime = 'python3.9'
     elif minor <= 10:
         expected_runtime = 'python3.10'
```
MRigal, line 381: Also here, we are Py 3.7+ if not 3.8+, a lot can disappear

jamesls:
Quick update, python 3.7 has been deprecated for a few months now and haven't seen any issues in the SDK/other projects so I think we should be good to move ahead with also dropping 3.7 support.  I think the simplest path here is to send a pr to drop 3.7 support, the rebase this PR on top.  Working on this now, should have an update shortly.

jamesls:
Py37 dropped in https://github.com/aws/chalice/pull/2095, going to rebase now and should be able to get a clean run through CI.

tests/unit/test_config.py
jamesls: Yeah, there's a number of places in the tests that still reference older versions of python that can be removed.  I'll send a follow up pull request to clean these up.

SamRemis (approved):

PR #2125, opened 2024-12-04
bumping readme suggested python version from 3.7 to 3.12

jogspokoen:
Documentation update only (Readme).

Even though Chalice supports 3.7, it seems wrong to suggest using obsolete 3.7 in Readme and Quickstart
I updated it to 3.12 with 2028-10 end of support.


jonathan343:
Closing. The README now uses Python 3.10 which is our minimum supported version. I'll consider changing this to use the latest supported version (3.14). In either case, we are now off 3.7.

PR #2140, opened 2025-05-13
[v2] Modernize repo with latest python standards

jamesls:
This modernizes the repo with the latest python standards for easier contribution

## Python version support - 3.12+

My thinking here is that for v2, we'll start with only worry about python 3.12/3.13 for now. I want the focus to be on an updated set of features if we only support modern python, and leave the discussion of what versions of python to support as a separate discussion.  This is because 1) we don't know when 2.0 will be finished and what the set of supported python vesrions will be with Lambda / botocore 2) In modern versions of python, the main differences in practice are minor typing enhancements and stdlib updates, very few language features I think we'd actually use that aren't available in 3.9/3.10+.  The effort of adding older Python version support after the fact is much easier than the py3/py2 times so I think it will be minimal work.

In addition to python version I've also:

* Standardize on uv
* Migrate to (only) `pyproject.toml`
* Use `ruff` for lintint.  Note that ruff did not catch all the issues that
  we could previously catch so it's unfortunately not a drop in replacement.
  As a result we still have a `pylint -E` pass nwe need.

cc @jonathan343 @nateprewitt @SamRemis


jamesls:
Ah, so I had to [remove the CDK test runs](https://github.com/aws/chalice/pull/2140/commits/28b0cb5548520c69de149d916b7ad42d0b62b20e) from the GitHub test run action.  My original branch removed the CDK feature entirely but I wanted to keep this PR focused on repo modernization for now (I'll send a follow up PR to remove the actual code bits related to CDK).

FWIW I'd still like to support CDK apps, but I have a different idea in mind for how to do it and I'd also like this to be in a separate package.

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+[project.urls]
+Homepage = "https://aws.github.io/chalice/"
+
+[tool.uv]
+dev-dependencies = [
+    "ruff",
+    "pylint",
+    "doc8<1.0.0",
+    "pydocstyle",
+    "flake8",
```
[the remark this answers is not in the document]
nateprewitt: Ok, no worries. May be good to create an issue/comment later to record any blockers to get us off pylint, but this is fine for now.

.github/workflows/run-tests.yml
```diff
@@ -17,39 +20,21 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
+        python-version: [3.12]
```
nateprewitt, line 23: Actually, last comment, I know we say we support 3.12+ and Lambda doesn't currently support 3.13. Would it hurt to test 3.13 now so things are ready to go as soon as Lambda releases?

nateprewitt: Nevermind, they already released [3.13](https://aws.amazon.com/blogs/compute/python-3-13-runtime-now-available-in-aws-lambda/). Would that be worth adding?

SamRemis (approved):
Looks good :) 

If we are making a new branch, is there some sort of timeline for when this will be publicly documented/distributed?

jamesls:
> If we are making a new branch, is there some sort of timeline for when this will be publicly documented/distributed?

Don't have a definite timeline of things just yet, I think it'll depend on when we get to a point where we're ready for others to start experimenting and trying out v2.

I plan to develop publicly on the v2 branch here while trying to keep the documentation updated, but initially wasn't going to worry about publishing the rendered docs on github pages.  At some point I think it would make sense to have start publishing the v2 preview docs.

As for distribution, I'll need to double check but assuming there's no issues with users unintentionally picking up a v2 alpha/beta release, I have no issues starting to publish a v2 package on pypi sooner rather than later.

Issue #2082, opened 2023-11-21
Update demo Applications to Python 3.10


rserranon:
Request to update demo Applications to at least Python version 3.10

I haven't been able to run the todo application, Python 3.7 is not supported by home-brew anymore. `"python@3.7 has been disabled because it is deprecated upstream!" `

jonathan343:
Hey @rserranon!

Not sure what demo applications you're referring to exactly. But in the PR linked above I moved a few references off of older / unsupported Python versions and onto at least Python 3.10.

I will close this out for now. If you find anything I missed, feel free to re-open.

Thanks!

PR #2162, opened 2026-05-01
Drop support for Python 3.9

jonathan343:
> [!NOTE]
> - Python 3.9 reached end-of-life on 2025-10-31. It is no longer supported and does not receive security updates. See [PEP 596](https://peps.python.org/pep-0596/) for more details.
> - AWS Lambda deprecated Python 3.9 on Dec 15, 2025. See [Deprecated runtimes](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated) for more details.

Addresses https://github.com/aws/chalice/issues/2082 and https://github.com/aws/chalice/pull/2125 by moving examples off of Python 3.7. 

## Overview

This PR drops Python 3.9 support and updates Chalice to require Python 3.10 or newer. This aligns the package metadata, runtime selection logic, CI matrix, docs, and tests around the currently supported Python runtime range.

## Testing

- Ran `make prcheck` successfully
- Re-generated Chalice docs to ensure the Sphinx upgrade didn't break anything
- Sanity checked a Python 3.9 Lambda function deployment using `chalice helloworld`, `chalice deploy`, calling the function, and `chalice delete`

---


.github/workflows/run-tests.yml
```diff
@@ -33,12 +33,12 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9, '3.10', 3.11, 3.12, 3.13]
+        python-version: ['3.10', 3.11, 3.12, 3.13]
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v6
```
jonathan343, line 39: **Note:** I'll create a follow up PR to move these to commit SHAs. I'd like to keep this as a separate PR instead of including it here.

.github/workflows/run-tests.yml
```diff
@@ -33,12 +33,12 @@ jobs:
...
         with:
-          node-version: '14'
+          node-version: '24'
```
jonathan343, line 41: This was required since AWS CDK dropped support for Node v14 in 2025-05-30 according to [Node.js versions no longer supported](https://docs.aws.amazon.com/cdk/v2/guide/node-versions.html#node-version-deprecated). I decided to bump us onto the latest supported version, Node v14.

setup.py
```diff
@@ -39,6 +39,7 @@ def recursive_include(relative_dir):
     author="James Saryerwinnie",
     author_email='js@jamesls.com',
     url='https://github.com/aws/chalice',
+    python_requires='>=3.10',
```
alexgromero, line 42: We should also update [`.python-version`](https://github.com/aws/chalice/blob/master/.python-version) from `3.7` to `3.10` to match this or remove it.

I tried creating a venv and got an error:
>error: No interpreter found for Python 3.7 in managed installations or search path

jonathan343: Addressed in 38321704e6474d35058881a09d3fd996e5e56589. I was using `uv venv -p <version>` directly so didn't run into this. Thanks!

alexgromero (approved):

PR #2171, opened 2026-05-06
Upgrade doc8 to fix release infra failures on Python 3.12+

jonathan343:
## Problem

Certain Python 3.12+ environments will fail with the following failure:

```
ModuleNotFoundError: No module named 'pkg_resources'
```

The root cause is `pbr`, a transitive dev dependency pulled in via `doc8`  -> `stevedore`. `pbr` registers a setuptools `egg_info.writers` entry point that unconditionally imports `pkg_resources`, which was removed in setuptools 82+. ([src](https://setuptools.pypa.io/en/stable/history.html#deprecations-and-removals))When chalice's packager tests invoke `setup.py egg_info` in a subprocess, this entry point fires and crashes the process before metadata is written.

The issue does not reproduce in all environments. On some platforms setuptools writes PKG-INFO before the entry point executes, masking the failure.

## Fix

Upgrade `doc8` from 0.11.2 to 1.1.2. The newer versions have dropped their dependency on pbr ([ref](https://github.com/PyCQA/doc8/pull/119/changes))

---

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -1,6 +1,6 @@
 -r requirements-test.in
 pylint<4.0.0
-doc8<1.0.0
+doc8<2.0.0
 pydocstyle
 flake8
 Sphinx==8.1.3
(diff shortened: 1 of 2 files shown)


arandito (approved):
:shipit:

Issue #2147, opened 2025-09-14
Is there any plan to support snapstart ?


jaeyoung0509:
AWS SnapStart is a powerful feature that significantly reduces cold start latency for Python function. 
However the current AWS chalice framework may not provide a native way to enable SnapStart.

`chalice deploy` command and the `.chalice/config.json` file lack a specific option to configure this feature 

I was wondering if there are any plans or a roadmap to natively support SnapStart within chalice?

jonathan343:
Hey @jaeyoung0509, I'd love to get support for snapstart added.

The main blocker right now is that we don't have support for versioning or aliases which are called out as a limitation in the AWS Lambda docs:

> **Note** You can use SnapStart only on [published function versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#configuration-versions-config) and [aliases](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) that point to versions. You can't use SnapStart on a function's unpublished version ($LATEST).

I'll be looking into getting support for these to added to unblock this. In the meantime, if you have feedback on what the optimal experience here is, I'll be happy to consider it!

rokclimb15:
@jonathan343 I'm glad to contribute support for versioning/aliases if it will help move this along

jonathan343:
Hey @rokclimb15, I'm happy to review any related contribution for this. However, I do have something working locally for a simple Chalice config that lets customers define an alias for Chalice to use. The idea would be that on deployment, Chalice will auto-version your Lambda function and point your alias to the latest version.

The current idea is something like:

```
# In .chalice/config.json

{
  "stages": {
    "prod": { "auto_alias": "myalias" }
  }
}
```

The prototype is working, however, I want to flesh out the customer experience and see if there is anything else I should make configurable. If you have any feedback that would be great. I'll try to get something up by tomorrow which you can take a look at.

rokclimb15:
@jonathan343 I think the MVP to provide support for SnapStart should auto-alias based on the stage name (prod, dev, etc) rather than requiring configuration. The alias should be fed to dependencies like API Gateway so the deployment is always stable. Later features might include rollback, deployment with traffic shifting, and version promotion like `chalice promote dev prod`.

jonathan343:
Thanks for the feedback. I opened #2173 with an initial version of the Lambda versioning/alias support.

[...]

Issue #24, opened 2016-07-11
Support for Lambda versions and versioning within API Gateway


davidvanleeuwen:
This is just an idea, so if you think this is not the place, please close this issue.

I've been thinking about a better versioning system for APIs on top of Lambda/API Gateway. It's nice you can create a stage variable and link that to an alias in Lambda. But Lambda also offers versioning, which is basically your API version... You would get something similar to https://zeit.co/now - where you deploy and get your URL back, but instead of just the stage, you could also get a version which lives forever and is hooked up to one of the aliases.

Although I have no clue yet on how to link that version to API gateway, it would be nice to think about it and see whether it is possible to get something like this done.

What do you think?

jamesls:
I sketched out a similar idea to what you proposed over in #20.  Essentially, every "deploy" would create a new function version, and depending on the stage, we'd also update a lambda alias to point to that latest function version.

 If I understand what you're proposing, the biggest difference would be to also have some way to make those lambda function versions also accessible right?

Another way to get similar behavior would be to allow chalice to flip the alias to a different version.

I think this is something I'd like to support (in some form), but the specifics are still up for discussion.  Let's consolidate threads over at: https://github.com/awslabs/chalice/issues/20

Issue #20, opened 2016-07-11
Add full support for stages


jamesls:
Right now, the stages feature needs more work to be feature complete.  The proposed idea is:
- `chalice deploy prod` - Will create a new API gateway stage if necessary
- The lambda function code is zipped up and sent to lambda.  A new function version is created
- A lambda alias for the the stage name points to the new function version updated.

I also need to update the API gateway integration request to forward to the appropriate lambda function based on the stage name.  I'll need to add a context variable with the stage name.

I'd also like to add an optimization to just update lambda aliases when you go from `dev->prod` where the lambda code has not changed.  In that case we can just flip the alias.

timff:
Yes, I agree accounts are much better way for resource isolation. 

In my opinion, stage only required if you have multiple environments in one account, for example, DEV and TEST in one account, while PREPROD and PROD in another. But this is really API Gateway specific.

JamieCressey:
In my mind, stages are a must for dev, test, prod environments. Creating additional accounts will not be an option for everyone.

garnaat:
@JamieCressey I'm curious how you would isolate other resources in that scenario.  For example, if you had a DynamoDB table that the API was fronting.

Also curious why additional accounts would not be an option.  What would be the barrier?

JamieCressey:
@garnaat assuming we are talking about AWS accounts here, there's a multitude of reasons why a corporation would have problems here; cost consolidation being one.

Depending on what isolation you mean, in our current setup (using Serverless) we have a custom role per Lambda function (and therefore API endpoint) which restricts access to the bare minimum per function.

garnaat:
I am talking about AWS accounts.

In my experience most people rely on AWS accounts for resource isolation between prod accounts and dev/test accounts.  I'm not saying there aren't other ways to get reasonable isolation but accounts are pretty bulletproof.

So I'm curious how people accomplish sufficient isolation for non-API Gateway resources using stages.

mannytoledo:
@JamieCressey On a billing level, AWS is more than willing to aggregate all costs into one account. This way you can have 1 place to go for billing but can still have a logical divide of multiple accounts.

AWS and most large corporations suggest multiple accounts for security reasons. It reduces the risk and complexity of isolating access to parts of infrastructure.

I've seen the model where there are 3 api endpoints/lambas for each tier, and it works well. It would really be nice to also have a multi-account model. Maybe a place where credentials can be configured in per tier or even using AWS profiles?

[...]

PR #2173, opened 2026-05-10
Add Lambda versioning and alias support

jonathan343:
Related: #20, #24, #2147

## Overview
This PR adds Lambda versioning and alias support through a new `lambda_alias` config option. When configured, Chalice publishes a Lambda version after updating `$LATEST`, creates or updates the configured alias, and routes Chalice-managed integrations through the alias ARN. This gives apps a stable, versioned invocation target while keeping the behavior opt-in.

The implementation supports direct deploys as well as generated SAM and Terraform templates. SAM templates use `AutoPublishAlias` with `AutoPublishAliasAllProperties`, and Terraform templates publish function versions and create `aws_lambda_alias` resources. The same config can be set globally, per stage, or per function.

Removing `lambda_alias` stops Chalice from publishing versions and updating the alias. On the next direct deploy, managed integrations move back to the unqualified function ARN, but Chalice does not delete the existing Lambda alias from AWS. For packaged deployments, the generated template no longer includes the alias resource.

## Testing

Deployed a new `helloworld` project with `"lambda_alias": "live"` and validated that versioning and alias support worked as expected.

---

### DIFF
--- a/chalice/awsclient.py
+++ b/chalice/awsclient.py
@@ -1002,6 +1002,87 @@ class TypedAWSClient(object):
         if kwargs:
             self._do_update_function_config(function_name, kwargs)
 
+    def publish_function_version(
+        self, function_name: str
+    ) -> Dict[str, Any]:
+        lambda_client = self._client('lambda')
+        try:
+            result = lambda_client.publish_version(
+                FunctionName=function_name
+            )
+        except lambda_client.exceptions.ResourceConflictException:
+            result = self._latest_published_function_version(function_name)
+        self._wait_for_active_function_version(
+            function_name, result['Version'])
+        return result
+
+    def _latest_published_function_version(
+        self, function_name: str
+    ) -> Dict[str, Any]:
+        lambda_client = self._client('lambda')
+        latest = None  # type: Optional[Dict[str, Any]]
+        kwargs = {'FunctionName': function_name}  # type: Dict[str, Any]
+        while True:
+            response = lambda_client.list_versions_by_function(**kwargs)
+            for version in response.get('Versions', []):
+                version_name = version.get('Version')
+                if version_name is not None and version_name.isdigit():
+                    if latest is None:
+                        latest = version
+                    elif int(version_name) > int(latest.get('Version', '0')):
+                        latest = version
+            marker = response.get('NextMarker')
+            if marker is None:
+                break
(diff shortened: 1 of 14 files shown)


chalice/config.py
```diff
@@ -343,6 +343,12 @@ def reserved_concurrency(self) -> int:
                                   varies_per_chalice_stage=True,
                                   varies_per_function=True)
 
+    @property
+    def lambda_alias(self) -> Optional[str]:
+        return self._chain_lookup('lambda_alias',
+                                  varies_per_chalice_stage=True,
+                                  varies_per_function=True)
+
```
jaeyoung0509, line 351: - One small UX question: should `lambda_alias` be validated somewhere in the config validation path against Lambda alias naming constraints?

- For example, Chalice could reject values like `$LATEST`, numeric-only aliases, empty strings, or names containing unsupported characters before deployment. That might give users a clearer Chalice-level error instead of surfacing a lower-level Lambda API error later.

jaeyoung0509 (commented):
Thanks for putting this together. 🎉
I left one small UX question inline, but overall this looks like a solid foundation for SnapStart support.

chalice/awsclient.py
```diff
@@ -1002,6 +1002,87 @@ def _update_function_config(
...
+    def publish_function_version(
+        self, function_name: str
+    ) -> Dict[str, Any]:
+        lambda_client = self._client('lambda')
+        try:
+            result = lambda_client.publish_version(
+                FunctionName=function_name
+            )
+        except lambda_client.exceptions.ResourceConflictException:
+            result = self._latest_published_function_version(function_name)
```
ubaskota, line 1014: When `ResourceConflictException` is caught, `_latest_published_function_version` would return the version form the previous deploy.  Instead of using the latest published version when there's a conflict, is it possible to add a delay, similar how its done [here](https://github.com/aws/chalice/blob/50e582ab928fb9314e20142373eedbf031fe2ecc/chalice/awsclient.py#L949)?

chalice/config.py
ubaskota: **+1** -> A validation function similar to the ones in [validate.py](https://github.com/aws/chalice/blob/50e582ab928fb9314e20142373eedbf031fe2ecc/chalice/deploy/validate.py#L189) would work well here.

PR #2190, opened 2026-09-09
Update generated S3 policy actions

alexgromero:
### Overview

Updates S3-related generated IAM behavior in two cases:
- Corrects the generated action for `upload_fileobj` to use `s3:AbortMultipartUpload`.
- Updates the S3 Lambda permission check so that when chalice is adding an account-scoped S3 permission, an existing statement without the expected source account condition is not skipped.

### Testing
- Adds regression coverage for both cases.
- ran `make prcheck`.

---


Alan4506 (changes requested):
Thanks @alexgromero! The fix looks good, but could you add a changelog entry? See https://github.com/aws/chalice/pull/2170/changes#diff-0bbdbb04ecfa94b44db713d8c232e8d6011ba9c06390c5309a6f44fc5955494f for example.

Alan4506 (approved):

PR #2191, opened 2026-09-10
Validate private API config during packaging

alexgromero:
### Overview

Applies the existing API Gateway resource policy validation in the package path before generating package resources.

This keeps `chalice package` aligned with `chalice deploy` for PRIVATE APIs: when no explicit API Gateway policy file is provided, a VPC endpoint id is required so chalice can generate the default resource policy.

This is a fail-fast change for a PRIVATE API configuration that `chalice deploy` already rejects and that previously produced an invalid packaged template.

### Testing
- Adds regression coverage for the package path.
- ran `make prcheck`.

---


.changes/next-release/464347671477125-bugfix-packaging-72900.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "bugfix",
+  "category": "packaging",
+  "description": "Validate private API resource policy configuration in ``chalice package``."
```
jonathan343: nit - For new entries, we should stick to markdown formatting. Looking at older entries, I think a double backtick renders fine, but we should just do the right thing.

```suggestion
  "description": "Validate private API resource policy configuration in `chalice package`."
```

